Issue #34, opened 2022-11-06
Support for modules embedding in Custom


JustSimplyKyle:
From what I've observed, you can only put `Widget` in the Custom modules, I wish you can put modules (e.g. `script`) in the custom modules.
I want this cause I want to do retractable modules

JakeStanger:
Hm, this could be quite complex to do. I could certainly replicate the `script` module behaviour inside without much difficulty, but I'm not sure about nesting the actual modules. 

Are there module types other than `script` you want to be able to do this with?

JustSimplyKyle:
Currently, I only need `script`.

https://user-images.githubusercontent.com/68589851/200174343-d0a23907-3adc-4204-a642-c4f3dc5dd7f4.mp4

This is what I hope to achieve, "retractable" modules

JakeStanger:
It looks like your video failed to upload properly, but I think I get the gist. 

There's a couple of ideas I've got that might help with this:

- To `custom`, add the ability to run polling/watching scripts using the existing syntax. I'm undecided on the syntax, but it'd either use the existing bang `!` or colon format and be something *like* `label = "poll:5000:/path/to/script"`. This would allow the module to display dynamic content, which would be super powerful.
- To all modules, add a config option like `show-if` that evaluates some shell command/script and only shows the module if it receives exit code 0. If not set, the module would always show. This should lift some of the logic out of your scripts and into the bar.

Let me know what you think or if you have any other suggestions

JustSimplyKyle:
Yeah, the `show-if` would be very helpful!
I think I'll personally would prefer the colon syntax

JakeStanger:
Okay I've got an implementation I think I'm happy with up and running, so I'll just need to tidy up and should have something pushed into master in the next few days.

Across the board, there is now a universal 'script' syntax. The places this can be utilised will definitely expand with time, and I still need to standardise the actual `script` module to be parallel. This comes in two flavours:

1. As an object:
```corn
$clock = {
  type = "clock"
  show-if.cmd = "exit 0"
  show-if.interval = 500
}
```

You can create an object with the `cmd` key and optionally the `interval` and `mode` keys.

2. As a string:
```corn
$clock = {
  type = "clock"
  show-if = "500:exit 0"
}
```

You can pass a string containing the command. This can optionally start with the interval and/or mode separated by colons. In cases where scripts are embedded into strings, you must use this syntax and not the object syntax.

I've got two new features that will make use of this syntax:

[...]

PR #38, opened 2022-11-28
feat: common module options (`show_if`, `on_click`, `tooltip`)

### LINKED ISSUE #36: Support for tooltips in various modules
Currently there is no way of customizing the tooltip(e.g. when you hover above the modules, something will pop-up)
I think most of the modules can add this functionality, except for `sys-info`?
I'm not sure how will the syntax be for `sys-info` tho 
https://github.com/JakeStanger/ironbar/wiki/sys-info

JakeStanger:
The first three of many options that are common to all modules.

Includes documentation for each new option.

Resolves #36. Resolves partially #34.

### DIFF
--- a/docs/Configuration guide.md
+++ b/docs/Configuration guide.md
@@ -1,6 +1,12 @@
 By default, you get a single bar at the bottom of all your screens.
 To change that, you'll unsurprisingly need a config file.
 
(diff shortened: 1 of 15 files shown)


JakeStanger:
Only formatting, next few pushes will sort that.

PR #48, opened 2023-01-03
s/pkgs.system/pkgs.hostPlatform.system/g

colemickens:
similar to https://github.com/hyprwm/Hyprland/pull/1296

this fixes using ironbar on my system where I have `nixpkgs.config.allowAliases` set to false. Additionally, this is more correct and explicit for what happens when nixpkgs is used in cross compiling scenarios.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -86,7 +86,7 @@
       ...
     }: let
       cfg = config.programs.ironbar;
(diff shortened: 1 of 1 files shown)


JakeStanger (approved):

JakeStanger:
Thanks for the fix!

colemickens:
np, thanks for ironbar!

PR #64, opened 2023-02-01
feat: add feature flags

### LINKED ISSUE #54: Build feature flags
Compile times are very high. A lot of that time is compiling dependencies only used by a single module that might not be in use.

Per-module feature flags would allow users to compile only what they want into the bar to reduce compile time every update.

Distributed binaries. Will probably do this once the project matures in addition.

JakeStanger:
Flags allow you to disable certain functionality and compile with only select features to reduce build time.

Resolves #54.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,7 +30,13 @@ jobs:
       - name: Check formatting
         run: cargo fmt --check
 
-      - name: Clippy
+      - name: Clippy (base features)
+        uses: actions-rs/clippy-check@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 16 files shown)


JakeStanger:
@yavko I suspect the flags could be controlled via Nix. If you get time to add them to the flake that'd be hugely appreciated :)

yavko:
> @yavko I suspect the flags could be controlled via Nix. If you get time to add them to the flake that'd be hugely appreciated :)
> 
> Docs on PR branch: https://github.com/JakeStanger/ironbar/blob/c347b6c9449ce4e16e2e133d7dd35544ab9a533c/docs/Compiling.md#features

For sure, I'll look into it in a bit!

PR #65, opened 2023-02-03
WIP: fix various system tray issues

oknozor:
This is a work in progress to get better system tray integration. 

These are the issue I found so far : 
- [x] Fallback to default icon theme when tray item does not have a theme 
- [ ] Some tray items are not displayed (ex: element chat)
- [ ] Tray items only show up if launched after iron bar has started (this is probably something I need to fix in stray)

Let me know if you found more issues with the systray.

### DIFF
--- a/src/modules/tray.rs
+++ b/src/modules/tray.rs
@@ -23,14 +23,21 @@ pub struct TrayModule {
 /// Gets a GTK `Image` component
 /// for the status notifier item's icon.
 fn get_icon(item: &StatusNotifierItem) -> Option<Image> {
-    item.icon_theme_path.as_ref().and_then(|path| {
-        let theme = IconTheme::new();
-        theme.append_search_path(path);
-
-        item.icon_name.as_ref().and_then(|icon_name| {
-            let icon_info = theme.lookup_icon(icon_name, 16, IconLookupFlags::empty());
-            icon_info.map(|icon_info| Image::from_pixbuf(icon_info.load_icon().ok().as_ref()))
+    let theme = item
+        .icon_theme_path
+        .as_ref()
+        .map(|path| {
+            let theme = IconTheme::new();
+            theme.append_search_path(path);
+            theme
         })
+        .unwrap_or_default();
+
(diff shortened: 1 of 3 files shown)


JakeStanger:
Hey, thanks for the PR! 

The main issue I'm aware of is that `stray` quite often bails out saying that no service can be found when initialising. It seems to be pretty random, which is what the [retry logic](https://github.com/JakeStanger/ironbar/blob/7c36f5cb0cf03191c9b03e2455b63829a64e402e/src/clients/system_tray.rs#L99) is there to prevent.

In regards to the last point,  that's either a `stray` bug or just a bug with some app's tray icon implementations. Sometimes when you create the connection to DBus, the events for some apps just don't come through. Steam seems to advertise itself on connection much more reliably than other apps.

src/modules/tray.rs
```diff
@@ -23,14 +23,21 @@ pub struct TrayModule {
...
-        theme.append_search_path(path);
-
-        item.icon_name.as_ref().and_then(|icon_name| {
-            let icon_info = theme.lookup_icon(icon_name, 16, IconLookupFlags::empty());
-            icon_info.map(|icon_info| Image::from_pixbuf(icon_info.load_icon().ok().as_ref()))
+    let theme = item
+        .icon_theme_path
+        .as_ref()
+        .map(|path| {
+            let theme = IconTheme::new();
```
JakeStanger, line 30: You can get an icon theme instance from `ModuleInfo`, which uses the icon theme the user has in their config (or should fall back to the default). I don't think there'd be an issue with appending a search path to that (possibly better to?)

yavko:
> * Some tray items are not displayed (ex: element chat)

I think this might've been fixed in #109

yavko:
Since one of the issues has already been fixed in a diff PR, and the other one is probably a `stray` thing, then why not just merge it?

JakeStanger:
That's a fair point, I hadn't considered that.  Will work on getting it merged now.

JakeStanger:
Closed via duplicate #135

PR #67, opened 2023-02-08
Add configurable margins around bar

ttoino:
Adds a config option for adding margins around the bar.

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -74,14 +74,39 @@ impl Default for BarPosition {
     }
 }
 
+#[derive(Debug, Deserialize, Copy, Clone, PartialEq, Eq)]
+pub struct MarginConfig {
+    #[serde(default)]
+    pub bottom: i32,
+    #[serde(default)]
+    pub left: i32,
+    #[serde(default)]
+    pub right: i32,
+    #[serde(default)]
+    pub top: i32,
+}
+
+impl Default for MarginConfig {
+    fn default() -> Self {
+        MarginConfig {
+            bottom: 0,
+            left: 0,
+            right: 0,
+            top: 0,
+        }
+    }
+}
+
 #[derive(Debug, Deserialize, Clone)]
 pub struct Config {
-    #[serde(default = "default_bar_position")]
+    #[serde(default)]
     pub position: BarPosition,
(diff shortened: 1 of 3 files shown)


src/config/mod.rs
```diff
@@ -97,10 +122,23 @@ const fn default_bar_position() -> BarPosition {
     BarPosition::Bottom
 }
 
+const fn default_bar_margin() -> MarginConfig {
```
JakeStanger, line 125: Since you're implementing `Default` above, you can remove this. You can then just use `serde(default)` without setting a func to use the default impl.

JakeStanger:
Thanks for the PR. Do you think there'd be any value in adding a shorthand `margin` option to set all the options in one go?

src/config/mod.rs
ttoino: You're right :sweat_smile: I haven't really used Rust extensively before so I just copied what you did for the bar position. I'll fix it

ttoino:
> Thanks for the PR. Do you think there'd be any value in adding a shorthand `margin` option to set all the options in one go?

I could add that, but at least in my setup that leads to a doubled margin above the bar (when it's on the bottom)

ttoino:
> Can you merge master into this PR please? It includes a formatting fix that's causing the build to fail right now.

Would you prefer a merge or a rebase?

JakeStanger:
> I could add that, but at least in my setup that leads to a doubled margin above the bar (when it's on the bottom)

Hm yeah, I suspect this will probably be the case most of the time. Leave it for now and if anybody wants it, it'll be easy to add in whenever.

> Would you prefer a merge or a rebase?

Good point! A squash/rebase so everything is contained in the singular feature commit would be perfect.

src/config/mod.rs
```diff
@@ -93,10 +118,6 @@ pub struct Config {
     pub monitors: Option<HashMap<String, MonitorConfig>>,
 }
 
-const fn default_bar_position() -> BarPosition {
-    BarPosition::Bottom
-}
-
```
JakeStanger, line 99: Nice spot

JakeStanger (approved):
Thanks!

PR #83, opened 2023-03-18
implement upower module

p00f:
There are a couple of issues:

1) I want to use the popup to show additional details like time to empty/full and battery levels of other connected devices like bluetooth headphones. I've hardcoded the function for now but nothing happens when I click the module

2) upower will tell me what icon to use depending on battery level and whether it is plugged in, how do I implement this? I did `    format: "icon:{icon} {percentage}%"` in config.yaml but that doesn't seem to work

image

Broadly it seems to work fine

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,10 +14,11 @@ default = [
     "music+all",
     "sys_info",
     "tray",
+    "upower",
     "workspaces+all"
 ]
-
 http = ["dep:reqwest"]
+upower = ["upower_dbus", "zbus", "futures-lite"]
 
 "config+all" = ["config+json", "config+yaml", "config+toml", "config+corn"]
 "config+json" = ["universal-config/json"]
@@ -87,6 +88,11 @@ sysinfo = { version = "0.28.4", optional = true }
 
 # tray
 stray = { version = "0.1.3", optional = true }
+
+# upower
+upower_dbus = { version = "0.3.2", optional = true }
+futures-lite = { version = "1.12.0", optional = true }
+zbus = { version = "3.11.0", optional = true }
 
 # workspaces
 swayipc-async = { version = "2.0.1", optional = true }
--- a/src/bar.rs
+++ b/src/bar.rs
@@ -223,6 +223,8 @@ fn add_modules(
             ModuleConfig::SysInfo(mut module) => add_module!(module, id),
             #[cfg(feature = "tray")]
             ModuleConfig::Tray(mut module) => add_module!(module, id),
+            #[cfg(feature = "upower")]
+            ModuleConfig::Upower(mut module) => add_module!(module, id),
             #[cfg(feature = "workspaces")]
             ModuleConfig::Workspaces(mut module) => add_module!(module, id),
         }
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -16,6 +16,8 @@ use crate::modules::script::ScriptModule;
 use crate::modules::sysinfo::SysInfoModule;
 #[cfg(feature = "tray")]
 use crate::modules::tray::TrayModule;
+#[cfg(feature = "upower")]
+use crate::modules::upower::UpowerModule;
 #[cfg(feature = "workspaces")]
 use crate::modules::workspaces::WorkspacesModule;
 use crate::script::ScriptInput;
@@ -57,6 +59,8 @@ pub enum ModuleConfig {
     SysInfo(Box<SysInfoModule>),
     #[cfg(feature = "tray")]
     Tray(Box<TrayModule>),
+    #[cfg(feature = "upower")]
+    Upower(Box<UpowerModule>),
     #[cfg(feature = "workspaces")]
     Workspaces(Box<WorkspacesModule>),
 }
--- a/src/modules/mod.rs
+++ b/src/modules/mod.rs
@@ -19,6 +19,8 @@ pub mod script;
 pub mod sysinfo;
 #[cfg(feature = "tray")]
 pub mod tray;
+#[cfg(feature = "upower")]
+pub mod upower;
 #[cfg(feature = "workspaces")]
 pub mod workspaces;
 
--- a/src/modules/upower.rs
+++ b/src/modules/upower.rs
@@ -1,0 +1,281 @@
+use crate::clients::upower::get_display_proxy;
+use crate::config::CommonConfig;
+use crate::image::ImageProvider;
+use crate::modules::{Module, ModuleInfo, ModuleUpdateEvent, ModuleWidget, WidgetContext};
+use crate::popup::Popup;
+use crate::{await_sync, error, send_async, try_send};
+use color_eyre::Result;
+use futures_lite::stream::StreamExt;
+use gtk::{prelude::*, Button};
+use gtk::{Label, Orientation};
+use serde::Deserialize;
+use tokio::spawn;
+use tokio::sync::mpsc::{Receiver, Sender};
+use upower_dbus::BatteryState;
+use zbus;
+
+const DAY: i64 = 24 * 60 * 60;
+const HOUR: i64 = 60 * 60;
+const MINUTE: i64 = 60;
+
+#[derive(Debug, Deserialize, Clone)]
+pub struct UpowerModule {
+    #[serde(default = "default_format")]
+    format: String,
+
+    #[serde(flatten)]
+    pub common: Option<CommonConfig>,
+}
+
+fn default_format() -> String {
+    String::from("{percentage}%")
+}
+
+#[derive(Clone, Debug)]
+pub struct UpowerProperties {
+    percentage: f64,
+    icon_name: String,
+    state: u32,
+    time_to_full: i64,
+    time_to_empty: i64,
+}
+
+impl Module<gtk::Box> for UpowerModule {
+    type SendMessage = UpowerProperties;
+    type ReceiveMessage = ();
+
+    fn name() -> &'static str {
+        "upower"
+    }
+
+    fn spawn_controller(
+        &self,
+        _info: &ModuleInfo,
+        tx: Sender<ModuleUpdateEvent<Self::SendMessage>>,
+        _rx: Receiver<Self::ReceiveMessage>,
+    ) -> Result<()> {
+        spawn(async move {
+            // await_sync due to strange "higher-ranked lifetime error"
+            let display_proxy = await_sync(async move { get_display_proxy().await });
+            let mut prop_changed_stream = display_proxy.receive_properties_changed().await?;
+
+            let device_interface_name =
+                zbus::names::InterfaceName::from_static_str("org.freedesktop.UPower.Device")
+                    .expect("failed to create zbus InterfaceName");
+
+            let properties = display_proxy.get_all(device_interface_name.clone()).await?;
+
+            let percentage = *properties["Percentage"]
+                .downcast_ref::<f64>()
+                .expect("expected percentage: f64 in HashMap of all properties");
+            let icon_name = properties["IconName"]
+                .downcast_ref::<str>()
+                .expect("expected IconName: str in HashMap of all properties")
+                .to_string();
+            let state = *properties["State"]
+                .downcast_ref::<u32>()
+                .expect("expected State: u32 in HashMap of all properties");
+            let time_to_full = *properties["TimeToFull"]
+                .downcast_ref::<i64>()
+                .expect("expected TimeToFull: i64 in HashMap of all properties");
+            let time_to_empty = *properties["TimeToEmpty"]
+                .downcast_ref::<i64>()
+                .expect("expected TimeToEmpty: i64 in HashMap of all properties");
+            let mut properties = UpowerProperties {
+                percentage,
+                icon_name: icon_name.clone(),
+                state,
+                time_to_full,
+                time_to_empty,
+            };
+
+            send_async!(tx, ModuleUpdateEvent::Update(properties.clone()));
+
+            while let Some(signal) = prop_changed_stream.next().await {
+                let args = signal.args().expect("Invalid signal arguments");
+                if args.interface_name != device_interface_name {
+                    continue;
+                }
+
+                for (name, changed_value) in args.changed_properties {
+                    match name {
+                        "Percentage" => {
+                            properties.percentage = changed_value
+                                .downcast::<f64>()
+                                .expect("expected Percentage to be f64");
+                        }
+                        "IconName" => {
+                            properties.icon_name = changed_value
+                                .downcast_ref::<str>()
+                                .expect("expected IconName to be str")
+                                .to_string();
+                        }
+                        "State" => {
+                            properties.state = changed_value
+                                .downcast::<u32>()
+                                .expect("expected State to be u32");
+                        }
+                        "TimeToFull" => {
+                            properties.time_to_full = changed_value
+                                .downcast::<i64>()
+                                .expect("expected TimeToFull to be i64");
+                        }
+                        "TimeToEmpty" => {
+                            properties.time_to_empty = changed_value
+                                .downcast::<i64>()
+                                .expect("expected TimeToEmpty to be i64");
+                        }
+                        _ => {}
+                    }
+                }
+
+                send_async!(tx, ModuleUpdateEvent::Update(properties.clone()));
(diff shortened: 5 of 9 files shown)


p00f:
@JakeStanger pinging you because I don't think you get notified of draft prs

JakeStanger:
I do get notified actually :)

1. You need a `gtk::Button` inside the widget which sends an open command on click. See how I'm doing it here:
https://github.com/JakeStanger/ironbar/blob/d4dd8c41ea54575173f045740991d8ff0e735211/src/modules/clock.rs#L68-L74

2. It looks like upower just returns the name and there's aren't actually attached to icons anywhere, but rather left to the implementer (that's us!). I would suggest adding a map into the config option like `music` does https://github.com/JakeStanger/ironbar/blob/d4dd8c41ea54575173f045740991d8ff0e735211/src/modules/music/config.rs#L79-L81 and then using an icon label https://github.com/JakeStanger/ironbar/blob/d4dd8c41ea54575173f045740991d8ff0e735211/src/modules/music/mod.rs#L454 to let users choose any text/gtk icon/image they want. You can then pick some sensible defaults from [nerd fonts](https://www.nerdfonts.com/cheat-sheet). This implementation does mean you cannot add the icon to the formatting, and instead it will need to be hard-coded to the start/end, but I think that's fine.

I appreciate that's fairly involved, so give me a heads up if you get stuck.

p00f:
The name of the icon is an actual icon in adwaita so I wanted to just pick that icon instead of making it configurable, this is what waybar does as well

JakeStanger:
Ah okay, that's handy. I think most of the above still applies. You can use `new_icon_label` and pass it the icon name prefixed with `icon:` (so like `icon:battery-good-charging-symbolic`) and it will automatically resolve the image from the icon theme for you. That also means it could be configurable to override the theme default in the future.

yavko:
Is there anything blocking the merge of this? If help is needed, I can help out. I would love to see this in ironbar!

p00f:
@yavko I can't get icons to work

JakeStanger:
Feel free to push your latest version up (might be wise to rebase onto master to get the latest changes) and I can have a look :)

p00f:
The conflict is in Cargo.lock which I'm not sure how to handle, I haven't worked on this after the last push

JakeStanger:
So long as the cargo.toml is fine, you can do whatever with the merge, then just delete cargo.lock and regen to ensure it generates a new valid lock. 

I'm not sure if that's necessarily best practise but it's the easy way lol - I'll just double check it when we get to review stage.

p00f:
@JakeStanger done

p00f:
oh sorry it doesn't compile

p00f:
fixed it

JakeStanger:
Now I'm looking at it, using `new_icon_label` here is a bit more fiddly than I would have liked, so might not be the best path. Luckily it's fairly simple so we can replicate the needed behaviour in a few lines.

What you should able to do here instead is create an image widget and add it to your container:

```rs
let icon = gtk::Image::new();
// ...set up/add widget...
```

Then inside your callback where you get the icon name, you can pass it (still in the `icon:name` format) to `ImageProvider::parse` to load the new icon into your image widget.

```rs
if let Err(err) = ImageProvider::parse(input, icon_theme, size)
            .and_then(|provider| provider.load_into_image(icon) {
    error!("{err:?}");
}
```

I'd advise caching the previous icon name to avoid re-loading the same image over and over, but we can get to bits like that when you're ready for a full review.

Cargo.toml
```diff
@@ -6,16 +6,7 @@ license = "MIT"
...
-    "http",
-    "config+all",
-    "clipboard",
-    "clock",
-    "music+all",
-    "sys_info",
-    "tray",
-    "workspaces+all"
-]
+default = ["http", "config+all", "clipboard", "clock", "music+all", "sys_info", "tray", "workspaces+all"]
```
JakeStanger, line 9: Can this be on multiple lines as it was please? 

Also can you add a `upower` feature flag that enables the new dependencies (enabled by default)?

Cargo.toml
```diff
@@ -95,3 +86,6 @@ futures-util = { version = "0.3.21", optional = true }
 
 # shared
 regex = { version = "1.6.0", default-features = false, features = ["std"], optional = true } # music, sys_info
+upower_dbus = "0.3.2"
+zbus = "3.11.0"
+futures-lite = "1.12.0"
```
JakeStanger, line 91: Please move these out of the shared section into a new `upower` section above.

src/bar.rs
```diff
@@ -225,6 +225,7 @@ fn add_modules(
             ModuleConfig::Tray(mut module) => add_module!(module, id),
             #[cfg(feature = "workspaces")]
             ModuleConfig::Workspaces(mut module) => add_module!(module, id),
+            ModuleConfig::Upower(mut module) => add_module!(module, id),
```
JakeStanger, line 228: Please move this above `Workspaces` to keep them in alphabetical order.

src/config/mod.rs
```diff
@@ -55,6 +56,7 @@ pub enum ModuleConfig {
     Script(Box<ScriptModule>),
     #[cfg(feature = "sys_info")]
     SysInfo(Box<SysInfoModule>),
+    Upower(Box<UpowerModule>),
```
JakeStanger, line 59: Please move this below `Tray` to keep them in alphabetical order.

src/modules/upower.rs
```diff
@@ -0,0 +1,236 @@
...
+#[derive(Debug, Deserialize, Clone)]
+pub struct UpowerModule {
+    format: String,
+
+    #[serde(flatten)]
+    pub common: Option<CommonConfig>,
+}
+
+impl Module<gtk::Box> for UpowerModule {
+    type SendMessage = HashMap<String, OwnedValue>;
```
JakeStanger, line 32: Since you are only consuming a few properties, it would be better to use a custom struct with just the consumed fields rather than a hashmap. 

This removes the need to send the whole properties object across the channel, and also means you can handle the processing (ie the type downcasting) in the controller.

src/modules/upower.rs
```diff
@@ -0,0 +1,236 @@
...
+        spawn(async move {
+            let dbus = zbus::Connection::system().await?;
+            let display_device = UPowerProxy::new(&dbus).await?.get_display_device().await?;
+            let display_device_path = display_device.path();
+            let display_proxy = zbus::fdo::PropertiesProxy::builder(&dbus)
+                .destination("org.freedesktop.UPower")?
+                .path(display_device_path)?
+                .cache_properties(zbus::CacheProperties::No)
+                .build()
+                .await?;
```
JakeStanger, line 54: Since `spawn_controller` runs for every module instance, this could spawn a lot of DBus connections. For other clients I've gone about this by creating a client struct in the clients module, and then wrapped it with `lazy_static`  and `AsyncOnce` to create a singleton instance.

The client struct then has a broadcast channel that the controller can subscribe to, in order to receive events. These can then be processed and sent to the module. See the tray as an example:
https://github.com/JakeStanger/ironbar/blob/8e99fd4d0f01cf4406cabf19fb635f03447753c8/src/clients/system_tray.rs#L96-L124

That can be quite fiddly so let me know if you need a hand.

src/modules/upower.rs
```diff
@@ -0,0 +1,236 @@
...
+    }
+    let hours = (seconds % DAY) / HOUR;
+    if hours > 0 {
+        time_string += &format!(" {hours}h");
+    }
+    let minutes = (seconds % HOUR) / MINUTE;
+    if minutes > 0 {
+        time_string += &format!(" {minutes}m");
+    }
+    time_string
```
JakeStanger, line 214: I think this could cause trailing spaces. For example if you have 0 days, it will start with eg " 3h". I think a trim will be sufficient to sort that.

src/modules/upower.rs
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    fn into_widget(
+        self,
+        context: WidgetContext<Self::SendMessage, Self::ReceiveMessage>,
+        info: &ModuleInfo,
+    ) -> Result<ModuleWidget<gtk::Box>> {
+        let icon_theme = info.icon_theme.clone();
+        let icon_re = Regex::new(r"\{icon}")?;
+        let percentage_re = Regex::new(r"\{percentage}")?;
```
JakeStanger, line 94: Since these are checking exact matches, you can just use `.contains()` on the string.

src/modules/upower.rs
```diff
@@ -0,0 +1,236 @@
...
+        time_string += &format!(" {hours}h");
+    }
+    let minutes = (seconds % HOUR) / MINUTE;
+    if minutes > 0 {
+        time_string += &format!(" {minutes}m");
+    }
+    time_string
+}
+
+fn u32_to_batterystate(number: u32) -> Result<BatteryState, u32> {
```
JakeStanger, line 217: This can be made constant:

```suggestion
const fn u32_to_batterystate(number: u32) -> Result<BatteryState, u32> {
```

src/modules/upower.rs
```diff
@@ -0,0 +1,236 @@
...
+    use BatteryState::*;
```
JakeStanger, line 218: I'm not a huge fan of the wildcard enum import (and neither is pedantic clippy). Could you remove this and qualify the enums please?

src/modules/upower.rs
```diff
@@ -0,0 +1,236 @@
...
+                .path(display_device_path)?
+                .cache_properties(zbus::CacheProperties::No)
+                .build()
+                .await?;
+
+            let mut prop_changed_stream = display_proxy.receive_properties_changed().await?;
+
+            let device_interface_name =
+                zbus::names::InterfaceName::from_static_str("org.freedesktop.UPower.Device")
+                    .unwrap();
```
JakeStanger, line 60: There are a few calls to `unwrap()` throughout. Can these be changed please:

- If it is a situation that should not occur under any scenario (ie indicates a severe bug), use `expect` and pass a clear message as to what went wrong.
- If it could error, please handle the error either by logging it or falling back appropriately.

JakeStanger (changes requested):
Thanks for all the work so far, it's really appreciated. This is looking generally very good so far :)

Most of the bits I've raised are trivial pedantic bits, but there's a couple of more complex issues so do let me know if you need a hand with any of those.

The build is failing currently due to compiler errors when `--no-default-features` is set. One of the reasons is because `regex` is behind a feature flag, but it looks like that's not actually needed here anyway. The other is due to the incomplete icon code, so that should resolve itself in time.

p00f:
the icon is wrong:
image

JakeStanger:
I don't have a device with a battery that I can test this against, but on my desktop the correct `battery-missing-symbolic` icon is showing.

src/modules/upower.rs
```diff
@@ -0,0 +1,297 @@
...
+use upower_dbus::{BatteryState, UPowerProxy};
+use zbus;
+
+const DAY: i64 = 24 * 60 * 60;
+const HOUR: i64 = 60 * 60;
+const MINUTE: i64 = 60;
+
+#[derive(Debug, Deserialize, Clone)]
+pub struct UpowerModule {
+    format: String,
```
JakeStanger, line 25: Please add a default serde value for format.

Cargo.toml
```diff
@@ -14,10 +14,11 @@ default = [
     "music+all",
     "sys_info",
     "tray",
-    "workspaces+all"
+    "workspaces+all",
+    "upower"
```
JakeStanger, line 18: Pedantry regarding the ordering again, but please swap these two around

src/modules/mod.rs
```diff
@@ -19,6 +19,7 @@ pub mod script;
 pub mod sysinfo;
 #[cfg(feature = "tray")]
 pub mod tray;
+pub mod upower;
```
JakeStanger, line 22: This should have a cfg attr:

```suggestion
#[cfg(feature = "upower")]
pub mod upower;
```

src/config/mod.rs
```diff
@@ -16,6 +16,7 @@ use crate::modules::script::ScriptModule;
 use crate::modules::sysinfo::SysInfoModule;
 #[cfg(feature = "tray")]
 use crate::modules::tray::TrayModule;
+use crate::modules::upower::UpowerModule;
```
JakeStanger, line 19: This should have a cfg attr:

```suggestion
#[cfg(feature = "upower")]
use crate::modules::upower::UpowerModule;
```

JakeStanger (changes requested):
A few more little bits, but looking like it's nearly there! Let me know when you've got these, and the few changes leftover from yesterday sorted, thanks!

p00f:
should i retry if connecting to the system bus (or anything in the dbus initialization portion) fails?

p00f:
the icon was fine, it was just black on black.

JakeStanger:
> should i retry if connecting to the system bus (or anything in the dbus initialization portion) fails?

I don't think so. I think if DBus ever fails to connect, that's a much greater problem that retrying probably won't fix, so the best action is probably to just log the error and exit (expect is probably fine in that case)

src/modules/upower.rs
```diff
@@ -0,0 +1,309 @@
...
+        Ok(())
+    }
+
+    fn into_widget(
+        self,
+        context: WidgetContext<Self::SendMessage, Self::ReceiveMessage>,
+        info: &ModuleInfo,
+    ) -> Result<ModuleWidget<gtk::Box>> {
+        let icon_theme = info.icon_theme.clone();
+        let percentage_re = Regex::new(r"\{percentage}")?;
```
JakeStanger, line 173: Please remove this regex - you can just use `.contains()`

src/modules/upower.rs
```diff
@@ -0,0 +1,309 @@
...
+                if args.interface_name != device_interface_name {
+                    continue;
+                }
+
+                for (name, changed_value) in args.changed_properties {
+                    match name {
+                        "Percentage" => {
+                            properties.percentage = changed_value
+                                .downcast::<f64>()
+                                .expect("expected Percentage to be f64")
```
JakeStanger, line 131:
```suggestion
                                .expect("expected Percentage to be f64");
```

src/modules/upower.rs
```diff
@@ -0,0 +1,309 @@
...
+            .orientation(Orientation::Horizontal)
+            .name("popup-upower")
+            .build();
+
+        let label = Label::builder().name("upower-details").build();
+        container.add(&label);
+
+        rx.attach(None, move |properties| {
+            let mut format = String::new();
+            let state: BatteryState = u32_to_batterystate(properties.state).unwrap();
```
JakeStanger, line 247: Please replace this unwrap

JakeStanger (changes requested):
Couple of tiny bits left, then I think it's just about there :)

yavko:
I just wanted to comment, I see you are using the icon from `upower`, which makes sense, but honestly, I think the icon it gives it pretty misleading sometimes, I know this is extra code, but it provides a more accurate icon, in case you want to see how I did it in the past. Not sure if this icon is in all icon packs, but after checking seems it's in papirus and adwaita, and breeze too, it just has a leading zero on all the numbers, so actually maybe this isn't the best idea, as it'd have to be checked at runtime. (also, this was written with an older version of `upower_dbus`, so I don't know if `BatteryState` still exists, or works the same)

```rs
fn get_icon_name(status: BatteryState, capacity: u64) -> String {
    let suffix = if status == BatteryState::Charging {
        "-charging"
    } else {
        ""
    };

    let cap: u8 = match capacity {
        0..=9 => 0,
        10..=19 => 10,
        20..=29 => 20,
        30..=39 => 30,
        40..=49 => 40,
        50..=59 => 50,
        60..=69 => 60,
        70..=79 => 70,
        80..=89 => 80,
        90..=99 => 90,
        100 => 100,
        _ => 0,
    };

    format!("battery-level-{cap}{suffix}-symbolic")
}
```

JakeStanger:
@p00f Thanks for all your work. I'm gonna complete the last few bits & add the docs, then get this merged.

@yavko That sounds good to me. I'll merge the current implementation for now, but if you want to follow up with a PR that adds your icon method, falling back to the upower icon if it's not present in the theme, that would be fantastic (if not that's cool too ofc, I'll add it to the todo list)

PR #113, opened 2023-04-22
Fix `nix run` support

yavko:
So, I had copied Hyprland's nix pkg metadata as a boilerplate, and forgot to replace one of the strings with `ironbar`, and someone brought it up with me that it was erroring weirdly, and so I fixed it, and also added direct support for `nix run` by defining `apps.<system>.default` instead of making it use a fallback ;). Also, I have tested this locally, can't show screenshot, because ironbar spits out a lot of stuff to stdout, and so I lost the output :P.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -57,6 +57,18 @@
           default = self.packages.${system}.ironbar;
         }
     );
+    apps = genSystems (system: let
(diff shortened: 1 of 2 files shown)


JakeStanger:
I'll trust you it works :P Thanks!

PR #140, opened 2023-05-05
[BREAKING] module-level `name` and `class` options

### LINKED ISSUE #75: [BREAKING] Improvements to selecting widgets with CSS
`nth-of-type` is not supported in GTK3 CSS, which makes it harder to style widgets when you have multiple of each instance.

- Widgets should use classes instead of names to be more standard - this is going to be a **BIG BREAKING CHANGE** to existing stylesheets.
- A widget-level `name` config option should be added to set a custom widget name.

You can currently work around using `nth-child` but this requires hard-coding the widget's index within the section.

**Additional context**
Origin: https://github.com/JakeStanger/ironbar/discussions/31

JakeStanger:
BREAKING CHANGE: To allow for the `name` property, any widgets that were previously targeted by name should be targeted by class instead. This affects **all modules and all popups**, as well as several widgets inside modules.

**This will break a lot of rules in your stylesheet**. To attempt to mitigate the damage, a migration script can be found [here](https://raw.githubusercontent.com/JakeStanger/ironbar/master/scripts/migrate-styles.sh) that should get you most of the way.

Resolves #75.

### DIFF
--- a/docs/Configuration guide.md
+++ b/docs/Configuration guide.md
@@ -310,9 +310,12 @@ For information on the `Script` type, and embedding scripts in strings, see [her
 | `transition_type`     | `slide_start` or `slide_end` or `crossfade` or `none` | `slide_start` | The transition animation to use when showing/hiding the widget.                                                    |
 | `transition_duration` | `Integer`                                             | `250`         | The length of the transition animation to use when showing/hiding the widget.                                      |
 
-#### Other
+#### Appearance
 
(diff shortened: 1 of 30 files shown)


JakeStanger:
Permalink to script in PR: https://github.com/JakeStanger/ironbar/blob/564abf136a63bebf54e0a1118872012576fe2cda/scripts/migrate-styles.sh

PR #147, opened 2023-05-09
feat: Add initial impl of favorites and hidden workspaces

### LINKED ISSUE #90: Always visible workspaces

I kinda just want 10 workspaces always showing

An option that allows setting a range of workspaces, that are always shown

Maybe a list of workspaces instead?

yavko:
Also changes some hyprland-rs stuff ;)
needs further testing
fixes #90


JakeStanger:
Awesome, let me know when you're ready to review

yavko:
> Awesome, let me know when you're ready to review

I'll unmark it as WIP, but it needs to be tested with multi monitor as I don't have multiple monitors

yavko:
Like with moving workspaces, also general multi-monitor

src/modules/workspaces.rs
```diff
@@ -176,19 +201,75 @@ impl Module<gtk::Box> for WorkspacesModule {
...
+                                                        &context.controller_tx,
+                                                    );
+                                                    container.add(&item);
+                                                    button_map.insert(name.to_string(), item);
+                                                    fav_names.push(name.to_string());
+                                                }
+                                                added.push(name);
+                                            }
+                                        }
+                                    }
```
JakeStanger, line 272: The nesting here is getting pretty deep. Could you try and refactor to be a bit flatter/concise please? You could definitely use iterators here to help with some of that. It may also be worth extracting into a separate function (assuming that doesn't lead to slowly unravelling and refactoring the whole thing...).

You can use `clippy --  -W clippy::pedantic -W clippy::unwrap_used` to turn on some clippy warnings that the standard config doesn't detect, some of which might have some ideas here (untested). Lots of them are also silly/plain wrong and worth ignoring.

src/modules/workspaces.rs
```diff
@@ -212,11 +293,19 @@ impl Module<gtk::Box> for WorkspacesModule {
...
                     }
                     WorkspaceUpdate::Add(workspace) => {
-                        if self.all_monitors || workspace.monitor == output_name {
+                        if fav_names.contains(&workspace.name) {
+                            let btn = button_map.get(&workspace.name);
+                            if let Some(btn) = btn {
+                                btn.style_context().remove_class("inactive");
+                            }
+                        } else if !hide.contains(&workspace.name) && self.all_monitors
+                            || workspace.monitor == output_name
```
JakeStanger, line 302: I think this is the same condition as above, and it's getting pretty lengthly. It might be worth refactoring out into a little helper func?

JakeStanger (changes requested):
Thanks for the work so far! A few bits to look at. I'll give this a test on both Sway and Hyprland and report back.

JakeStanger:
Just tested on Sway. You currently have to set `all_monitors = true` for either of the new options to take effect. `hidden` does not hide the workspace from the monitor it is currently present on.

2
```diff
@@ -0,0 +1,363 @@
+use crate::clients::compositor::{Compositor, WorkspaceUpdate};
```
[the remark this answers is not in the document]
yavko: lol when did I make this lol

yavko:
> Just tested on Sway. You currently have to set `all_monitors = true` for either of the new options to take effect. `hidden` does not hide the workspace from the monitor it is currently present on.

That's weird, global variant doesn't do any monitor checking

JakeStanger (changes requested):
Changes look good, ty, but this still isn't working as I'd expect. `favourites` still only applies when `all_monitors` is enabled.

Also the `hidden` option means the workspace doesn't show even when active. I'm unsure whether it should be permanently hidden like that? I think I'd personally prefer it to show only when active, although that said I don't personally have a use-case on that so feel free to push back on that if you disagree.

yavko:
> Changes look good, ty, but this still isn't working as I'd expect. `favourites` still only applies when `all_monitors` is enabled.

I don't know how to fix it, ill try to review the code once more 

> Also the `hidden` option means the workspace doesn't show even when active. I'm unsure whether it should be permanently hidden like that? I think I'd personally prefer it to show only when active, although that said I don't personally have a use-case on that so feel free to push back on that if you disagree.

Yeah that makes sense

yavko:
> Changes look good, ty, but this still isn't working as I'd expect. `favourites` still only applies when `all_monitors` is enabled.

Just checked on my machine, and I have `all_monitors` disabled, so idk why you are experiencing this. Code also doesn't have any reason for this happening

src/modules/workspaces.rs
```diff
@@ -176,19 +208,72 @@ impl Module<gtk::Box> for WorkspacesModule {
...
+                            }
+                            match &favs {
+                                Favorites::Global(names) => {
+                                    for name in names {
+                                        if workspaces
+                                            .iter()
+                                            .any(|work| &work.name == name || &work.id == name)
+                                        {
+                                            continue;
+                                        }
```
JakeStanger, line 235: I've only had a brief play but I think the problem is here. `workspaces`` will always be the full list of workspaces (present across all monitors). This code doesn't take `all_monitors`  into consideration (when false) so skips workspaces even if they're not present on the current screen.

JakeStanger:
BTW I'm on holiday next week so won't be able to re-test this now until the week after. I should (*hopefully*) be a little less busy on returning though

src/modules/workspaces.rs
yavko: Ah, lemme try fixing it

yavko: Wait but this only checks if the workspace has already been added, it is to prevent duplicates, has nothing to do with monitor

JakeStanger: I've just tested by commenting out that `continue`` and it definitely is this. You're right that it is the duplicate check logic, but it needs to take `all_monitors` into consideration, because otherwise it's checking for duplicates across ALL bars instead of the current:

- You have a config where workspace 1 is favourited (`favorites = ["1"]`)
- Bar A has workspace 1 already so a button does not need to be explicitly added for the favourite. In this case the `continue` is correct.
- Bar B does not, but the check detects the button on bar A and incorrectly `continues` , leaving no favourite button on B.

yavko: Oh, I see now

yavko: I hope it's fixed now, but I can't really test it, since I don't have another monitor 😛

yavko:
I added silent to hyprland-rs, which makes it not show messages for missing event types

yavko:
Oops forgot to fix hidden to show when active

docs/modules/Workspaces.md
```diff
@@ -8,12 +8,20 @@ Shows all current workspaces. Clicking a workspace changes focus to it.
...
+| Name           | Type                           | Default        | Description                                                                                                                                                               |
+|----------------|--------------------------------|----------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `name_map`     | `Map<string, string/image>`    | `{}`           | A map of actual workspace names to their display labels/images. Workspaces use their actual name if not present in the map. See [here](images) for information on images. |
+| `favorites`    | `Map<string, string>/string[]` | `[]`           | Workspaces to always show, see [#favorites](#favorites)                                                                                                                   |
```
JakeStanger, line 14:
```suggestion
| `favorites`    | `Map<string, string>` or `string[]` | `[]`           | Workspaces to always show, see [#favorites](#favorites)                                                                                                                   |
```

docs/modules/Workspaces.md
```diff
@@ -8,12 +8,20 @@ Shows all current workspaces. Clicking a workspace changes focus to it.
...
+| `icon_size`    | `integer`                      | `32`           | Size to render icon at (image icons only).                                                                                                                                |
+| `all_monitors` | `boolean`                      | `false`        | Whether to display workspaces from all monitors. When `false`, only shows workspaces on the current monitor.                                                              |
+| `sort`         | `added` or `alphanumeric`      | `alphanumeric` | The method used for sorting workspaces. `added` always appends to the end, `alphanumeric` sorts by number/name.                                                           |
+
+### Favorites
+Can be either:
+- A map of monitor names to a list of workspace names
+- A list of workspace names to show on all monitors, will only show if no workspace with the same name exists
+> `all_monitors` will show workspaces even if a matching workspace exists, only when using list
```
JakeStanger, line 24: This last point isn't super clear. I think it'd be better if you re-worked it into the bullet points, so that each one explicitly states how it behaves.

src/modules/workspaces.rs
```diff
@@ -176,19 +209,71 @@ impl Module<gtk::Box> for WorkspacesModule {
...
+                                                        false,
+                                                        true,
+                                                        &name_map,
+                                                        &icon_theme,
+                                                        icon_size,
+                                                        &context.controller_tx,
+                                                    );
+                                                    container.add(&item);
+                                                    button_map.insert(name.to_string(), item);
+                                                    fav_names.push(name.to_string());
```
JakeStanger, line 271: This is still deeper than I'd like. Could you see if you can refactor it to make it a bit flatter, or move the whole match into its own method? Hopefully that should help a bit.

JakeStanger (changes requested):
Cheers for updating the docs too.

Just a couple of tiny bits but otherwise LGTM. I'll test this later today and confirm if it's working as expected.

JakeStanger:
I've just given it a test and it still seems buggy. `hidden` seems to work fine, but `favorites` doesn't.

- With `all_montors = false`,  `favorites` does not appear to work at all, with both the global and per-monitor configs.

image
image

image
image


- With `all_monitors = true`, favorites get duplicated with the global config. It *does* work correctly with per-monitor.

image
image

image
image

yavko:
> I've just given it a test and it still seems buggy. `hidden` seems to work fine, but `favorites` doesn't.
> 
> - With `all_montors = false`,  `favorites` does not appear to work at all, with both the global and per-monitor configs.
> 
> image
> image
> 
> image
> image
> 
> 
> - With `all_monitors = true`, favorites get duplicated with the global config. It *does* work correctly with per-monitor.
> 
> image
> image
> 
> image
> image
> 

Oof, having another monitor would be great to test lol

JakeStanger:
It looks like it [may be possible](https://github.com/any1/wayvnc/issues/7#issuecomment-657092231) to create virtual monitors with Sway

yavko:
> It looks like it [may be possible](https://github.com/any1/wayvnc/issues/7#issuecomment-657092231) to create virtual monitors with Sway

I know Hyprland can do it, but I have no clue on actually viewing the monitor

yavko:
> It looks like it [may be possible](https://github.com/any1/wayvnc/issues/7#issuecomment-657092231) to create virtual monitors with Sway

I tried this out with wayvnc & some web vnc client, and got it working! Now I just need to fix the bug..

yavko:
Finally, back from a vacation, I can work on this again. But I had an idea, could the hidden option also support regex, since some people (like i) have workspaces with a prefix, that don't want to be shown?

JakeStanger:
I think that'd be a good edition, but I also think it should be opt-in (ie don't use regex by default). Perhaps if an entry explictly starts/ends with a slash like `hidden = [ "/pattern/" ]` ?

It'd also be worth considering if there are any other areas (inc. other modules) where it'd also make sense to have regex support, since I'd rather it not be a one-off feature if that makes sense.

yavko:
> I think that'd be a good edition, but I also think it should be opt-in (ie don't use regex by default). Perhaps if an entry explictly starts/ends with a slash like `hidden = [ "/pattern/" ]` ?
> 
> It'd also be worth considering if there are any other areas (inc. other modules) where it'd also make sense to have regex support, since I'd rather it not be a one-off feature if that makes sense.

Good idea, I'll make a issue to track overall regex support then

JakeStanger:
The code to update Hyprland can now be dropped, since that's in master.

@yavko are you likely to get time to finish this in the next couple of days? No worries if not, I'll pick it up to get it across the line so it can be included in 0.13.

My gut feeling is that the regex changes should go into their own PR rather than here to keep this one focused & give a little more time to plan around that.

yavko:
> The code to update Hyprland can now be dropped, since that's in master.

Yeah

> @yavko are you likely to get time to finish this in the next couple of days? No worries if not, I'll pick it up to get it across the line so it can be included in 0.13.

I can't finish it, like I'm not even joking my laptop's battery exploded. I'm looking into buying a new one, but that'll take a while. It was already swollen but then I went on a plane to go back to the US, and now it doesn't work 🫠.

> My gut feeling is that the regex changes should go into their own PR rather than here to keep this one focused & give a little more time to plan around that.

I agree

yavko:
I'm so sorry for not finishing this, I just couldn't due to not having a second monitor. And when I got vnc working and tried to fix the bug, life kept getting in the way.

JakeStanger:
Jesus that's not fun at all, that's pretty scary! I hope you get a new one sorted out soon, and I'll try and get this sorted in the meantime.

No need to apologize at all, I appreciate your efforts getting it this far - as far as I'm concerned everything you've written is code I don't have to :p Sometimes it's just like that, life's been pretty hectic for me too. Nobody here's getting paid to work on this so nobody owes anybody anything.

JakeStanger:
I think between us we were over-complicating the logic a bit (I have a habit of that...). The implementation was close, but I've had a little bit of a re-jiggle of the logic, managed to simplify it a bit and got it all working.

Ofc feel free to have a review. I'll review myself with fresh eyes either late tonight or tomorrow, then I'll update the docs, squash down the commits (keeping your authorship), and merge :)

JakeStanger (approved):

yavko:
> I think between us we were over-complicating the logic a bit (I have a habit of that...). The implementation was close, but I've had a little bit of a re-jiggle of the logic, managed to simplify it a bit and got it all working.

Maybe it was a bit complicated, also thanks for getting it working!

> Ofc feel free to have a review. I'll review myself with fresh eyes either late tonight or tomorrow, then I'll update the docs, squash down the commits (keeping your authorship), and merge :)

Won't let me leave a review lol, but changes seem good to me

JakeStanger:
And we're in! Thanks for all your work on this :)

PR #191, opened 2023-06-16
Add Support For Flatpak Icons

### LINKED ISSUE #190: Add Support For Flatpak Icons
The current implementation in `src/desktop_file.rs` doesn't look for flatpak icons

I manged to make it work 
image

I have brave, lutris and discord installed from flatpak

I'll make a pull request once I finish

body20002:
Fixes #190

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,13 @@ ipc = ["dep:serde_json"]
 
 http = ["dep:reqwest"]
 
-"config+all" = ["config+json", "config+yaml", "config+toml", "config+corn", "config+ron"]
+"config+all" = [
+  "config+json",
+  "config+yaml",
+  "config+toml",
+  "config+corn",
+  "config+ron",
+]
 "config+json" = ["universal-config/json"]
 "config+yaml" = ["universal-config/yaml"]
 "config+toml" = ["universal-config/toml"]
@@ -58,7 +64,15 @@ workspaces = ["futures-util"]
 gtk = "0.17.0"
 gtk-layer-shell = "0.6.0"
 glib = "0.17.10"
-tokio = { version = "1.28.2", features = ["macros", "rt-multi-thread", "time", "process", "sync", "io-util", "net"] }
+tokio = { version = "1.28.2", features = [
+  "macros",
+  "rt-multi-thread",
+  "time",
+  "process",
+  "sync",
+  "io-util",
+  "net",
+] }
 tracing = "0.1.37"
 tracing-subscriber = { version = "0.3.17", features = ["env-filter"] }
 tracing-error = "0.2.0"
@@ -73,7 +87,9 @@ notify = { version = "6.0.1", default-features = false }
 wayland-client = "0.30.2"
 wayland-protocols = { version = "0.30.0", features = ["unstable", "client"] }
 wayland-protocols-wlr = { version = "0.1.0", features = ["client"] }
-smithay-client-toolkit = { version = "0.17.0", default-features = false, features = ["calloop"] }
+smithay-client-toolkit = { version = "0.17.0", default-features = false, features = [
+  "calloop",
+] }
 universal-config = { version = "0.4.0", default_features = false }
 ctrlc = "3.4.0"
 
@@ -117,7 +133,9 @@ hyprland = { version = "=0.3.1", optional = true }
 futures-util = { version = "0.3.21", optional = true }
 
 # shared
-regex = { version = "1.8.4", default-features = false, features = ["std"], optional = true } # music, sys_info
+regex = { version = "1.8.4", default-features = false, features = [
+  "std",
+], optional = true } # music, sys_info
 
 [patch.crates-io]
 stray = { git = "https://github.com/jakestanger/stray", branch = "fix/connection-errors" }
--- a/src/desktop_file.rs
+++ b/src/desktop_file.rs
@@ -1,16 +1,33 @@
-use std::collections::HashMap;
-use std::fs::File;
-use std::io;
-use std::io::BufRead;
-use std::path::PathBuf;
-use walkdir::WalkDir;
+use lazy_static::lazy_static;
+use std::collections::{HashMap, HashSet};
+use std::fs;
+use std::path::{Path, PathBuf};
+use std::sync::Mutex;
+use tracing::warn;
+use walkdir::{DirEntry, WalkDir};
 
-/// Gets directories that should contain `.desktop` files
+use crate::lock;
+
+type DesktopFile = HashMap<String, Vec<String>>;
+
+lazy_static! {
+    static ref DESKTOP_FILES: Mutex<HashMap<PathBuf, DesktopFile>> =
+        Mutex::new(HashMap::new());
+
+    /// These are the keys that in the cache
+    static ref DESKTOP_FILES_LOOK_OUT_KEYS: HashSet<&'static str> =
+        HashSet::from(["Name", "StartupWMClass", "Exec", "Icon"]);
+}
+
+/// Finds directories that should contain `.desktop` files
 /// and exist on the filesystem.
 fn find_application_dirs() -> Vec<PathBuf> {
-    let mut dirs = vec![PathBuf::from("/usr/share/applications")];
-    let user_dir = dirs::data_local_dir();
+    let mut dirs = vec![
+        PathBuf::from("/usr/share/applications"), // system installed apps
+        PathBuf::from("/var/lib/flatpak/exports/share/applications"), // flatpak apps
+    ];
 
+    let user_dir = dirs::data_local_dir(); // user installed apps
     if let Some(mut user_dir) = user_dir {
         user_dir.push("applications");
         dirs.push(user_dir);
@@ -19,55 +36,123 @@ fn find_application_dirs() -> Vec<PathBuf> {
     dirs.into_iter().filter(|dir| dir.exists()).collect()
 }
 
-/// Attempts to locate a `.desktop` file for an app id
-/// (or app class).
-///
-/// A simple case-insensitive check is performed on filename == `app_id`.
-pub fn find_desktop_file(app_id: &str) -> Option<PathBuf> {
+/// Finds all the desktop files
+fn find_desktop_files() -> Vec<PathBuf> {
     let dirs = find_application_dirs();
-
-    for dir in dirs {
-        let mut walker = WalkDir::new(dir).max_depth(5).into_iter();
-
-        let entry = walker.find(|entry| {
-            entry.as_ref().map_or(false, |entry| {
-                let file_name = entry.file_name().to_string_lossy().to_lowercase();
-                let test_name = format!("{}.desktop", app_id.to_lowercase());
-                file_name == test_name
-            })
-        });
-
-        if let Some(Ok(entry)) = entry {
-            let path = entry.path().to_owned();
-            return Some(path);
-        }
-    }
-
-    None
+    dirs.into_iter()
+        .flat_map(|dir| {
+            WalkDir::new(dir)
+                .max_depth(5)
+                .into_iter()
+                .filter_map(Result::ok)
+                .map(DirEntry::into_path)
+                .filter(|file| file.is_file() && file.extension().unwrap_or_default() == "desktop")
+        })
+        .collect()
 }
 
-/// Parses a desktop file into a flat hashmap of keys/values.
-fn parse_desktop_file(path: PathBuf) -> io::Result<HashMap<String, String>> {
-    let file = File::open(path)?;
-    let lines = io::BufReader::new(file).lines();
+/// Attempts to locate a `.desktop` file for an app id
+pub fn find_desktop_file(app_id: &str) -> Option<PathBuf> {
+    // this is necessary to invalidate the cache
+    let files = find_desktop_files();
 
-    let mut map = HashMap::new();
-
-    for line in lines.flatten() {
-        if let Some((key, value)) = line.split_once('=') {
-            map.insert(key.to_string(), value.to_string());
-        }
+    if let Some(path) = find_desktop_file_by_filename(app_id, &files) {
+        return Some(path);
     }
 
-    Ok(map)
+    find_desktop_file_by_filedata(app_id, &files)
+}
+
+/// Finds the correct desktop file using a simple condition check
+fn find_desktop_file_by_filename(app_id: &str, files: &[PathBuf]) -> Option<PathBuf> {
+    let app_id = app_id.to_lowercase();
+
+    files
+        .iter()
+        .find(|file| {
+            let file_name: String = file
+                .file_name()
+                .expect("file name doesn't end with ...")
+                .to_string_lossy()
+                .to_lowercase();
+
+            file_name.contains(&app_id)
+                || app_id
+                    .split(&[' ', ':', '@', '.', '_'][..])
+                    .any(|part| file_name.contains(part)) // this will attempt to find flatpak apps that are like this
+                                                          // `com.company.app` or `com.app.something`
+        })
+        .map(ToOwned::to_owned)
+}
+
+/// Finds the correct desktop file using the keys in `DESKTOP_FILES_LOOK_OUT_KEYS`
+fn find_desktop_file_by_filedata(app_id: &str, files: &[PathBuf]) -> Option<PathBuf> {
+    let app_id = &app_id.to_lowercase();
+    let mut desktop_files_cache = lock!(DESKTOP_FILES);
+
+    files
+        .iter()
+        .filter_map(|file| {
+            let Some(parsed_desktop_file) = parse_desktop_file(file) else { return None };
+
+            desktop_files_cache.insert(file.clone(), parsed_desktop_file.clone());
+            Some((file.clone(), parsed_desktop_file))
+        })
+        .find(|(_, desktop_file)| {
+            desktop_file
+                .values()
+                .flatten()
+                .any(|value| value.to_lowercase().contains(app_id))
+        })
+        .map(|(path, _)| path)
+}
+
+/// Parses a desktop file into a hashmap of keys/vector(values).
+fn parse_desktop_file(path: &Path) -> Option<DesktopFile> {
+    let Ok(file) = fs::read_to_string(path) else {
+        warn!("Couldn't Open File: {}", path.display());
+        return None;
+    };
+
+    let mut desktop_file: DesktopFile = DesktopFile::new();
+
+    file.lines()
+        .filter_map(|line| {
+            let Some((key, value)) = line.split_once('=') else { return None };
+
+            let key = key.trim();
+            let value = value.trim();
+
+            if DESKTOP_FILES_LOOK_OUT_KEYS.contains(key) {
+                Some((key, value))
+            } else {
+                None
+            }
+        })
+        .for_each(|(key, value)| {
+            desktop_file
+                .entry(key.to_string())
+                .or_insert_with(Vec::new)
+                .push(value.to_string());
+        });
+
+    Some(desktop_file)
 }
 
 /// Attempts to get the icon name from the app's `.desktop` file.
 pub fn get_desktop_icon_name(app_id: &str) -> Option<String> {
-    find_desktop_file(app_id).and_then(|file| {
-        let map = parse_desktop_file(file);
-        map.map_or(None, |map| {
-            map.get("Icon").map(std::string::ToString::to_string)
-        })
-    })
+    let Some(path) = find_desktop_file(app_id) else { return None };
+
+    let mut desktop_files_cache = lock!(DESKTOP_FILES);
+
+    let desktop_file = match desktop_files_cache.get(&path) {
+        Some(desktop_file) => desktop_file,
+        _ => desktop_files_cache
+            .entry(path.clone())
+            .or_insert_with(|| parse_desktop_file(&path).expect("desktop_file")),
+    };
+
+    let mut icons = desktop_file.get("Icon").into_iter().flatten();
+
+    icons.next().map(std::string::ToString::to_string)
 }


JakeStanger (changes requested):
Hey, thanks for the PR. 

It looks like there's a lot of code in here that isn't relevant to the fix, and some I'm struggling to understand the reasoning behind. Could you remove any changes that aren't needed from the PR please?

It also looks like the build isn't passing at the moment - can you run `cargo fmt` and check for any  `cargo clippy` warnings before re-committing?

If you're able to amend your changes into the single commit that'd be perfect. Could you also change the commit message to fall in line with [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) to ensure your change is included in the changelog please:

```
fix: not resolving flatpak application icons
```

If you have any questions or want any help let me know :)

body20002:
Thanks for replying @JakeStanger
The reason behind changing the implementation in both `get_desktop_icon_name` and `find_desktop_file` is because the current implementation of `find_desktop_file` is too simple to actually get the required icon, it simply tries to look for `app_id.desktop` which flatpak simply doesn't use, flatpak uses `com.company.app.desktop` this can be found in the docs [here](https://docs.flatpak.org/en/latest/conventions.html#application-ids)

and the reason behind changing `get_desktop_icon_name` was it was calling `find_desktop_file` and in each call it'll find all the `.desktop` files and check them, so I moved out the "finding `.desktop` files" part outside and it only checks, I'm planing on improving it further but I don't know much rust and I'm restricted by the compiler.

You may also found that I used `HashMap<String, Vec<String>>` for parsed desktop files that because some apps like `brave` have multiple entry to one attribute like `"Name"` attribute as shown below

```
[Desktop Entry]
Version=1.0
Name=Brave

[Desktop Action new-window]
Name=New Window

[Desktop Action new-private-window]
Name=New Incognito Window
```
_I removed the irrelevant parts_ 

which if you tried with the old implementation you would only get the last one and get no matches with the `app_id`

Another thing is this part you may find it very confusing, same as me
```rust
let icons = desktop_file.get("Icon").into_iter().flatten();
for icon in icons.clone() {
    return Some(icon.to_owned());
}
```
I couldn't come with a cleaner way to do it.
I tried using the `match` statement in `get_desktop_icon_name` but the code was too messy and nested badly so I just used simple `if` statements.

If you have any idea on how to do it please tell me.
I'll make sure next commit will follow the [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) and run `cargo fmt` and `cargo clippy` before re-committing :)

JakeStanger:
Thanks for the explanation, that's cleared a lot up. I wasn't aware Flatpak did that. 

I am not going to be able to merge the current implementation for a few reasons I'm afraid, which I've gone into below. I have also included a review of the code as it stands to hopefully give a few pointers.

> I don't know much rust and I'm restricted by the compiler.

Rust is a hard language to learn, but it's a great one to, and I'd love to see more of it. I always appreciate anybody who shows an interest and wants to contribute to the project, so I'd like to keep this open and work with you to get this to where it needs to be. For now, could you mark this as a draft please?

## Implementation Issues

One issue is the code quality, but we can get to that through some refactoring.

The main reason I'm not going to be able to merge this is efficiency; remember this code runs **per image** (assuming it needs to be looked up).

Let's start by looking at my current impl, which I'll admit could be improved upon:

1. The application directories are walked, checking for a match on `app_id` by filename.
2. If a match is found, the file is parsed into a map.
3. The `icon` key is loaded from the map.

And compare that to yours:

1. The application directories are walked, finding every `.desktop` file.
2. Each of these files is then parsed into a map. A map of each file against its parsed map is then created.
3. Every single file is then checked, pulling out an array of all the potentially relevant keys, and checking each one for a match.
4. If a match is found, the path to the file is returned.
5. The returned path is then used to pull out the desktop file map which we just had, and pull out the icon key.

The two big problems here are:

- This causes a *lot* of unnecessary I/O, as this means Ironbar will read hundreds of files multiple times on startup. This will need to fundamentally change.
- The implementation creates a lot of complex data structures only to read the values it just had, then throw them away again. This requires a lot of memory allocation and de-allocation, and can be slow. We should be able to sort this with some refactoring.

---

So how do we go about improving this, given Flatpak's requirement to be difficult and give things their own names?

- Prioritise looking at the file names, and avoid reading the file unless absolutely necessary. This can be done by amending my current check to be a bit broader when walking the directories.
- Use properties other than `app_id` . It just so happens that most of the time this matches up, but a window may have other properties on it, we can grab to help. You may also be able to use some GTK/XDG/Flatpak tool to query the data, or at least get close to it (I've not looked into this, there might not be anything).
- If we really do have to read a file, be more efficient about what we do with it. Instead of converting the whole file to a map, only to throw it away again (I'm guilty of this too!), we can just loop over the keys and only look at the ones we want. Something akin to `if (key == "name" && value == app_id) { return icon }`.
- Cache the results! Rather than building out a map every time an icon is read (which could potentially be the same icon), we could cache these and then check the cache for subsequent runs. This avoids the need to read/parse the same file more than once.

## Code Review

Rust's tricky and different and I get that, so I think it's a useful learning exercise to go over some of the things I've spotted and how you could change those still. I'll keep it inline rather than doing a proper GH review because I think that makes sense in this context. I've given some specific examples in context, but I've not covered everything.

Several of the things I've raised here are actually covered by Clippy. If you run that, it'll point out any shortcomings, give reasons, and offer fixes for you. You can turn on some additional checks too like `cargo clippy -- -W clippy::pedantic` (note you'll probably get a lot of stuff that isn't worth paying attention to across the whole program).

A lot of it comes down to learning about iterators. These are rust's superpower, as they let you do some crazy stuff in one chain, remove a lot of boilerplate code, and are smart enough that the compiler can compile them down into a single loop, so you lose no performance. Iterators are available on lots of types by default, including `Option`, `Result` and just about any collection type.

### `.unwrap()` 

Using `unwrap` is [generally discouraged](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap), because where no value is found, it will cause the entire application to panic and exit, and provides little context for users as to why.

In scenarios where crashing is appropriate (ie something has gone *very* wrong internally, and continuing would be bad), you should prefer to use `.expect()` which makes it clear in code and to users what went wrong.

Otherwise, if you would expect the program to carry on and handle a missing value, you should prefer to use `if let` or `match`, or even fall back to a default with something like `unwrap_or`.

In cases where a function returns `Option` or `Result`, you can often use the `?` operator which propagates the error upwards (think of it as unwrap or return).

I've noticed in a few cases, you're checking for an error/none, then unwrapping afterwards - this is logically sound, but it results in the check being done twice - `unwrap()` internally will check if the value is `error` or `none`. 

You can represent this in a slightly cleaner way:

```rs
let entry = match entry {
  Ok(entry) => entry,
  Err(_) => continue, // you may wish to log a warning/error before continuing
}
```

You can also use iterator methods like `.map` and `and_then` to turn `Option<T>` into `Option<TNew>`, meaning you can chain all the work that *should* happen if the value exists, and do nothing if it doesn't. This works for `Result` too.

### `.clone()`

Every call to `clone`* does quite literally take a value in memory and make a full copy of it. Since we're dealing with strings, this means more memory has to be allocated on the heap, and a new pointer assigned on the stack. This can cause the program to ask the OS for memory, which can be slow.

There are scenarios where cloning is required (sometimes you need to consume the data in multiple places). Where you can though, you should try to avoid it and abide by the below:

- Avoid calling the same method multiple times (just good practise in general). For example, all operations you perform on `DirEntry` (`entry`) are done on the underlying path . Your multiple calls to `.into_path` could be replaced with a single variable assignment, removing the need to clone any data.
- Use references where possible. Several methods will work on references anyway, and Rust is smart enough to handle a lot of operations for you. For example `&str == String` is supported, so you can borrow a string value.

The borrow checker is a tricky beast to master, so if you have any problems don't worry too much, we can address those at the end.

* The `Rc` and `Arc` smart pointers are exceptions to this, as they use reference counting and just create a new pointer. Any types that implement `Copy` are also exceptions, since they can be directly assigned on the stack.

### `for { if { return } }`

In cases like these (there's a few variants), we can deploy some  :sparkle: functional programming :sparkle:  Iterators are once again super helpful and let you flatten the code.

For example, take this check:

```rs
for name in names {
    if name.to_lowercase().contains(app_id) {
        return Some(path);
    }
}
```

This could instead be written as:

```rs
if names.any(|name| name.contains(app_id)) { return Some(path); }
```

I would also generally recommend against multiple `returns` in a function. There are cases where it makes sense, as sometimes it just makes it needlessly hard to get the data out or just results in loads of nesting otherwise, but in most cases I would argue it makes code harder to follow. I would instead of for a single point of exit.  

In the case of the example above, you could combine all your checks into one:

```rs
let has_match = startup_wm_classes.any(|class| class.contains(app_id))
    || names.any(|name| name.contains(app_id));
    || // ...

        if has_match {
    Some(path)
} else {
    None
}
```

### `find_desktop_file` and `_find_desktop_file`

These two could easily be combined, since the first does very little.

 I'd also recommend against using `_` at the start of function names. In some languages, I know this is used to show that it should be treated as private, but in rust everything is private by default. Using `pub` will make the method public.

Two methods with near-identical names is also generally not a good idea, as it makes it hard to tell how they differ without looking at the implementation details.

### That `for { return }` at the end

I'm glad you spotted this one yourself :p

So there is a few things we could do here to improve this. 

We'll start with how it currently is:

```rs
for icon in icons.clone() {
    return Some(icon.to_owned());
}

None
````
Now obviously this is going to loop through `Icon`, then return on the first item, so there is no looping really happening. You should **never** do this.

Instead, you can just use `icons` as an iterator (it already is one!), and get the next item. This helpfully returns, an `Option` for us. 

Since this only gets as a reference, we do need to make a clone to take ownership. Helpfully, `Option` has a method called `.cloned()` which does this without needing to unwrap it.

That means the above can be reduced down to this:

```rs
icons.next().cloned()
````

But we can go further! The previous implementation is getting a *reference* to Icon, but why do we need one? `desktop_file` is dropped straight after, so we can just pull the value right out with `remove`  and take ownership:

```rs
let mut icon = desktop_file
    .remove("Icon")
    .into_iter()
    .flatten();

icon.next()
```

We can go one further and combine the whole desktop file lookup and icon lookup into one iterator chain at the end of the function:

```rs
dekstop_files
        .get(&path)
        .and_then(|mut file| file.remove("Icon").into_iter().flatten().next())
```

(Note there is a typo in `dekstop_files` too).

There's room to keep going on this one I'm sure, but I'll let you play with that.

---

I appreciate this is a very long reply, but I do hope it's useful. If anything I've said above isn't clear, or you want some more pointers, please ask and I'll do my best to answer when I can!

body20002:
@JakeStanger is it okay to add this [crate](https://docs.rs/cached/latest/cached/)?

body20002:
if all is fine I'll squish all the changes into 1 commit

JakeStanger:
> @JakeStanger is it okay to add this [crate](https://docs.rs/cached/latest/cached/)?

I think it should be possible to do what we need without. You can use [once_cell](https://doc.rust-lang.org/std/cell/struct.OnceCell.html) (added in Rust 1.70) with a hashmap, and just read/write to that either side of the main logic.

body20002:
> > @JakeStanger is it okay to add this [crate](https://docs.rs/cached/latest/cached/)?
> 
> I think it should be possible to do what we need without. You can use [once_cell](https://doc.rust-lang.org/std/cell/struct.OnceCell.html) (added in Rust 1.70) with a hashmap, and just read/write to that either side of the main logic.

`OnceCell` is a very different approach and would require changing a lot of the code to make it work but I'll try.

body20002:
I forgot to add a way to invalidate the cache in the previous [commit](https://github.com/JakeStanger/ironbar/pull/191/commits/6cee4bbe0388860ca0fbb260d5b49c01b51b4695)

I don't think it's very clean to use OnceCell instead of just using the above [cached](https://docs.rs/cached/latest/cached/) crate

running `cargo clippy` gives me [this](https://rust-lang.github.io/rust-clippy/master/index.html#/declare_interior_mutable_const)
I'm not sure what to do

yavko:
> running `cargo clippy` gives me [this](https://rust-lang.github.io/rust-clippy/master/index.html#/declare_interior_mutable_const)
> I'm not sure what to do

Switch to a static, when using atomics or mutexes that should always be available use static as const doesn't pass references it copies of the value every time

src/desktop_file.rs
```diff
@@ -1,16 +1,25 @@
+use std::cell::OnceCell;
 use std::collections::HashMap;
 use std::fs::File;
 use std::io;
 use std::io::BufRead;
 use std::path::PathBuf;
-use walkdir::WalkDir;
+use walkdir::{DirEntry, WalkDir};
+
+const DESKTOP_FILES_CELL: OnceCell<HashMap<PathBuf, HashMap<String, Vec<String>>>> =
```
yavko, line 9: Replace `const` with `static`

src/desktop_file.rs
```diff
@@ -1,16 +1,25 @@
...
+    OnceCell::new();
+
+const ENTRIES_CELL: OnceCell<Vec<DirEntry>> = OnceCell::new();
```
yavko, line 12: Replace `const` with `static`

yavko (changes requested):

JakeStanger:
> I don't think it's very clean to use OnceCell instead of just using the above [cached](https://docs.rs/cached/latest/cached/) crate

It's generally good to avoid introducing more dependencies where possible, because they increase the build time, and increase the amount of 3rd party code we rely on. `cached` under the hood just uses `OnceCell`, with some extra stuff we don't need here like memoization. 

> I forgot to add a way to invalidate the cache 

What's your reasoning for needing cache invalidation? In my mind, we should never need to remove anything from the cache, since the .desktop files we're reading can be assumed static for the lifetime of Ironbar (it's only held in memory, so the cache gets emptied when you restart anyway).

body20002:
> What's your reasoning for needing cache invalidation? In my mind, we should never need to remove anything from the cache, since the .desktop files we're reading can be assumed static for the lifetime of Ironbar (it's only held in memory, so the cache gets emptied when you restart anyway).
 
 I know that, but from the user pov it would seem like a bug, and I personally wouldn't like restarting ironbar very time I install an app just to get the icons. 
 
 > Replace `const` with `static`
 
when I replace it I get this error message
``` `std::cell::OnceCell<HashMap<PathBuf, HashMap<std::string::String, Vec<std::string::String>>>>` cannot be shared between threads safely the trait `std::marker::Sync` is not implemented for `std::cell::OnceCell<HashMap<PathBuf, HashMap<std::string::String, Vec<std::string::String>>>>`if you want to do aliasing and mutation between multiple threads, use `std::sync::OnceLock` instead shared static variables must have a type that implements `Sync` ```

should I switch to `OnceLock` or what?

> From a quick read through the code, it looks like an improvement, so thanks for your continued work on this. There are still a few things I mentioned in my long post above, but we can go through those in more detail when you're ready.

I wouldn't be able to do it without your guidance :)

**Comment by JakeStanger** [comment]
> I know that, but from the user pov it would seem like a bug, and I personally wouldn't like restarting ironbar very time I install an app just to get the icons.

That's a fair point.

The issue is we still end up walking multiple dirs per icon. I guess that isn't too bad (so long as the .desktop files themselves are cached for Ironbar's lifetime), and I think any alternative is start going to get tricky, so we'll live with that for now.

> should I switch to OnceLock or what?

I've just been fighting the same thing actually. I think you need to use `OnceLock`, and then have a `Mutex` inside that if you need mutability. If you can't get anywhere with that though, the trusty alternative is `lazy_static`.  I've used it a bit for similar patterns, so there's a few examples of that dotted around the code you should be able to use.

**Comment by yavko** [comment]
> > should I switch to OnceLock or what?
> 
> I've just been fighting the same thing actually. I think you need to use `OnceLock`, and then have a `Mutex` inside that if you need mutability. If you can't get anywhere with that though, the trusty alternative is `lazy_static`.  I've used it a bit for similar patterns, so there's a few examples of that dotted around the code you should be able to use.

Doesn't `OnceLock` already offer mutability?

**Comment by JakeStanger** [comment]
> Doesn't OnceLock already offer mutability?

It does, but as far as I could tell you have to declare the static as `mut`, which is unsafe. I'd rather delegate that to a `Mutex` or `RwLock` that handles that for us.

**Comment by body20002** [comment]
> The issue is we still end up walking multiple dirs per icon. I guess that isn't too bad (so long as the .desktop files themselves are cached for Ironbar's lifetime), and I think any alternative is start going to get tricky, so we'll live with that for now.

I think it'd be much more efficient to ask the os how many files are in these dirs or the size of these dirs and just cache it but I think it's just the same as walking these dirs

> I've just been fighting the same thing actually. I think you need to use OnceLock, and then have a Mutex inside that if you need mutability. If you can't get anywhere with that though, the trusty alternative is lazy_static. I've used it a bit for similar patterns, so there's a few examples of that dotted around the code you should be able to use.

Interesting, I'll take a look at the source code
___
I also noticed that ironbar uses a more of memory using the `OnceLock` approach instead of using the cached crate about 4.7 MB more and the app overall usage is 140+ MB is that normal?

I don't think my code is calling the more expensive version of `get_desktopfile`

I'm using this to get the memory usage
```bash
sudo ps -eo size,pid,user,command --sort -size | \
    awk '{ hr=$1/1024 ; printf("%13.2f Mb ",hr) } { for ( x=4 ; x<=NF ; x++ ) { printf("%s ",$x) } print "" }' |\
    cut -d "" -f2 | cut -d "-" -f1 | grep 'ironbar'
```
_I found it on stack overflow_

JakeStanger:
> I think it'd be much more efficient to ask the os how many files are in these dirs or the size of these dirs and just cache it but I think it's just the same as walking these dirs

Feel free to play with this if you want. I'd be interested to know how much of a difference it makes.

--- 

I'd expect memory usage to be higher on a production build, but that is abnormally high (my prod config runs at about 25MB). That said, I would be surprised if this is related unless you are putting *way* too much into the cache. It only needs to contain a map of app id -> icon name, which would be a few bytes of text per entry.

body20002:
> Feel free to play with this if you want. I'd be interested to know how much of a difference it makes.

Couldn't find a way unfortunately

> my prod config runs at about 25MB

How did you use the above command?

JakeStanger:
> Did you use the above command?

I checked using my system monitor. Using the command, it comes out at more like 170Mb so I wouldn't worry. I don't have an in depth enough knowledge of the kernel to understand what's going on there, but it seems fine.

body20002:
@JakeStanger I think everything is good and ready to merge

body20002:
I have a question, why are we even parsing and looking for desktop icons?
doesn't gtk have a function called `lookup_icon`?
and in gtk4 `lookup_icon` have `fallback` args which we can do something like this

```rust
let icon_path = gtk::IconTheme.lookup_icon(
    app_id, 
    [someother_id_for_app], 
    size, 
    scale, 
    direction, 
    flags)
    .get_file().unwrap().get_path().unwrap()
```

and use this inside the widgets like `launcher` and `tray`

body20002:
I know that we in `desktop_file.rs` parse the files because of the app_id
my PR doesn't change the API of `desktop_file.rs` is simply fixes a bug related to flatpak icons

I think It's better if we move to gtk4 and refactor the widgets to use `lookup_icon` instead

JakeStanger:
> why are we even parsing and looking for desktop icons? doesn't gtk have a function called lookup_icon?

Check the image resolver [here](https://github.com/JakeStanger/ironbar/blob/d121dc3d1e9468a67deb528c35fc3897c3840f77/src/image/provider.rs#L80-L82) - we are using that function. Desktop icons are a last-resort fallback if we're unable to find the image by other means. Sometimes apps use icons that don't match their app_id, or sometimes they use absolute paths to images. In these cases we need to check their .desktop file to find these.

> I think It's better if we move to gtk4 and refactor the widgets to use lookup_icon instead

Refactoring to GTK4 is a huge undertaking that will take a long time, and gtk4-layer-shell is tricky to install right now. I have started this though.

src/desktop_file.rs
```diff
@@ -19,55 +35,149 @@ fn find_application_dirs() -> Vec<PathBuf> {
...
-    let file = File::open(path)?;
+/// Given A map of desktop files
+/// find the correct desktop file using the keys in `DESKTOP_FILES_LOOK_OUT_KEYS`
+fn find_desktop_file_by_filedata(
+    app_id: &str,
+    desktop_files_map: MutexGuard<'_, HashMap<PathBuf, DesktopFile>>,
+) -> Option<PathBuf> {
+    let app_id = &app_id.to_lowercase();
+
+    for (path, desktop_file) in desktop_files_map.iter() {
```
JakeStanger, line 119: Prefer to use `.find`() rather than `for { if { return } }`

src/desktop_file.rs
```diff
@@ -19,55 +35,149 @@ fn find_application_dirs() -> Vec<PathBuf> {
...
         if let Some((key, value)) = line.split_once('=') {
-            map.insert(key.to_string(), value.to_string());
+            if !DESKTOP_FILES_LOOK_OUT_KEYS.contains(key) {
+                continue;
+            }
+
+            if let Some(vector) = desktop_file.get_mut(key.trim()) {
+                vector.push(value.trim().to_string());
+            } else {
+                desktop_file.insert(key.trim().to_string(), vec![value.trim().to_string()]);
```
JakeStanger, line 153: There's a lot of duplicate trim and to_string calls going on here. Please only do these once per value - assign them to variables before the check.

src/desktop_file.rs
```diff
@@ -19,55 +35,149 @@ fn find_application_dirs() -> Vec<PathBuf> {
...
-        })
-    })
+    let Some(path) = find_desktop_file(app_id) else { return None };
+
+    let mut desktop_files_cache = DESKTOP_FILES.lock().expect("desktop_files");
+
+    let desktop_file = match desktop_files_cache.get(&path) {
+        Some(desktop_file) => desktop_file,
+        _ => {
+            let desktop_file = parse_desktop_file(path.clone()).expect("desktop_file");
```
JakeStanger, line 170: You might be able to get away with passing a `Path` instead of `PathBuf` to `parse_desktop_file`, which would remove the need to clone.

src/desktop_file.rs
```diff
@@ -19,55 +35,149 @@ fn find_application_dirs() -> Vec<PathBuf> {
...
+            desktop_files_cache.insert(path.clone(), desktop_file);
+            desktop_files_cache.get(&path).expect("desktop_file")
```
JakeStanger, line 172: Since `get` takes a reference to `path`, you will still have ownership of it afterwards. This means you could swap these two calls and remove the need to clone.

src/desktop_file.rs
```diff
@@ -19,55 +35,149 @@ fn find_application_dirs() -> Vec<PathBuf> {
...
+        }
+    };
+
+    let mut icons = desktop_file.get("Icon").into_iter().flatten();
+
+    if let Some(icon) = icons.next() {
+        return Some(icon.to_string());
+    }
+
+    None
```
JakeStanger, line 182:
```suggestion
    icons.next().map(std::string::ToString::to_string)

**Review by JakeStanger — CHANGES_REQUESTED** [review]
This is a lot better than it was - thanks again for your effort so far - but there's still a good number of general code quality issues. I've done a proper review this time to drill down into the specifics. 

Have another read of my big post, especially the bits around iterators, and let me know if you have any questions about any of the bits I've flagged. Once you've got those all resolved, I'll do another review and drill down a bit further.

File: Cargo.toml [hunk 8]
```diff
@@ -6,17 +6,7 @@ license = "MIT"
...
-    "config+all",
-    "clipboard",
-    "clock",
-    "music+all",
-    "sys_info",
-    "tray",
-    "upower",
-    "workspaces+all"
-]
+default = ["http", "config+all", "clipboard", "clock", "music+all", "sys_info", "tray", "upower", "workspaces+all"]
```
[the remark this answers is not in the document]
body20002: This was `cargo fmt` doing not me.

src/desktop_file.rs
```diff
@@ -19,55 +35,149 @@ fn find_application_dirs() -> Vec<PathBuf> {
...
+    dirs.into_iter()
+        .flat_map(|dir| {
+            WalkDir::new(dir)
+                .max_depth(5)
+                .into_iter()
+                .filter_map(|entry| entry.ok())
+                .filter(|file| {
+                    file.clone().into_path().exists()
+                        && file.clone().into_path().is_file()
+                            & file.file_name().to_string_lossy().ends_with("desktop")
```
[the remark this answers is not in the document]
body20002: thanks for telling me about that single `&` I didn't notice it before XD

src/desktop_file.rs
```diff
@@ -1,16 +1,32 @@
...
-use walkdir::WalkDir;
+use std::sync::{Mutex, MutexGuard};
+use walkdir::{DirEntry, WalkDir};
 
-/// Gets directories that should contain `.desktop` files
+type DesktopFile = HashMap<String, Vec<String>>;
+
+lazy_static! {
+    static ref DESKTOP_FILES: Mutex<HashMap<PathBuf, DesktopFile>> =
+        Mutex::new(HashMap::new());
```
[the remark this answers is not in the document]
body20002: Because we need them in `find_desktop_file_by_filedata` these values are the key to find the correct desktop file if `find_desktop_file_by_filename` failed.

if we want to cache the the resolved icons we just need to cache the output of `get_desktop_icon_name` (i.e. we need another hashmap or modify `DESKTOP_FILES` to `HashMap<AppId, (PathBuf, DesktopFile)>`)

src/desktop_file.rs
```diff
@@ -19,55 +35,149 @@ fn find_application_dirs() -> Vec<PathBuf> {
...
 
-    let mut map = HashMap::new();
+    let mut desktop_file: DesktopFile = DesktopFile::new();
 
     for line in lines.flatten() {
         if let Some((key, value)) = line.split_once('=') {
-            map.insert(key.to_string(), value.to_string());
+            if !DESKTOP_FILES_LOOK_OUT_KEYS.contains(key) {
+                continue;
+            }
```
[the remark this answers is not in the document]
body20002: Not sure if I fully understand what you want.
won't `filter_map` and `filter` also increase the time complexity?

src/desktop_file.rs
body20002: This is not true because we want the last value, the last statement is a return statement  
instead of cloning the `desktop_file` I chose to do a `get` after `insert` to get the reference value

Cargo.toml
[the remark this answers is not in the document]
JakeStanger: `cargo fmt` does not touch anything except Rust code

src/desktop_file.rs
```diff
@@ -19,55 +35,149 @@ fn find_application_dirs() -> Vec<PathBuf> {
...
-                file_name == test_name
-            })
-        });
+    for file in files {
+        let file_name = file.file_name().to_string_lossy().to_lowercase();
+        let has_match = file_name.contains(&app_id)
+            || app_id
+                .split(&['-', ' ', ':', '@', '.', '_'][..])
+                .any(|part| file_name.contains(part)); // this will attempt to find flatpak apps that are like this
+                                                       // `com.company.app` or `com.app.something`
```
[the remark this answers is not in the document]
JakeStanger: Oh right I see, I misread this code. I think this could still be problematic - could it end up mis-matching on anything containing `com` for example?

src/desktop_file.rs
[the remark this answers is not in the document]
JakeStanger: Nope! Rust iterators are super smart in that they'll compile down to a single loop. 

I'm not a fan of needing to wrap everything inside an `if let`, which could be moved to a `filter_map`. The `if .contains` check could also be put into a `filter`, which is a bit cleaner than having a `continue` right at the top.

src/desktop_file.rs
JakeStanger: Passing `&PathBuf` is generally recommended against, for the same reasons you should avoid passing `&Vec` or `&String`. What error are you getting? As far as I can see it's only being used for `File::open` which should work fine.

JakeStanger:
Looks like a few of my review points were wrong, so sorry about that. I'll move this out of draft state though, since the implementation seems to be pretty much there now.

src/desktop_file.rs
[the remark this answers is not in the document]
body20002: I don't think any app have the id of `com` or something like `app.com` and if we had `app.com` we would properly catch it in the first check not having to go through the split part (the desktop file would properly have `app.com` in it's name)

if the function returned the wrong path it would be very rare and I think it's better to leave it like this until someone report it so we can make finer adjustment

src/desktop_file.rs
body20002: > Passing &PathBuf is generally recommended against, for the same reasons you should avoid passing &Vec or &String.

Interesting, may I ask why? and how do I do it without the use of clone? since I moved to `&Vec` and `&PathBuf` due to cloning.

> What error are you getting?

nvm I was doing `Path` instead of `&Path`

src/desktop_file.rs
```diff
@@ -19,55 +37,156 @@ fn find_application_dirs() -> Vec<PathBuf> {
...
+                .filter_map(|entry| entry.ok())
+                .map(DirEntry::into_path)
+                .filter(|file| {
+                    file.exists()
+                        && file.is_file()
+                        && file
+                            .file_name()
+                            .expect("file name doesn't end with ...")
+                            .to_string_lossy()
+                            .ends_with("desktop")
```
JakeStanger, line 57: This check is very long and the nesting getting quite deep. Could it be moved out into its own func please?

src/desktop_file.rs
```diff
@@ -19,55 +37,156 @@ fn find_application_dirs() -> Vec<PathBuf> {
...
+    }
+
+    find_desktop_file_by_filedata(app_id, desktop_files_cache)
+}
+
+/// Finds the correct desktop file using a simple condition check
+fn find_desktop_file_by_filename(app_id: &str, files: &Vec<PathBuf>) -> Option<PathBuf> {
+    let app_id = app_id.to_lowercase();
+
+    for file in files {
```
JakeStanger, line 100: This could be changed to use `files.find()` to avoid the need for `has_match` and needing two returns

src/desktop_file.rs
```diff
@@ -19,55 +37,156 @@ fn find_application_dirs() -> Vec<PathBuf> {
...
+) -> Option<PathBuf> {
+    let app_id = &app_id.to_lowercase();
 
-    let mut map = HashMap::new();
+    let has_match = desktop_files_map.iter().find(|(_, desktop_file)| {
+        desktop_file
+            .values()
+            .flatten()
+            .any(|value| value.to_lowercase().contains(app_id))
+    });
```
JakeStanger, line 134: Here you could map to get the path inline:

```suggestion
    desktop_files_map.iter().find(|(_, desktop_file)| {
        desktop_file
            .values()
            .flatten()
            .any(|value| value.to_lowercase().contains(app_id))
    })
    .map(|(path, _)| path.to_owned())
```

Then remove the `match` underneath.

src/desktop_file.rs
```diff
@@ -19,55 +37,156 @@ fn find_application_dirs() -> Vec<PathBuf> {
...
+        Some((path, _)) => Some(path.to_owned()),
+        _ => None,
     }
+}
+
+/// Parses a desktop file into a hashmap of keys/vector(values).
+fn parse_desktop_file(path: &Path) -> Option<DesktopFile> {
+    let file = match File::open(path) {
+        Ok(v) => v,
+        _ => return None, // maybe log here
```
JakeStanger, line 146: Definitely good to add a log here so any IO errors can be seen. I'd say a `warn` is probably suitable?

JakeStanger (changes requested):
I think we're nearly there :)

So the one thing I have noticed that I think should change to try and further reduce IO:

Currently you are reading/parsing EVERY desktop file, then checking through all of these for a match.

It'd be a lot cheaper to read/parse and check for a match in one go, so you only need to read/parse as many files as it takes to get there. 

So I think something akin to this should work:

```rs
files
  .map(parse_desktop_file) // check & add to cache here
  .find(is_match) // check the parsed file keys here
```

src/desktop_file.rs
body20002: it is in its own function `find_desktop_files`

JakeStanger:
If you can get that last change done, then squash your changes down, I reckon we're good to get this merged!

body20002:
@JakeStanger Done

JakeStanger:
The build is failing because it's been rebased onto a broken version of master, which I'm pretty sure is my fault. I'm going to try and fix that for you and make a couple of final refactors that clippy's pedantic mode has detected, then I should be able to merge.

JakeStanger (approved):

JakeStanger:
I found & fixed a bug in the parse_desktop_file code where it was not actually reading files. I've changed it from using a buffered reader to just `fs::read_to_string`. The iterator code there was also a little more complex than it needed to be, so I've rejigged it a bit to flatten it.

I also removed `-` from the list of characters to split on when looking for a partial match on the filename, since it caused some invalid matches - for example my `jetbrans-pycharm` IDE was getting the `jetbrains-toolbox` icon since it matched on `jetbrains`. There may be a need to remove other chars, but I'll wait for bug reports before doing that and at least we'll know where to look :P

Clippy also flagged up a couple of other bits related to simplifying closures and stuff like that, nothing too notable.

JakeStanger:
Thanks again for all your work on this :)

PR #210, opened 2023-07-01
Find Additional Applications dirs in XDG_DATA_DIR

christoph00:
Applications installed via the Home Manager on Nixos are missing icons on the taskbar.  This patch searches for additional application paths using the XDG_DATA_DIR environment variable.

### DIFF
--- a/src/desktop_file.rs
+++ b/src/desktop_file.rs
@@ -1,5 +1,6 @@
 use lazy_static::lazy_static;
 use std::collections::{HashMap, HashSet};
+use std::env;
 use std::fs;
 use std::path::{Path, PathBuf};
 use std::sync::Mutex;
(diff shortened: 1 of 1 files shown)


src/desktop_file.rs
```diff
@@ -33,6 +34,19 @@ fn find_application_dirs() -> Vec<PathBuf> {
         dirs.push(user_dir);
     }
 
+    let extra_data_dirs = match env::var_os("XDG_DATA_DIRS") {
+        Some(val) => {
+            env::split_paths(&val).map(PathBuf::from).collect()
+        },
+        None => {
+            vec![PathBuf::from("/usr/local/share"),
+                 PathBuf::from("/usr/share")]
```
yavko, line 43: I'd recommend adding the flatpak dir as support for flatpak was just added in #191

JakeStanger: This adds to that same vec of dirs, so not necessary to add twice

JakeStanger (changes requested):
Thanks for the fix. 

A couple of bits before this gets merged please:

- Currently this creates a vector for the new dirs jus to join it into the existing vec. Are you able to phase that out, so that:
  -  If it's `Some`, you just keep the iter (don't `collect()`)
  - If it's `None`, you push directly into the `dirs` vec

- Could you please change the commit message to follow [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary), so that the change is included in the changelog? For example `fix: failing to resolve icons with home_manager`. If you push any further changes up please make sure they're squashed into a single commit.

Can you also make sure to `cargo fmt` and fix anything from `cargo clippy` to ensure the build passes?

src/desktop_file.rs
yavko: Oh yeah I see now, I though it replaced it lol

JakeStanger (approved):
LGTM, cheers!

PR #220, opened 2023-07-03
Default configuration

### LINKED ISSUE #219: [Onboarding] Start ironbar with default config if one doesn't exist
Currently when a user tries `ironbar` for the first time they're greeted with an error message.

Ironbar should create a default config if one doesn't exist and start as expected.

Alternatively show a more helpful error message to the user, possibly asking them to run a command to copy an example config.

JakeStanger:
When no config file is found, the bar will now automatically instead load a hard-coded default consisting of the `focused` and `clock` modules, and a `label` informing you the bar is not configured. Instructions are also printed to the log.

Resolves #219


donovanglover:
Just tested this and it looks good to me!

PR #227, opened 2023-07-06
ci: Add binaries to releases to avoid rebuilding on install

slowsage:
I want to get cargo binstall working so the install time can be cut down. As a first step, i have added one target: x86_64-linux-gnu. I have left he boiler plate code for the other targets from https://github.com/nicolas-van/rust-cross-compile-example commented out so they can be added if someone can get them working.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,154 @@
+name: Release
+
+on:
+  workflow_dispatch:
+  release:
+    types: [created]
+
+env:
+  CARGO_TERM_COLOR: always
+
+defaults:
+  run:
+    # necessary for windows
+    shell: bash
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install build deps
+        run: |
+          sudo apt-get update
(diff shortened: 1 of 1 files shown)


yavko:
Shouldn't be too hard with the CI already set up, just need to add the other targets and the releasing of binaries

yavko:
> Shouldn't be too hard with the CI already set up, just need to add the other targets and the releasing of binaries

Wait this is a PR lol this GH client. Sorry my bad

yavko:
Btw just wanna say u wanna follow conventional commits in this repo. So maybe like change ur commit name.

JakeStanger:
Hey, thanks for the PR but I'm going to have to close this.

Binaries are something I do want to add eventually, but I would like to have some discussion first around some of the specifics to make sure it's done right, such as the targets and formats and methods of distribution.

This PR also adds a full new release pipeline, duplicating a lot of the work of the existing one including creating duplicate releases. It also includes Windows/MacOS specifics, which aren't applicable.

> FWIW: It seemed to have issues with gdk-sys when i tried for arch.

I'd check the [PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=ironbar-git) if you're building on Arch, there shouldn't be any special requirements.

Please feel free to open an issue.

PR #236, opened 2023-07-12
[Feature] Add basic bar visibility change to IPC

A-Cloud-Ninja:
Adds a `set_visible` IPC command which takes a boolean `visibility` and string `bar_name` fields.
Simultaneously adds a new config `name` option to name each bar individually (defaults to the monitor name)

Behavior as-is will show/hide *every* bar with a given name. e.g, 2 bars both with the name `test` will be shown/hidden if we pass
`{"type":"set_visible","visible"=true,bar_name="test"}`

### DIFF
--- a/src/bar.rs
+++ b/src/bar.rs
@@ -21,6 +21,13 @@ pub fn create_bar(
     config: Config,
 ) -> Result<()> {
     let win = ApplicationWindow::builder().application(app).build();
+    let bar_name = config
+        .name
+        .clone()
+        .unwrap_or_else(|| format!("bar-{}", get_unique_usize()));
+
+    win.set_widget_name(&bar_name);
+    info!("Creating bar {}", bar_name);
 
     setup_layer_shell(
         &win,
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -97,6 +97,7 @@ pub struct Config {
     pub margin: MarginConfig,
     #[serde(default = "default_popup_gap")]
     pub popup_gap: i32,
+    pub name: Option<String>,
 
     /// GTK icon theme to use.
     pub icon_theme: Option<String>,
@@ -125,6 +126,7 @@ impl Default for Config {
             position: Default::default(),
             height: default_bar_height(),
             margin: Default::default(),
+            name: None,
             popup_gap: default_popup_gap(),
             icon_theme: None,
             ironvar_defaults: None,
--- a/src/ipc/commands.rs
+++ b/src/ipc/commands.rs
@@ -37,4 +37,18 @@ pub enum Command {
         /// The path to the sheet.
         path: PathBuf,
     },
+
+    /// Set the visibility of the bar with the given name.
+    SetVisible {
+        ///Bar name to target.
+        bar_name: String,
+        /// The visibility status.
+        visible: bool,
+    },
+
+    /// Get the visibility of the bar with the given name.
+    GetVisible {
+        /// Bar name to target.
+        bar_name: String,
+    },
 }
--- a/src/ipc/server.rs
+++ b/src/ipc/server.rs
@@ -146,6 +146,33 @@ impl Ipc {
                 }
             }
             Command::Ping => Response::Ok,
+            Command::SetVisible { bar_name, visible } => {
+                let windows = application.windows();
+                let found = windows
+                    .iter()
+                    .find(|window| window.widget_name() == bar_name);
+
+                if let Some(window) = found {
+                    window.set_visible(visible);
+                    Response::Ok
+                } else {
+                    Response::error("Bar not found")
+                }
+            }
+            Command::GetVisible { bar_name } => {
+                let windows = application.windows();
+                let found = windows
+                    .iter()
+                    .find(|window| window.widget_name() == bar_name);
+
+                if let Some(window) = found {
+                    Response::OkValue {
+                        value: window.is_visible().to_string(),
+                    }
+                } else {
+                    Response::error("Bar not found")
+                }
+            }
         }
     }
 
--- a/docs/Configuration guide.md
+++ b/docs/Configuration guide.md
@@ -267,21 +267,22 @@ Check [here](config) for an example config file for a fully configured bar in ea
 
 The following table lists each of the top-level bar config options:
 
-| Name               | Type                                   | Default   | Description                                                                             |
-|--------------------|----------------------------------------|-----------|-----------------------------------------------------------------------------------------|
-| `position`         | `top` or `bottom` or `left` or `right` | `bottom`  | The bar's position on screen.                                                           |
-| `anchor_to_edges`  | `boolean`                              | `false`   | Whether to anchor the bar to the edges of the screen. Setting to false centres the bar. |
-| `height`           | `integer`                              | `42`      | The bar's height in pixels.                                                             |
-| `popup_gap`        | `integer`                              | `5`       | The gap between the bar and popup window.                                               |
-| `margin.top`       | `integer`                              | `0`       | The margin on the top of the bar                                                        |
-| `margin.bottom`    | `integer`                              | `0`       | The margin on the bottom of the bar                                                     |
-| `margin.left`      | `integer`                              | `0`       | The margin on the left of the bar                                                       |
-| `margin.right`     | `integer`                              | `0`       | The margin on the right of the bar                                                      |
-| `icon_theme`       | `string`                               | `null`    | Name of the GTK icon theme to use. Leave blank to use default.                          |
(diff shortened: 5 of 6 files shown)


src/bar.rs
```diff
@@ -21,7 +21,12 @@ pub fn create_bar(
     config: Config,
 ) -> Result<()> {
     let win = ApplicationWindow::builder().application(app).build();
-
+    let bar_name = config
+        .clone()
+        .name
+        .unwrap_or_else(|| monitor_name.to_string());
+    win.set_widget_name(&bar_name);
+    info!("Creating bar {}", bar_name);
```
JakeStanger, line 29:
```suggestion

    win.set_widget_name(&bar_name);
    info!("Creating bar {}", bar_name);
    
```

src/ipc/commands.rs
```diff
@@ -37,4 +37,18 @@ pub enum Command {
...
+
+    /// Sets the visibility of the specified window(s) with the given name (Will operate on
+    /// multiple windows if multiple windows with the same name exist).
+    SetVisible {
+        ///Bar name to target
+        bar_name: String,
+        /// The visibility status.
+        visible: bool,
+    },
+    /// Gets the visibility of windows with the given name.
```
JakeStanger, line 49:
```suggestion

    /// Get the visibility of windows with the given name.
```

src/ipc/commands.rs
```diff
@@ -37,4 +37,18 @@ pub enum Command {
         /// The path to the sheet.
         path: PathBuf,
     },
+
+    /// Sets the visibility of the specified window(s) with the given name (Will operate on
+    /// multiple windows if multiple windows with the same name exist).
```
JakeStanger, line 42:
```suggestion
    /// Set the visibility of the specified window(s) with the given name. 
    /// If multiple windows with the same name exist, this will operate on all of them.
```

src/ipc/commands.rs
```diff
@@ -37,4 +37,18 @@ pub enum Command {
...
+    SetVisible {
+        ///Bar name to target
```
JakeStanger, line 44:
```suggestion
        /// Bar name to target.
```

src/ipc/commands.rs
```diff
@@ -37,4 +37,18 @@ pub enum Command {
...
+        bar_name: String,
+        /// The visibility status.
+        visible: bool,
+    },
+    /// Gets the visibility of windows with the given name.
+    GetVisible {
+        /// Bar name to target
```
JakeStanger, line 51:
```suggestion
        /// Bar name to target.
```

src/ipc/server.rs
```diff
@@ -14,6 +14,7 @@ use tokio::net::{UnixListener, UnixStream};
 use tokio::spawn;
 use tokio::sync::mpsc::{self, Receiver, Sender};
 use tracing::{debug, error, info, warn};
+///gtk::WidgetExt trait
```
JakeStanger, line 17: Please remove this line

src/ipc/server.rs
```diff
@@ -146,6 +147,40 @@ impl Ipc {
...
             }
             Command::Ping => Response::Ok,
+            Command::SetVisible { bar_name, visible } => {
+                let found = application
+                    .windows()
+                    .iter()
+                    .filter(|window| window.widget_name() == bar_name)
+                    .map(|window| window.set_visible(visible))
+                    .collect::<Vec<_>>();
+                if !found.is_empty() {
```
JakeStanger, line 157:
```suggestion

                if !found.is_empty() {
```

src/ipc/server.rs
```diff
@@ -146,6 +147,40 @@ impl Ipc {
...
+                            value: "true".to_string(),
+                        }
+                    } else {
+                        Response::OkValue {
+                            value: "false".to_string(),
+                        }
+                    }
+                } else {
+                    Response::error("Bar not found")
+                }
```
JakeStanger, line 182: Since this is just getting a list of whether each one is visible or not, then only looking for the first record, you could replace the `.filter()` with `.find()`. You could then replace `.map()` with `.any()` to get a single `bool` out.

You can then also just call `.to_string()` straight on the bool, which should eliminate the need for the nested if.

JakeStanger (changes requested):
Thanks for the feature! The config `name` addition is especially useful - I actually need to get this merged to get what I'm working on done :)

I've raised Just a few general code quality bits. If you can resolve those for me please, then squash the commit down to follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary), I'll update the docs and get this merged.

Example commit:
`feat(ipc): add commands to get/set bar visibility`

src/ipc/server.rs
```diff
@@ -146,6 +147,40 @@ impl Ipc {
...
+                let found = application
+                    .windows()
+                    .iter()
+                    .filter(|window| window.widget_name() == bar_name)
+                    .map(|window| window.set_visible(visible))
+                    .collect::<Vec<_>>();
+                if !found.is_empty() {
+                    Response::Ok
+                } else {
+                    Response::error("Bar not found")
```
JakeStanger, line 160: I'm not a big fan of the use of `map()` here because it produces a vector of unit types. Can you see if you can drop both the `map` and `collect`, then just stick in a normal `for` loop inside the `if !is_empty()` pls?

src/config/mod.rs
```diff
@@ -125,6 +126,7 @@ impl Default for Config {
             position: Default::default(),
             height: default_bar_height(),
             margin: Default::default(),
+            name: None,
```
JakeStanger, line 129: I think every bar should have a name to make the internals a little easier, rather than this being optional.

 The easiest way would be to just use `get_random_usize()` and convert to a string (unless you want to add a `get_random_string`() method...).

src/ipc/server.rs
yavko: Also wouldn't it be better to include the generic in the variable deceleration, for code clarity?

src/config/mod.rs
A-Cloud-Ninja: Shouldn't this still be optional *in the config* and the bar otherwise generates one? Internally a name is still assigned at the moment.

JakeStanger: Ah yeah that's fair, that makes sense.

src/ipc/server.rs
JakeStanger: I think in this instance the fact that you're iterating over `application.windows()`, it's pretty clear what the type is going to be, so I don't have issue with that personally.

src/ipc/commands.rs
```diff
@@ -37,4 +37,17 @@ pub enum Command {
...
     },
+
+    /// Set the visibility of the specified window with the given name.
+    SetVisible {
+        ///Bar name to target.
+        bar_name: String,
+        /// The visibility status.
+        visible: bool,
+    },
+    /// Get the visibility of the specified window with the given name.
```
JakeStanger, line 48: Please add whitespace between lines:

```suggestion

    /// Get the visibility of the specified window with the given name.
```

src/ipc/server.rs
```diff
@@ -146,6 +147,31 @@ impl Ipc {
                 }
             }
             Command::Ping => Response::Ok,
+            Command::SetVisible { bar_name, visible } => {
+                let windows = application.windows();
+                let found = windows
+                    .iter()
+                    .find(|window| window.widget_name() == bar_name);
+                if let Some(window) = found {
```
JakeStanger, line 155: Please add whitespace between lines:

```suggestion

                if let Some(window) = found {
```

src/ipc/server.rs
```diff
@@ -146,6 +147,31 @@ impl Ipc {
...
+                } else {
+                    Response::error("Bar not found")
+                }
+            }
+            Command::GetVisible { bar_name } => {
+                let windows = application.windows();
+                let found = windows
+                    .iter()
+                    .find(|window| window.widget_name() == bar_name);
+                if let Some(window) = found {
```
JakeStanger, line 167: Please add whitespace between lines:

```suggestion

                if let Some(window) = found {
```

src/bar.rs
```diff
@@ -21,7 +21,12 @@ pub fn create_bar(
     config: Config,
 ) -> Result<()> {
     let win = ApplicationWindow::builder().application(app).build();
-
+    let bar_name = config
+        .clone()
+        .name
+        .unwrap_or_else(|| format!("bar-{}", get_unique_usize()));
```
JakeStanger, line 27: This clones the entire config just to get the name out, which wastes a lot of memory. Instead, pls just clone the name (`config.name.clone()`)

JakeStanger (changes requested):

JakeStanger:
A couple of issues left from yesterday & a few more minor tweaks, but the implementation looks solid now. Cheers :)

JakeStanger (approved):

PR #258, opened 2023-07-25
ci(nix): Add Crane and Naersk builders

yavko:
Right now the builder is on crane, but it's switchable to naersk (one string)

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,8 +1,65 @@
 {
   "nodes": {
+    "crane": {
+      "inputs": {
+        "flake-compat": "flake-compat",
+        "flake-utils": "flake-utils",
+        "nixpkgs": [
+          "nixpkgs"
+        ],
+        "rust-overlay": "rust-overlay"
+      },
+      "locked": {
+        "lastModified": 1688772518,
+        "narHash": "sha256-ol7gZxwvgLnxNSZwFTDJJ49xVY5teaSvF7lzlo3YQfM=",
+        "owner": "ipetkov",
+        "repo": "crane",
+        "rev": "8b08e96c9af8c6e3a2b69af5a7fa168750fcf88e",
+        "type": "github"
+      },
+      "original": {
+        "owner": "ipetkov",
+        "repo": "crane",
+        "type": "github"
+      }
+    },
+    "flake-compat": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1673956053,
+        "narHash": "sha256-4gtG9iQuiKITOjNQQeQIpoIB6b16fm+504Ch3sNKLd8=",
+        "owner": "edolstra",
+        "repo": "flake-compat",
+        "rev": "35bb57c0c8d8b62bbfd284272c928ceb64ddbde9",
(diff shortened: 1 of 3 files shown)


yavko:
Great the merge broke the lockfile

JakeStanger:
The build took twice as long as usual. Is that normal with crane? 

My concern is that the extended build could outweigh the caching, although I guess we might have to see how that plays out.

yavko:
> The build took twice as long as usual. Is that normal with crane? 
> 
> My concern is that the extended build could outweigh the caching, although I guess we might have to see how that plays out.

I think it's because it's running build, check, and test

JakeStanger:
Is there the option to turn some of those off? I don't really think we need to run both `build` and  `check`, or `test` (since there's no GTK tests rn) in the Nix build if those are running in the main build, we only really need to check it compiles.

yavko:
> Is there the option to turn some of those off? I don't really think we need to run both `build` and `check`, or `test` (since there's no GTK tests rn) in the Nix build if those are running in the main build, we only really need to check it compiles.

There is (this is also for test):
image

yavko:
We can try naersk instead to see if it does caching better for some reason the dependencies weren't cached with crane

JakeStanger:
Looks like the speed is pretty much on par with the original, plus some extra time for caching, so I think we can roll with that. Are you able to squash your commits down into one or two, and use the `ci(nix)` prefix pls, then I'll merge this?

yavko:
> Looks like the speed is pretty much on par with the original, plus some extra time for caching, so I think we can roll with that. Are you able to squash your commits down into one or two, and use the `ci(nix)` prefix pls, then I'll merge this?

Idk how to squash it cuz of the merge commit, I swear the github merge conflict resolver always makes you shoot yourself in the foot by not rebasing

JakeStanger:
Yeah, I always tend to rebase PR commits because merges just add noise and make it difficult to do anything. Give me a bit, I'll see if I can sort it :)

yavko:
> Yeah, I always tend to rebase PR commits because merges just add noise and make it difficult to do anything. Give me a bit, I'll see if I can sort it :)

Thanks! I have very limited amount of knowledge on how to do any merging with git

JakeStanger:
Don't think I would have worked this out in a million years: https://stackoverflow.com/a/69827502

yavko:
> Don't think I would have worked this out in a million years: [stackoverflow.com/a/69827502](https://stackoverflow.com/a/69827502)

Can u co-author me in the commit? I want to have more contributions on my gh profile lol

JakeStanger:
Oh crap sorry, you *were* I'm not sure how I lost that. Fixed now :)

Definitely not trying to steal your glory, I want you credited for your work in the repo.

Will merge as soon as the build completes, thanks again!

yavko:
Thanks hehe

yavko:
Huh normal build failed
image

JakeStanger:
Apparently I committed up every test file I had lying around with this :|

PR #271, opened 2023-08-13
fix(launcher): popup not closing when hover leaves widget

### LINKED ISSUE #224: [Launcher] Popup still visible when no longer hovering over icon
**Describe the bug**
When an application is on the leftmost or rightmost side of the bar with multiple windows, the popup for that application shows indefinitely when leaving the module area.

**To Reproduce**

1. Hover over leftmost/rightmost applications with multiple windows while the bar is centered
2. Observe popup staying until ironbar is clicked or another application icon is hovered

**Expected behavior**
Popup disappears when not hovered over

**System information:**

- Distro: NixOS
- Compositor: Hyprland
- Ironbar version: https://github.com/JakeStanger/ironbar/commit/f8d8c063008340922b20aa665d8f893ab63e82d7

**Configuration**
```json
      "center": [
        {
          "type": "launcher"
        }
      ]
```

JakeStanger:
Fixes #224

### DIFF
--- a/src/modules/launcher/item.rs
+++ b/src/modules/launcher/item.rs
@@ -1,5 +1,6 @@
 use super::open_state::OpenState;
 use crate::clients::wayland::ToplevelHandle;
+use crate::config::BarPosition;
 use crate::gtk_helpers::IronbarGtkExt;
 use crate::image::ImageProvider;
 use crate::modules::launcher::{ItemEvent, LauncherUpdate};
@@ -7,7 +8,7 @@ use crate::modules::ModuleUpdateEvent;
 use crate::{read_lock, try_send};
 use color_eyre::{Report, Result};
 use gtk::prelude::*;
-use gtk::{Button, IconTheme, Orientation};
+use gtk::{Button, IconTheme};
 use indexmap::IndexMap;
 use std::rc::Rc;
 use std::sync::RwLock;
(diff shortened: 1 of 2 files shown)


donovanglover:
This PR causes the popup to close before the user is able to select anything if there is a gap between the bar and the popup. Here's my [style.css](https://github.com/donovanglover/nix-config/blob/10ccd698facdc9306c7b15291ada968b4b9cf8c6/home/ironbar.nix#L51).

JakeStanger:
Is that on the latest commit (https://github.com/JakeStanger/ironbar/commit/54f0f232f208602699e5021942c3d0f3947ca6de) or master, as opposed to the previous one before I force-pushed? That was an issue I identified but resolved, at least for myself.

I've tried your config/CSS and it works fine for me. From the styles, the borders/gaps also line up with my own config.

If this is an issue on the latest commit/master, what resolution/scaling is your display?

donovanglover:
It's on master. I think it's because of my GTK theme since I can get it to work occasionally when using `all: unset`. It only works when I move the cursor really slow though. Any fast movements seem to close the popup

1920x1080 1x

JakeStanger:
Hm yeah, I think it does have something to do with different box sizes in different themes. Now I try it, it is definitely noticeable for me on my 1080p monitor too, whereas my 1440p is (basically) fine.

I might have to do some more digging and see if I can better implement the 'leave' logic.

JakeStanger:
Actually - could you try the commit I've just pushed up pls? I might have just missed something obvious before that gets it to 'good enough' for now.

donovanglover:
Latest commit works and it's good enough for me, thanks :+1:

JakeStanger:
Perfect, cheers for confirming :)

PR #283, opened 2023-08-25
feat(workspaces): visible CSS selector

### LINKED ISSUE #282: [Workspaces] Visible workspaces CSS selector
I'm migrating from Waybar and noticed that there's no CSS selector for visible workspaces. I use two monitors, but since I do all my work on my left monitor, workspaces buttons for my right monitor are never highlighted (active).

A CSS selector such as `.workspaces .item.visible` which is active for workspace buttons whose workspaces are currently visible, including the active one.

After writing the previous section, I noticed that there's `.workspaces .item.inactive` ([implementation](https://github.com/JakeStanger/ironbar/blob/master/src/modules/workspaces.rs#L97)) from #147, but it doesn't do anything. Looking at the call sites, all 3 of them have the `inactive` parameter set to `false`, which would explain why. I think this is what the variable is supposed to represent, but there's no documentation aside from the comment in #90. I'll try to get it working for this purpose, but I figured I should post the issue regardless.

malicean:
Adds CSS class `visible` to all workspace buttons whose workspaces are visible (displayed on a monitor), including the focused workspace. Like `focused`, the rule is `.workspaces .item.visible`.

Implementation replaces `focused: bool` of `clients::compositor::Workspace` with `visibility: Visibility`, which represents focused, visible, and hidden (not on a monitor). Both compositors' listeners now fill the changed field, `clients::compositor::WorkspaceUpdate::Focus` now carries `Workspace`s rather than workspace names, and `modules::workspaces::WorkspacesModule` now adds/removes the new CSS class according to the `Workspace`s.

Fixes #282

### DIFF
--- a/docs/modules/Workspaces.md
+++ b/docs/modules/Workspaces.md
@@ -103,6 +103,7 @@ end:
 | `.workspaces`                  | Workspaces widget box                |
 | `.workspaces .item`            | Workspace button                     |
 | `.workspaces .item.focused`    | Workspace button (workspace focused) |
+| `.workspaces .item.visible`    | Workspace button (workspace visible, including focused) |
 | `.workspaces .item.inactive`   | Workspace button (favourite, not currently open)
 | `.workspaces .item .icon`      | Workspace button icon (any type)     |
 | `.workspaces .item .text-icon` | Workspace button icon (textual only) |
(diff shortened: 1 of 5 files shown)


JakeStanger (commented):
Thanks for the PR, really appreciate the effort you've gone to in order to update the docs & refactor the existing code around it too.

From a quick review it all looks solid, but I want to drill down and test everything before I merge, which I hope to get to in the next couple of days.

The build is failing currently due to some formatter changes in Rust 1.72 that I've just sorted. If you're able to rebase onto that, that'd be fantastic but no worries if not.

malicean:
Just rebased, see if that works.

JakeStanger (approved):

PR #328, opened 2023-10-14
Update Styling guide: add sentences explaining how to increase specificity of entries

Schweber:


Schweber:
I just gave it a go. ;-) I didn't know that i can do that from the browser and don't need to install software and clone everything locally first.

JakeStanger:
Hey, thanks for the addition. Since you're not particularly familiar with the tooling yet, I've made a few amendments for you, which I've explained below:

- I've re-positioned, re-worded and reformatted your text a little bit, just so it makes a little more sense contextually and to avoid the lines in the source file getting too long.
- I've removed the text about looking at the "accessible role" in the inspector as this isn't necessarily correct. You want to look at a widget's object type (so `GtkButton` --> `button`), name or classes for selectors, which is already detailed on the wiki.
- I've 'squashed' your commits down into a single one, and reworded it to adhere to [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/). This ensures you'll be automatically credited in the changelog when the next version eventually goes out.

Schweber:
> * I've 'squashed' your commits down into a single one, and reworded it to adhere to [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).

Thanks! I'll read up on those things.

Issue #358, opened 2023-11-13
Tray icons only show text


chmanie:
**Describe the bug**
With a minimal config, I'm not getting any icons for the tray. This works in waybar.

**To Reproduce**
Open ironbar with the minimal config down below (probably not easily reproducible?) Please let me know what I can do to provide any more info about this. 

I can provide the log though:

<details><summary>Log</summary>

```
2023-11-13T23:48:13.942934Z  INFO ironbar: 92: Ironbar version 0.13.0
2023-11-13T23:48:13.942954Z  INFO ironbar: 93: Starting application

(ironbar:361259): Gtk-WARNING **: 00:48:13.964: Theme parsing error: gtk.css:6691:68: Invalid name of pseudo-class
2023-11-13T23:48:13.985151Z  INFO ironbar::ipc::server: 43: Starting IPC on /run/user/1000/ironbar-ipc.sock
2023-11-13T23:48:13.985243Z  INFO ironbar: 231: Creating bar on 'eDP-1'
2023-11-13T23:48:13.985694Z  INFO ironbar::bar: 33: Creating bar bar-9
2023-11-13T23:48:13.987418Z  INFO system_tray::notifier_watcher: 34: Starting notifier watcher
[...]

PR #360, opened 2023-11-15
Re-use existing IconTheme during icon retrieval

chmanie:
Hey, thank you for starting this great project! This bar works just wonderfully for me.

This PR is an attempted fix for this one thing I noticed here (#358). The bar wasn't showing any icons from my icon theme, whereas the [gtk example](https://github.com/JakeStanger/stray/tree/main/gtk-tray) from stray worked for me.

I am no GTK expert at all, so I just compared the code. Apparently, it's important to re-use an initialized IconTheme after it was initially created (maybe it's a singleton?).

In any case, this fixes it for me. Let me know if that's an acceptable change or if I need to dig deeper here :)

Thanks again!


src/modules/tray.rs
```diff
@@ -27,15 +27,11 @@ pub struct TrayModule {
...
-    let theme = item
-        .icon_theme_path
-        .as_ref()
-        .map(|path| {
-            let theme = IconTheme::new();
-            theme.append_search_path(path);
-            theme
-        })
-        .unwrap_or_default();
+    let theme = IconTheme::default().unwrap_or(IconTheme::new());
```
JakeStanger, line 30: Rather than creating a new theme, you can use the one on the `ModuleInfo` instance passed into `into_module`. This is shared throughout the bar and uses the theme specified in the config.

src/modules/tray.rs
```diff
@@ -27,15 +27,11 @@ pub struct TrayModule {
...
+
+    if let Some(path) = item.icon_theme_path.as_ref() {
+        theme.append_search_path(path);
```
JakeStanger, line 33: The reason I held off this was because of `append_search_path` - I need to be sure that it doesn't just keep appending the same path and gradually slowing down, but never got round to testing that. If it does, a check will need to be put in to avoid adding duplicate paths. Are you able to look at that for me please?

JakeStanger (changes requested):
Hey, thanks for the PR! You're right about the theme, and I think this is the only exception throughout the application. I wasn't aware of the implication, so glad to get that fixed.

I've left a couple of comments if you're able to look at those please? 

Can you use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) to ensure you're included in the changelog (and squash any further commits down if neccessary)? Something like `fix(tray): icons sometimes rendering as text` so it's clear what has been fixed. Any further info can be in the body.

src/modules/tray.rs
chmanie: Of course, can do!

src/modules/tray.rs
chmanie: I looked into an old comment of yours suggesting that. Sadly it's not very straightforward as it is moved into a closure and potentially needs to be ``static`. I'll take another look

JakeStanger: `gtk-rs` structs are basically just wrappers around pointers, which means you can `.clone()` very cheaply and get ownership of the same instance.

chmanie:
Okay, lots of bad news here.

First off sadly the `IconTheme` in gtk-rs is not `Copy`, so I can't really clone it.

Secondly, you're correct, the `append_search_path` function is indeed [always adding to the search paths array](https://gitlab.gnome.org/GNOME/gtk/-/blob/gtk-3-24/gtk/gtkicontheme.c#L972-977). And while gtk has a [`get_search_path` function](https://gitlab.gnome.org/GNOME/gtk/-/blob/gtk-3-24/gtk/gtkicontheme.c#L928), `gtk-rs` [does not](https://github.com/gtk-rs/gtk3-rs/blob/master/gtk/src/auto/icon_theme.rs). There is also a [`set_search_path` function](https://gitlab.gnome.org/GNOME/gtk/-/blob/gtk-3-24/gtk/gtkicontheme.c#L891) in gtk but I think this wouldn't help either (and is also not supported by `gtk-rs`). So the last resort for this would probably be to keep track of the search paths ourselves (at least for now and maybe make a commit upstream to `gtk-rs` even though I have not the slightest idea how [`gir`](https://github.com/gtk-rs/gir) works).

JakeStanger:
> First off sadly the IconTheme in gtk-rs is not Copy, so I can't really clone it.

It doesn't implement Copy, but it does implement Clone as all GObjects do. This doesn't actually clone the object as that exists within C space, but just clones the reference to it (much like how an `Rc` or `Arc` works). You can clone any GTK object without worrying :)

> So the last resort for this would probably be to keep track of the search paths ourselves

I think this is fine as a stop-gap, just a static (using `lazy_static` or whatever) `Vec` will do the job, and it'll be cheap enough to not worry. 

You *can* still do it natively if you're up for it, as `gtk-rs` does export all the methods [via ffi](https://docs.rs/gtk-sys/0.18.0/gtk_sys/fn.gtk_icon_theme_get_search_path.html). This does require some unsafe code as you have to use it C-style and pass in the pointers, but it's not too bad. I've done it [here](https://github.com/JakeStanger/ironbar/blob/ae75fcf2b2bb2695254fc58f27e0872a8d0156b8/src/image/provider.rs#L214-L218) for image loading if you want an example

chmanie:
Thank you, lots of good pointers there! I have my work cut out for me then :)

chmanie:
OK, here we go :)

I added a `get_icon_theme_search_paths` method that gets a hashmap of the existing search paths. Furthermore I am checking if the path is empty (which happens to me quite a bit), so we don't do any processing there.

What fixed my issue in the end was just to use the existing icon theme (not the appending of icon search paths).

But this should be improved now in a lot of ways, let me know what you think :)

JakeStanger (approved):
Code lgtm, I'll give it a quick test myself as soon as I can to make sure it all looks good, and then hopefully merge. Thanks for your work on this!

> BTW: This was so much fun, I learned a lot about GTK and how to use C bindings through FFIs. If you have any more issues I could take on, maybe tag them with "entry level" or "good first issue" so that I can see what I can potentially tackle!

Love to hear it! I'll definitely go through and tag them as it'd be great to get a hand wherever possible :)

chmanie:
> Did you run into that yourself at all, just out of curiosity?

I haven't really noticed anything like that but I'll test it again and pay some attention to it. Thanks for merging and tagging the issues! Looking forward to helping out more :)

Issue #129, opened 2023-04-28
Not able to get the intel CPU temp 


bluebyt:
I am not able to find the right label for the intel CPU temperature.
Bases on your example, I google the label and this is an AMD CPU.
{temp_c:k10temp_Tccd1}
![slurped](https://user-images.githubusercontent.com/18442224/235263248-46deae57-72b9-4612-8517-d1758ccd188c.png)

I tried to do the same thing with my intel CPU.
$sensors coretemp-isa-0000

```
coretemp-isa-0000
Adapter: ISA adapter
Package id 0:  +45.0°C  (high = +84.0°C, crit = +100.0°C)
Core 0:        +42.0°C  (high = +84.0°C, crit = +100.0°C)
Core 1:        +41.0°C  (high = +84.0°C, crit = +100.0°C)
Core 2:        +45.0°C  (high = +84.0°C, crit = +100.0°C)
Core 3:        +40.0°C  (high = +84.0°C, crit = +100.0°C)
Core 4:        +42.0°C  (high = +84.0°C, crit = +100.0°C)
Core 5:        +42.0°C  (high = +84.0°C, crit = +100.0°C)
```

Here all labels I tried: 
{temp_c:coretemp_Package id 0}
{temp_c:coretemp-isa-0000_Package id 0}

I always see the label itsef in the bar instead of the temperature value:
![slurped](https://user-images.githubusercontent.com/18442224/235262827-81c181d1-cfc2-4cb8-a7e1-9ba42d65eeec.png)

sensors
https://pastebin.com/xyq2692g

JakeStanger:
This is an area I need to improve and I'm surprised it's not come up sooner tbh. I aim to add a command to the CLI when I eventually get round to that which will dump the sensor names to make this a bit easier.

The `sysinfo` library under the hood uses a few different formats for naming the sensors depending on what is available:
https://github.com/GuillaumeGomez/sysinfo/blob/a21fd2a5bd62b70d2f3ffb15ff5b343eeee2049a/src/linux/component.rs#L285-L292

Ironbar then replaces any spaces in the label with a `-` (which should probably be an `_` , or arguably doesn't need to happen at all, so I might have to address that in the future):
https://github.com/JakeStanger/ironbar/blob/e5bc44168f29374dab6e15fa049627a10df14f3a/src/modules/sysinfo.rs#L291

I think based off that, if you try replacing the spaces with `-` hyphens, you should be along the right lines.

[...]

PR #379, opened 2023-12-15
Fix cpu temp examples

cyhyraethz:
Use - instead of _ for [sensor], as in #129


JakeStanger:
Thanks for this, I'll merge this to sort the docs. 

As per my comment in #129 I will probs eventually switch the names to use `_`` instead of `-` for consistency, but that will likely not be until I touch the other sys_info related issues.

Issue #403, opened 2024-01-11
Make wlr-foreign-toplevel-management a feature flag


thmasq:
#### The context:
Currently Ironbar requires the [wlr-foreign-toplevel-management](https://wayland.app/protocols/wlr-foreign-toplevel-management-unstable-v1) protocol is a hard requirement, and the program will panic if it is not available. This protocol is essential for the `launcher` and `focused` modules of Ironbar, because it gives waybar access to context menus for the running programs, and also their status within the compositor.

#### Reasoning:
This protocol is not universal, with quite a few compositors out there currently not supporting it, such as Gnome's Mutter, KDE's Kwin, and compositors made using Smithay. There is a very interesting [discussion](https://www.mail-archive.com/wayland-devel@lists.freedesktop.org/msg39978.html) in the mailing lists for Wayland, where Gnome's Jonas Ådahl raises some security concerns related to the protocol, as it breaks breaks client isolation. This could make a point for use-cases where this extra security is necessary.
Additionally, users who do not want to display running applications could simply disable the feature to reduce compilation time and have an overall leaner desktop.

#### Solution
Create a new feature flag in the modules category called `runner`, and have it disable both the `launcher` and `focused` modules, which depend on the toplevel manager. The client code will also have to be modified accordingly to disable the checks that would otherwise make the program panic.

---

I plan on trying to implement this feature by the next month as I'm kind of busy for now, but I think I can do it.

thmasq:
I see, that's a good point indeed. It wouldn't hurt to make it as modular as possible if we can, and I don't think it would be a lot harder to implement. Perhaps it might be an interesting idea to separate the modules in tiers of support (similar to how rustc does), I was thinking, but for that I should open another issue after this one is closed. This idea needs more thorough thought, but it would be nice for users to know what modules are expected to work on different implementations of Wayland. Just a thought that passed my mind, really.

thmasq:
As a POC, I tried just deleting the parts related to the launcher and focused modules, and can confirm it runs on Pinnacle just fine. Now to put this under feature flags.
![Screenshot from 2024-01-11 17-21-56](https://github.com/JakeStanger/ironbar/assets/78622691/5e6cc4f5-0f59-4b01-a772-b90db9a1ea01)

JakeStanger:
> Perhaps it might be an interesting idea to separate the modules in tiers of support 

[...]

PR #404, opened 2024-01-11
Make feature flags for the "focused" and launcher "modules"

thmasq:
Implementation to solve issue [403](https://github.com/JakeStanger/ironbar/issues/403). Please tell me if there is something I should change or refactor.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,11 +11,13 @@ keywords = ["gtk", "bar", "wayland", "wlroots", "gtk-layer-shell"]
 [features]
 default = [
     "cli",
-    "ipc",
-    "http",
-    "config+all",
     "clipboard",
     "clock",
+    "config+all",
+    "focused",
+    "http",
+    "ipc",
+    "launcher",
     "music+all",
     "sys_info",
     "tray",
@@ -44,6 +46,10 @@ http = ["dep:reqwest"]
 clipboard = ["nix"]
 
 clock = ["chrono"]
+
+focused = []
+
+launcher = []
 
 music = ["regex"]
 "music+all" = ["music", "music+mpris", "music+mpd"]
--- a/src/clients/wayland/mod.rs
+++ b/src/clients/wayland/mod.rs
@@ -1,13 +1,9 @@
 mod macros;
 mod wl_output;
 mod wl_seat;
-mod wlr_foreign_toplevel;
 
 use crate::error::ERR_CHANNEL_RECV;
-use crate::{
-    arc_mut, delegate_foreign_toplevel_handle, delegate_foreign_toplevel_manager, lock,
-    register_client, send, Ironbar,
-};
+use crate::{arc_mut, lock, register_client, send, Ironbar};
 use std::sync::{Arc, Mutex};
 
 use calloop_channel::Event::Msg;
@@ -26,11 +22,17 @@ use tokio::sync::{broadcast, mpsc};
 use tracing::{debug, error, trace};
 use wayland_client::globals::registry_queue_init;
 use wayland_client::{Connection, QueueHandle};
-
-use wlr_foreign_toplevel::manager::ToplevelManagerState;
-
 pub use wl_output::{OutputEvent, OutputEventType};
-pub use wlr_foreign_toplevel::{ToplevelEvent, ToplevelHandle, ToplevelInfo};
+
+cfg_if! {
+    if #[cfg(any(feature = "focused", feature = "launcher"))] {
+        mod wlr_foreign_toplevel;
+        use crate::{delegate_foreign_toplevel_handle, delegate_foreign_toplevel_manager};
+        use wlr_foreign_toplevel::manager::ToplevelManagerState;
+        pub use wlr_foreign_toplevel::{ToplevelEvent, ToplevelHandle, ToplevelInfo};
+
+    }
+}
 
 cfg_if! {
     if #[cfg(feature = "clipboard")] {
@@ -56,6 +58,7 @@ cfg_if! {
 #[derive(Debug)]
 pub enum Event {
     Output(OutputEvent),
+    #[cfg(any(feature = "focused", feature = "launcher"))]
     Toplevel(ToplevelEvent),
     #[cfg(feature = "clipboard")]
     Clipboard(ClipboardItem),
@@ -68,7 +71,9 @@ pub enum Request {
     #[cfg(feature = "ipc")]
     OutputInfoAll,
 
+    #[cfg(any(feature = "focused", feature = "launcher"))]
     ToplevelInfoAll,
+    #[cfg(feature = "launcher")]
     ToplevelFocus(usize),
 
     #[cfg(feature = "clipboard")]
@@ -85,6 +90,9 @@ pub enum Response {
     #[cfg(feature = "ipc")]
     OutputInfoAll(Vec<smithay_client_toolkit::output::OutputInfo>),
 
+    #[cfg(feature = "launcher")]
+    ToplevelInfo(Option<ToplevelInfo>),
+    #[cfg(any(feature = "focused", feature = "launcher"))]
     ToplevelInfoAll(Vec<ToplevelInfo>),
 
     #[cfg(feature = "clipboard")]
@@ -106,6 +114,7 @@ pub struct Client {
     rx: Arc<Mutex<std::sync::mpsc::Receiver<Response>>>,
 
     output_channel: BroadcastChannel<OutputEvent>,
+    #[cfg(any(feature = "focused", feature = "launcher"))]
     toplevel_channel: BroadcastChannel<ToplevelEvent>,
     #[cfg(feature = "clipboard")]
     clipboard_channel: BroadcastChannel<ClipboardItem>,
@@ -119,6 +128,7 @@ impl Client {
         let (response_tx, response_rx) = std::sync::mpsc::channel();
 
         let output_channel = broadcast::channel(32);
+        #[cfg(any(feature = "focused", feature = "launcher"))]
         let toplevel_channel = broadcast::channel(32);
 
         #[cfg(feature = "clipboard")]
@@ -131,6 +141,7 @@ impl Client {
         // listen to events
         {
             let output_tx = output_channel.0.clone();
+            #[cfg(any(feature = "focused", feature = "launcher"))]
             let toplevel_tx = toplevel_channel.0.clone();
 
             #[cfg(feature = "clipboard")]
@@ -141,6 +152,7 @@ impl Client {
                 while let Some(event) = event_rx.recv().await {
                     match event {
                         Event::Output(event) => send!(output_tx, event),
+                        #[cfg(any(feature = "focused", feature = "launcher"))]
                         Event::Toplevel(event) => send!(toplevel_tx, event),
                         #[cfg(feature = "clipboard")]
                         Event::Clipboard(item) => send!(clipboard_tx, item),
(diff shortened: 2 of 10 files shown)


JakeStanger:
Thanks for the PR. As it is, the build is failing as you need to ensure code is formatted with  `cargo fmt` .

There are also places that still need the feature flag (such as the `mod` statements in the Wayland client). If you compile with `cargo build --no-default-features`  you will spot these, most likely as unused code warnings. Don't worry if you can't every single one of these, but try to put everything you reasonably can behind a flag.

thmasq:
I'm sorry for the trouble if that's the case. This is my first time contributing to someone else's repository, and I'm not quite experienced with git. This time I tested the changes with more combinations of the build flags to make extra sure it will compile, but there are still some warnings I haven't managed to fix. These warnings only show up when compiling without the focused and launcher features though. When compiling with default features, the only new warning I get is related to a match arm for on_request() events in src/clients/mod.rs. The compiler was requiring me to add this case, even though it is unreachable:
```rust
smithay_client_toolkit::reexports::calloop::channel::Event::Msg(
                crate::clients::wayland::Request::ToplevelInfoAll,
            ) => unreachable!(),
```
I guess I'm at the end of my skill to actually fix this.

Cargo.toml
```diff
@@ -20,7 +20,9 @@ default = [
     "sys_info",
     "tray",
     "upower",
-    "workspaces+all"
+    "workspaces+all",
+    "focused",
+    "launcher"
```
JakeStanger, line 25: Please move these to be in alphabetical order with the entries above.

src/clients/wayland/mod.rs
```diff
@@ -337,6 +362,10 @@ impl Environment {
                 send!(env.response_tx, Response::ClipboardItem(item));
             }
             calloop_channel::Event::Closed => error!("request channel unexpectedly closed"),
+            #[cfg(not(any(feature = "focused", feature = "launcher")))]
+            smithay_client_toolkit::reexports::calloop::channel::Event::Msg(
+                crate::clients::wayland::Request::ToplevelInfoAll,
+            ) => unreachable!(),
```
JakeStanger, line 368: The `ToplevelInfoAll` request and response variants can both be behind the feature flag, which would allow you to remove this match case.

src/config/mod.rs
```diff
@@ -140,7 +144,16 @@ impl Default for Config {
...
-            center: Some(vec![ModuleConfig::Focused(Box::default())]),
+            center: {
+                #[cfg(feature = "focused")]
+                {
+                    Some(vec![ModuleConfig::Focused(Box::default())])
+                }
+                #[cfg(not(feature = "focused"))]
+                {
+                    None
+                }
```
JakeStanger, line 155: Please remove these braces around Some and None as they are not necessary

JakeStanger (changes requested):

JakeStanger:
> I'm sorry for the trouble if that's the case. This is my first time contributing to someone else's repository, and I'm not quite experienced with git. 

No need to apologise. Git is a tricky beast, and this is fairly complex low level stuff. The only way to learn these things is to get in the weeds and try and find your way out, so I'm more than happy to lend a helping hand.

> Do you feel there's a need to add some note to the build documentation on what wayland protocols are needed to enable the focused and launcher modules? Perhaps the clipboard module as well.

I think a little [alert](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts) of the appropriate kind at the top of each applicable module page would be best. Just to say "this module requires a wlroots compositor" or something akin to that. 

---

Thanks for updating - there's a few more bits I've noticed after reviewing if you can have a look at those please. I'll try and do a deeper dive this evening and compile for myself just to verify too.

Once that's all clear, we can have a look at squashing down the commits.

src/config/mod.rs
thmasq: You see, I was trying to rewrite it some other way, but the compiler won't accept that pattern.
I tried this:
```rust
center: 
#[cfg(feature = "focused")]
Some(vec![ModuleConfig::Focused(Box::default())]),
#[cfg(not(feature = "focused"))]
None,
```

But the compiler will give me this error, which I couldn't think of any other way to fix, other than by using those braces
```
error[E0658]: attributes on expressions are experimental
   --> src/config/mod.rs:139:13
    |
139 |             #[cfg(feature = "focused")]
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #15701 <https://github.com/rust-lang/rust/issues/15701> for more information
```
The other way I can think to fix that would be to use the same pattern you were using previously, with `cfg_if`

thmasq: I guess another concise way to do this would be with the cfg macro, which I could put directly in the expression like this:
```rust
center: if cfg!(feature = "focused") {
    Some(vec![ModuleConfig::Focused(Box::default())])
} else {
    None
},
```
I can change it to whichever you think would be best for your coding style

JakeStanger: How strange, I wonder why that doesn't work. You can't use `cfg` like that because it'll still compile in the true-case code, which it what we want to avoid. Instead you can use `cfg_if`, whiich is what I'm already doing for the `end` module just above. I'd say just copy that for now so it matches what's there at least.

thmasq:
I got the changes implemented now. I added notes in the docs as well, and for the sake of consistency, I put a note on the clipboard module to tell the user the module requires the [wlr-data-control](https://gitlab.freedesktop.org/wlroots/wlr-protocols/-/blob/master/unstable/wlr-data-control-unstable-v1.xml) protocol for it to work. I don't think the other modules need this as they only rely on standardized protocols

JakeStanger:
I think that conflict has just been introduced by me merging in other PRs - let me know if you need my help resolving that.  Otherwise, code lgtm now so thanks for that.

The note in the docs I think should clearly say you require a wlroots-based compositor like I had above. It's fine specifying the required protocol in there, but most people won't know much about the underlying protocols and will know if they're on a wlroots compositor.

Then lastly, if you can squash the commits down please. I think a single commit is best now, since both feature flags are kinda tied together:

> build: add `focused` and `launcher` feature flags

Once that'

thmasq:
I think that's all changes implemented. Please review it, and tell me if there's annything that needs to be changed. Also I created another PR by accident, and I'm not sure if there's a way to delete it rather than just closing. I thought that using github's interface to sync the changes would be a good idea, but I was a fooled. Different from using the terminal, on the web interface it's hard to tell which changes are applied to remote before I can review them.

JakeStanger:
Just having a look at the build results, and there are a lot of unused code warnings for code in various places when you run with `--no-default-features` . Some of them I think I'm happy to ignore as it's just a result of having next to no modules in the bar, but there are a couple still that I think would definitely make sense to put behind the feature flags:

```
warning: variant `ToplevelFocus` is never constructed
  --> src/clients/wayland/mod.rs:76:5
```

```
warning: variant `OpenPopupAt` is never constructed
  --> src/modules/mod.rs:76:5
```

If you can amend the commit to resolve those, and the related pattern matches, then I think that should be good enough.

JakeStanger:
> Also I created another PR by accident, and I'm not sure if there's a way to delete it rather than just closing.

I don't think there is, but don't worry about it. It's easy to muck it up, especially when you're using a GUI and you're not sure what it's doing under the hood.

thmasq:
> Just having a look at the build results, and there are a lot of unused code warnings for code in various places when you run with `--no-default-features` . Some of them I think I'm happy to ignore as it's just a result of having next to no modules in the bar, but there are a couple still that I think would definitely make sense to put behind the feature flags:
> 
> ```
> warning: variant `ToplevelFocus` is never constructed
>   --> src/clients/wayland/mod.rs:76:5
> ```
> 
> ```
> warning: variant `OpenPopupAt` is never constructed
>   --> src/modules/mod.rs:76:5
> ```
> 
> If you can amend the commit to resolve those, and the related pattern matches, then I think that should be good enough.

I tested putting these two under some compile flag, and it does work when I compile with absolutely no modules. The problem is that as soon as I needed any module like the clock, or clipboard icons, which need widgets, I get a compile time error.

Example: with `cargo run --release --no-default-features \
  --features config+toml,clock`
  
  I'll get this error message:
  ```
  error[E0599]: no variant or associated item named `ToplevelFocus` found for enum `wayland::Request` in the current scope
   --> src/clients/wayland/mod.rs:338:26
    |
68  | pub enum Request {
    | ---------------- variant or associated item `ToplevelFocus` not found for this enum
...
338 |             Msg(Request::ToplevelFocus(id)) => {
    |                          ^^^^^^^^^^^^^ variant or associated item not found in `Request`

error[E0599]: no variant or associated item named `OpenPopupAt` found for enum `ModuleUpdateEvent` in the current scope
   --> src/modules/mod.rs:323:32
    |
64  | pub enum ModuleUpdateEvent<T: Clone> {
    | ------------------------------------ variant or associated item `OpenPopupAt` not found for this enum
...
323 |             ModuleUpdateEvent::OpenPopupAt(geometry) => {
    |                                ^^^^^^^^^^^
    |                                |
    |                                variant or associated item not found in `ModuleUpdateEvent<_>`
    |                                help: there is a variant with a similar name: `OpenPopup`
    ```
    But as soon as I add the flag I set it to recognize, it will compile. I'm honestly out of ideas to fix this with feature flags, unless I list all the flags on a any block `#[cfg(any())]`

thmasq:
I tried putting those two variant constructors under either the `focused` and `launcher` flags with a cfg(any()), which eliminates the warning when I'm compiling with them disabled, but I still get the same warning when compiling with them. I'm not sure what to think of this

JakeStanger:
I think you're making things more complex than they need to be.

I missed this when I reviewed because of the way GitHub shows only diffs (at least that's my excuse...) but when it comes to the `match` blocks, you can just stick the feature flag on the whole arm. 

So for example, you currently have:

```rust
            Msg(Request::ToplevelFocus(id)) => {
                #[cfg(any(feature = "focused", feature = "launcher"))]
                let handle = env
                    .handles
                    .iter()
                    .find(|handle| handle.info().map_or(false, |info| info.id == id));

                #[cfg(not(any(feature = "focused", feature = "launcher")))]
                let handle: Option<()> = None;

                #[cfg(any(feature = "focused", feature = "launcher"))]
                if let Some(handle) = handle {
                    let seat = env.default_seat();
                    handle.focus(&seat);
                }

                send!(env.response_tx, Response::Ok);
            }
```

Instead, you only need (and want) one feature flag, so that the whole `Msg(Request::ToplevelFocus(id)) ` arm never gets compiled in - this is required because you've stuck a flag on the actual `ToplevelFocus` definition in the enum. If the variant doesn't exist, you can't match against it.

(In this case too, `ToplevelFocus` is only used by `launcher`, so you don't need to compile it in for the `focused` module)

So this can become just:

```rust
        #[cfg(any(feature = "launcher"))]
            Msg(Request::ToplevelFocus(id)) => {
                let handle = env
                    .handles
                    .iter()
                    .find(|handle| handle.info().map_or(false, |info| info.id == id));

                let handle: Option<()> = None;

                if let Some(handle) = handle {
                    let seat = env.default_seat();
                    handle.focus(&seat);
                }

                send!(env.response_tx, Response::Ok);
            }
```

thmasq:
Ok, got it. With this implemented there are considerably fewer warnings than before. Though there's one still pestering me quite a bit, which is this one:
```
warning: variant `ToplevelInfo` is never constructed
  --> src/clients/wayland/mod.rs:94:5
   |
86 | pub enum Response {
   |          -------- variant in this enum
...
94 |     ToplevelInfo(Option<ToplevelInfo>),
   |     ^^^^^^^^^^^^
   |
   = note: `Response` has a derived impl for the trait `Debug`, but this is intentionally ignored during dead code analysis
   = note: `#[warn(dead_code)]` on by default
```
I removed that line of code, and it compiled just fine, so I was wondering what it's for

JakeStanger (approved):

JakeStanger:
> so I was wondering what it's for

Ah, that's a trick one just to confuse you. I added it during some refactoring last week, ended up not using it, then deleted it earlier today. You can ignore that one :)

All lgtm now, thanks for all your work on this!

PR #417, opened 2024-01-24
Various Small Changes

JakeStanger:
docs: add info on speeding up compilation time
      refactor(desktop file): simplify some none-type handling
      docs(examples): fix issues with example css


yavko:
Wouldn't it have been a good idea to also mention sccache?

JakeStanger:
It would have. I even have it set up and was reading the docs as I was writing the docs section 🤦I'll add a section in later

PR #439, opened 2024-02-07
feat: enable use of markup in clock module formats

Dridus:
hello! this PR makes a simple change to allow the use of pango markup in clock formats. the plain `label` module does this, so I figured it wouldn't be out of place.

I read in the contributing guidance to open an issue or discussion as best practice, but since this is so small and I wanted it for myself I figured I'd just go straight to PR and I can close or revise it as you like.

thanks for writing ironbar!

### DIFF
--- a/docs/modules/Clock.md
+++ b/docs/modules/Clock.md
@@ -10,8 +10,8 @@ Clicking on the widget opens a popup with the time and a calendar.
 
 | Name           | Type     | Default                            | Description                                                                         |
 |----------------|----------|------------------------------------|-------------------------------------------------------------------------------------|
-| `format`       | `string` | `%d/%m/%Y %H:%M`                   | Date/time format string.                                                            |
-| `format_popup` | `string` | `%H:%M:%S`                         | Date/time format string to display in the popup header.                             |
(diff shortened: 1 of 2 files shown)


JakeStanger:
Thanks for the PR, agreed this makes sense to enable. 

 That particular contribution guideline is more for your own benefit, so that you don't spend time developing something that ultimately gets rejected. I appreciate you reading and otherwise sticking to the guidelines though :)

A  couple of small changes please:

- Since the clock module also offers `format_popup`, could you also enable pango there to keep things consistent?
- Could you please update the clock module docs page to mention that markup is supported in each row of the config table?

I've got 0.14.1 hopefully going out this weekend, which'll be a bugfix only update. This can then follow after that and will likely be the first feature of 0.15.0.

Dridus:
@JakeStanger okay, those changes should be made. I also made the `format` one match the `format_popup` one in terms of using `LabelBuilder`, to be consistent. if it should be consistently not using builder or inconsistent though let me know.

JakeStanger (approved):

Dridus:
I realized that I flubbed the squash somehow, so I corrected it in the second force-push.

JakeStanger:
First feature of 0.15, cheers for your work on this :)

PR #441, opened 2024-02-10
feat(upower): add new formatting properties

### LINKED ISSUE #440: upower widget doesn't parse {time_to_empty}
**Describe the bug**
The upower widget doesn't seem to format any other string besides {percentage}

**To Reproduce**
1. Change config.json module's format to show any property other than percentage. e.g {time_to_full}
2. Launch ironbar
3. See that string isn't fully formatted. e.g {time_to_empty} appears as a plain string

**Expected behavior**
The formatting is applied and {time_to_empty} and other properties become actual values

**System information:**
 - Distro: [e.g. Arch Linux, Ubuntu 22.10]
- Compositor: HyprLand
 - Ironbar version: 0.14.1-pre 

**Configuration**
```
{
    "monitors": {
        "eDP-1": {
            "position": "bottom",
            "height": 30,
            "start" : [

            ],
            "center": [

            ],
            "end": [
                {
                    "type": "upower",
                    "format": "{percentage}% {icon_name} {state} {time_to_full} {time_to_empty}"
                }
            ]
        }
    }
}
```

**Screenshots**
image

Disr0:
Adds parsing in format for properties other than {percentages} :face_with_spiral_eyes: 

Resolves #440

### DIFF
--- a/src/modules/upower.rs
+++ b/src/modules/upower.rs
@@ -15,7 +15,7 @@ use crate::modules::PopupButton;
 use crate::modules::{
     Module, ModuleInfo, ModuleParts, ModulePopup, ModuleUpdateEvent, WidgetContext,
 };
-use crate::{error, glib_recv, send_async, spawn, try_send};
+use crate::{glib_recv, send_async, spawn, try_send};
 
 const DAY: i64 = 24 * 60 * 60;
 const HOUR: i64 = 60 * 60;
@@ -45,7 +45,7 @@ const fn default_icon_size() -> i32 {
 pub struct UpowerProperties {
     percentage: f64,
     icon_name: String,
-    state: u32,
+    state: BatteryState,
     time_to_full: i64,
     time_to_empty: i64,
 }
@@ -84,9 +84,12 @@ impl Module<gtk::Button> for UpowerModule {
                 .downcast_ref::<str>()
                 .expect("expected IconName: str in HashMap of all properties")
                 .to_string();
-            let state = *properties["State"]
-                .downcast_ref::<u32>()
-                .expect("expected State: u32 in HashMap of all properties");
+            let state = u32_to_battery_state(
+                *properties["State"]
+                    .downcast_ref::<u32>()
+                    .expect("expected State: u32 in HashMap of all properties"),
+            )
+            .unwrap_or(BatteryState::Unknown);
             let time_to_full = *properties["TimeToFull"]
                 .downcast_ref::<i64>()
                 .expect("expected TimeToFull: i64 in HashMap of all properties");
@@ -123,9 +126,9 @@ impl Module<gtk::Button> for UpowerModule {
                                 .to_string();
                         }
                         "State" => {
-                            properties.state = changed_value
-                                .downcast::<u32>()
-                                .expect("expected State to be u32");
+                            properties.state =
+                                u32_to_battery_state(changed_value.downcast::<u32>().unwrap_or(0))
+                                    .expect("expected State to be BatteryState");
                         }
                         "TimeToFull" => {
                             properties.time_to_full = changed_value
@@ -185,7 +188,17 @@ impl Module<gtk::Button> for UpowerModule {
 
         let rx = context.subscribe();
         glib_recv!(rx, properties => {
-            let format = format.replace("{percentage}", &properties.percentage.to_string());
+            let state = properties.state;
+            let is_charging = state == BatteryState::Charging || state == BatteryState::PendingCharge;
+            let time_remaining = if is_charging {
+                seconds_to_string(properties.time_to_full)
+            }
+            else {
+                seconds_to_string(properties.time_to_empty)
+            };
+            let format = format.replace("{percentage}", &properties.percentage.to_string())
+                .replace("{time_remaining}", &time_remaining)
+                .replace("{state}", battery_state_to_string(state));
             let icon_name = String::from("icon:") + &properties.icon_name;
 
             ImageProvider::parse(&icon_name, &icon_theme, false, self.icon_size)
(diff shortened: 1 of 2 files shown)


JakeStanger:
Thanks for the PR. This is definitely a good addition. There are a few small bits that I think need addressing before this goes in:

This is a new feature, not a bug fix :) Could you amend your commit message to reflect that please so it gets added to the changelog correctly. Something like: `feat(upower): add new formatting properties`.

I don't think adding `icon_name` as a property is particularly helpful, unless there's something smart you could do with this and Pango? Otherwise it's sort of an internal identifier that gets exposed with little purpose.

Arguably the `time_to_empty` and `time_to_full` properties should be combined into one (perhaps just `time_remaining`?), which checks the state to see which underlying property to use. Otherwise you end up having to add both to have the ETA consistently display:

```corn
{ type = "upower" format = "{percentage}% {state} ({time_to_empty}{time_to_full})" }
```

I'm not sure whether the combine property should be instead of or in addition to the separated ones? Would people ever want to show one and not the other? I'm open to feedback on that one.

Once that's all done I'll re-review and update the docs page. Let me know if you need any help with any of the above :)

JakeStanger (changes requested):
[Changes requested above]

Disr0:
Yeah, ```icon_name``` isn't particularly useful, I'll remove it from the formatting.
Combining `time_to_empty` and `time_to_full` sounds more correct, yeah, but how should it be done? I'm not sure if adding `if` switch is a good idea, but maybe it's better than just concatenating `time_to_empty` and `time_to_full` as a string and sending it as `time_remaining`?

src/modules/upower.rs
```diff
@@ -290,3 +296,16 @@ const fn u32_to_battery_state(number: u32) -> Result<BatteryState, u32> {
...
+fn battery_state_to_string(state: u32) -> &'static str {
+    match state {
+        0 => "Unknown",
+        1 => "Charging",
+        2 => "Discharging",
+        3 => "Empty",
+        4 => "Fully charged",
+        5 => "Pending charge",
+        6 => "Pending discharge",
+        7_u32..=u32::MAX => "Unknown",
```
yavko, line 309: Shouldn't this be `_ => "Unknown"` ?

Disr0: I based this on the enum description from here
https://docs.rs/upower_dbus/latest/upower_dbus/enum.BatteryState.html

JakeStanger: Yeah good spot. It doesn't actually make any difference here, but the `7_u32..=u32::MAX` case could just be replaced with a catch-all `_` as it covers all other possibilities.

Disr0: Ah, I see

yavko: > I based this on the enum description from here [docs.rs/upower_dbus/latest/upower_dbus/enum.BatteryState.html](https://docs.rs/upower_dbus/latest/upower_dbus/enum.BatteryState.html)

if thats the case you don't want to cast as an int:
```rs
match state {
        BatteryState::Unkown => "Unknown",
        BatteryState::Charging => "Charging",
        BatteryState::Discharging => "Discharging",
        BatteryState::Empty => "Empty",
        BatteryState::FullyCharged => "Fully charged",
        BatteryState::PendingCharge => "Pending charge",
        BatteryState::PedingDischarge => "Pending discharge",
        _ => "Unknown",
}
```

yavko: This is so you get type checking if a new variant begins to exist

JakeStanger: The client returns a `u32` and that gets mapped onto the `BatteryState` on an ad-hoc basis, so the casting is always `u32` -> `BatteryState`. Really this is a problem at the controller & `UPowerProperties` struct definition level. It might make sense to refactor that.

yavko: Yeah, I see that now, https://github.com/JakeStanger/ironbar/pull/441/files#diff-6fb3d399ffa41e0be3e32a6faac7a73913a6f5a9759c6f6d28dd0345348a9e17R48 needs to be changed

src/modules/upower.rs
```diff
@@ -185,7 +188,18 @@ impl Module<gtk::Button> for UpowerModule {
...
-            let format = format.replace("{percentage}", &properties.percentage.to_string());
+            let state = properties.state;
+            let is_charging = state == BatteryState::Charging || state == BatteryState::PendingCharge;
+            let time_remaining;
+            if is_charging {
+                time_remaining = seconds_to_string(properties.time_to_full);
+            }
+            else {
+                time_remaining = seconds_to_string(properties.time_to_empty);
+            }
```
JakeStanger, line 199: In Rust `if else` is actually an expression, so you can use it directly for assignment:

```suggestion
            let time_remaining = if is_charging {
                seconds_to_string(properties.time_to_full);
            }
            else {
                seconds_to_string(properties.time_to_empty);
            }
```

(I don't think that suggestion is `rustfmt` compliant so you'll probs need to run after applying)

JakeStanger (changes requested):
Looking nearly there, cheers for your work on this. Just one minor adjustment I think, then I'm happy with it. 

Are you able to squash your commits down into a single one which matches the PR title? I can give a hand with that if not.

src/modules/upower.rs
Disr0: Sadly this doesn't work, as then the `time_remaining` variable will be of type `&()`

JakeStanger: My bad - the semicolons make the difference there due to Rust's implicit returns. Updated.

Disr0: Implemented the change.

JakeStanger (approved):

JakeStanger:
Cheers, lgtm. I'll update the docs to add in the relevant section & merge in the next few days.

PR #449, opened 2024-02-13
Add `direction` option for tray module

calops:
Add the `direction` option for the tray module, giving the ability for vertical trays.
image

**Note:** This also sets the direction `top_to_bottom` *by default* for vertical bars. This is technically a **breaking change**, but I imagine nobody actually wants a horizontal tray on a vertical bar.

### DIFF
--- a/docs/modules/Tray.md
+++ b/docs/modules/Tray.md
@@ -6,7 +6,10 @@ Displays a fully interactive icon tray using the KDE `libappindicator` protocol.
 
 > Type: `tray`
 
-***This module provides no configuration options.***
+
+| Name        | Type     | Default                                                         | Description                                                                                                              |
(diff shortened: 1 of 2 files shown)


yavko:
> Add the `orientation` option for the tray module, giving the ability for vertical trays. image
> 
> **Note:** This also sets the direction `top_to_bottom` _by default_ for vertical bars. This is technically a **breaking change**, but I imagine nobody actually wants a horizontal tray on a vertical bar.

if it's breaking, don't you have to do `!` with conventional commits, eg. `feat(tray)!: add orientation option`

calops:
> if it's breaking, don't you have to do `!` with conventional commits, eg. `feat(tray)!: add orientation option`

Absolutely, this slipped my mind. Let me amend that real quick.

calops:
Yeah that's fair enough, I'm fine with considering that part a bugfix. I'll amend this later today.

PR #450, opened 2024-02-13
docs: correct formatting token time_remaining in upower module

Disr0:
Docs have a type: the token is actually `time_remaining` and not `remaining`
Thought it'd be pretty annoying for users to not see the intended behavior


JakeStanger:
Cheers!

PR #460, opened 2024-02-18
Volume module

### LINKED ISSUE #50: Volume module
A volume module with a popup widget that has a slider. option to choose backend (wpctl, pactl, etc). usage of on-scroll events to change volume without opening the widget.

JakeStanger:
The volume module is now complete and stable enough that I believe it warrants opening a PR. The code is a bit rough in places so I'm going to refactor it over the next couple of weeks before it gets merged in. This hopefully also gives a chance for people to have a poke at it.

The initial version of the module provides the following:

- A button on the bar which indicates volume level by icon
- A popup with controls for output device switching, output device volume changing, and individual output stream (ie per program) level controls.
- Support for Pulseaudio (inc pipewire-pulse).

To do still:

- [x] Fix a few edge cases (such as popup state when nothing is playing)
- [x] Add `name` and `class` elements to widgets
- [x] Ability to show more on the bar (such as current vol level)
- [x] More options
- [x] Write documentation
- [x] Update example config & styles

Resolves #50

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,7 +34,7 @@ jobs:
       - name: Install build deps
         run: |
           sudo apt-get update
-          sudo apt-get install libgtk-3-dev libgtk-layer-shell-dev
+          sudo apt-get install libgtk-3-dev libgtk-layer-shell-dev libpulse-dev
 
       - name: Clippy
         run: cargo clippy --no-default-features --features config+json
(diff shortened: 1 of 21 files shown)


slowsage:
I'm thinking of adding an audio device switcher using pactl subscribe to track events. Will wait for 1st cut of volume module to be merged. I want to add a way to show current audio device (headphone/speaker etc, next to volume icon.) with a way to customize the icons shown based on regex.

JakeStanger:
It sounds like this may be most of the way to that already. The popup supports device switching, and you can display the current device name next to the icon on the bar. 

The only part of that is missing is regex support. Generally I'm a little cautious of adding regex in as it can be a bit heavy-handed, and less intuitive to configure. That doesn't mean it's out the question, but I think it'd be worth opening an issue to discuss what you're after before jumping into anything.

PR #466, opened 2024-02-21
Added support for using a cover image which has a specific static

eternalfrustation:
Added support for using a cover image which has a specific static path in the music module

### DIFF
--- a/src/modules/music/config.rs
+++ b/src/modules/music/config.rs
@@ -97,6 +97,8 @@ pub struct MusicModule {
     #[serde(default = "default_cover_image_size")]
     pub(crate) cover_image_size: i32,
 
+    pub(crate) cover_image_path: Option<PathBuf>,
+
     // -- Common --
     pub(crate) truncate: Option<TruncateMode>,
 
--- a/src/modules/music/mod.rs
+++ b/src/modules/music/mod.rs
@@ -404,91 +404,103 @@ impl Module<Button> for MusicModule {
 
             let mut prev_cover = None;
             glib_recv!(rx, event =>  {
-                match event {
-                    ControllerEvent::Update(Some(update)) => {
-                        // only update art when album changes
-                        let new_cover = update.song.cover_path;
-                        if prev_cover != new_cover {
-                            prev_cover = new_cover.clone();
-                            let res = if let Some(image) = new_cover.and_then(|cover_path| {
-                                ImageProvider::parse(&cover_path, &icon_theme, false, image_size)
-                            }) {
-                                album_image.show();
-                                image.load_into_image(album_image.clone())
-                            } else {
-                                album_image.set_from_pixbuf(None);
-                                album_image.hide();
-                                Ok(())
+                            match event {
+                                ControllerEvent::Update(Some(update)) => {
+                                    // only update art when album changes
+                                    dbg!("{:#?}", self.cover_image_path.clone());
+                                    if let Some(ref static_cover_image_path) = self.cover_image_path {
+                                        match ImageProvider::parse(static_cover_image_path.to_str().unwrap(), &icon_theme,false,image_size) {
+                                            Some(image) => {
+                                                match image.load_into_image(album_image.clone()) {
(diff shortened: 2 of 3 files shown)


JakeStanger:
Hey, thanks for the PR. I'm wondering what the reasoning/use case for this is?

I'm away this week, so will review properly when I'm back.

eternalfrustation:
The main purpose for this was that i was using mpd-notification for notifications and writing the images that were extracted from the media files to a tmp directory.
I thought about using scripts to update the album cover in iron bar, but found that the cover path is hardcoded.

eternalfrustation:
> It seems your last commit is unrelated to the pull request

Yea, i am still new to contributing to other people's projects, I did want to make it a separate PR, but i didn't realise that pushing to the same repo would append the commit to the PR

yavko:
> > It seems your last commit is unrelated to the pull request
> 
> Yea, i am still new to contributing to other people's projects, I did want to make it a separate PR, but i didn't realise that pushing to the same repo would append the commit to the PR

Here's some info that may or may not be useful

You have to create separate branches, ur contributing from master, which usually you don't do. What you want to do when contributing to a repo is:
1. fork & clone the repo
2. make a new branch `git branch <branch-name>`
3. switch to it `git checkout <branch-name>`
4. make ur changes
5. stage ur changes `git add <files>`
6. commit ur changes `git commit -m <msg>`
7. push ur changes to ur github fork `git push` & make a PR on the github website

Now, for other changes you can make a new branch, so you don't push those changes to unrelated PRs.
To fix ur current issue, you'll probably have to stash the last commit somehow, also amend ur previous commit to conventional commits (cuz this repo uses it), then force push. Then you can make a new branch (and switch to it) unstash ur unrelated commit, and you can create a second PR.

yavko:
> The main purpose for this was that i was using mpd-notification for notifications and writing the images that were extracted from the media files to a tmp directory. I thought about using scripts to update the album cover in iron bar, but found that the cover path is hardcoded.

I'm pretty sure the cover image changes automatically in ironbar, why would you want to do it manually?

JakeStanger:
> I'm pretty sure the cover image changes automatically in ironbar 

The current behaviour with MPD is that it looks for album art in the same folder as the current playing track. You need to specify the music directly MPD uses in your config as MPD only provides paths relative to this.

I think if this isn't catching every scenario reliably, it makes more sense to try and expand the logic around that than introduce new options, especially if they require entirely manually scripting the update logic.

eternalfrustation:
> Now, for other changes you can make a new branch, so you don't push those changes to unrelated PRs.
> To fix ur current issue, you'll probably have to stash the last commit somehow, also amend ur previous commit to conventional commits (cuz this repo uses it), then force push. Then you can make a new branch (and switch to it) unstash ur unrelated commit, and you can create a second PR.
Thanks for the info, i will try to do this as soon as i have the time.

> I think if this isn't catching every scenario reliably, it makes more sense to try and expand the logic around that than introduce new options, especially if they require entirely manually scripting the update logic.

True, but unfortunately, as far as i can see,  the only other option would be to try to extract the cover image from the audio file itself, or allow executing a script which tells us the path to the cover image

eternalfrustation:
I just thought that just adding an optional field would be the simplest solution which would cover the most amount of usecases

JakeStanger:
> as far as i can see, the only other option would be to try to extract the cover image from the audio file itself, or allow executing a script which tells us the path to the cover image

I think both of these are much more preferable to the change proposed here, especially the first one. Although more complex to implement, they both offer a solution which is simpler to the end-user, as they remove the need for them to configure anything extra or write any custom scripts. MPD even offers a `readpicture` command since 0.22 which can read the art from a file.

The other included commit for sys-info is also not something I want to merge at the moment. Currently I am targeting a rework of the internals of that module for 0.16 which will include the ability to specify units.

Based on that I'm going to close this for now I'm afraid. Feel free to open an issue to discuss further, or open a PR in future that better suits the requirements.

Issue #228, opened 2023-07-06
[Launcher] Incorrect icons being used


donovanglover:
**Describe the bug**
For example, if I add an `a.b.c`  favorite I get a librewolf icon. [Memento](https://ripose-jp.github.io/Memento/) uses an htop icon, which I assume is because the class name is `projects.ripose.`. Any programs that have the class name `.program-wrapped` appear to be broken as well.

**To Reproduce**
1. Open `memento` or use a random string of characters as a favorite

**Expected behavior**
Incorrect icons should never be visible to the user. How about an icon indicating that none was found instead?

**System information:**
 - Distro: NixOS
- Compositor: Hyprland
 - Ironbar version: https://github.com/JakeStanger/ironbar/commit/abf490b04eda87841639c29893dda98408805a3d

**Additional context**
The `memento.desktop` file has an `Icon=memento` field. How feasible is it to use this instead of the app_id?

Would it be better to let users specify their own class name and icon overrides? Should these overrides be mapped to desktop files?

JakeStanger:
>  The memento.desktop file has an Icon=memento field. How feasible is it to use this instead of the app_id?

This *should* be happening if a reasonable match isn't found by filename. I suspect filenames are currently matching a bit too eagerly, and I'll need to tone this down to resolve this.

Issue #146, opened 2023-05-08
Focused and Launcher modules attempt to render unknown image types.


A-Cloud-Ninja:
**Describe the bug**
Possibly due to flatpak'd versions of Spotify and Discord (and otherwise icon-missing applications), apps that don't properly have an icon do not get verified, and an error is printed.
**To Reproduce**
1. Use launcher or focused with show_icon
2. Have an app that either doesn't have an icon or otherwise isn't in the correct location
3. Launch ironbar, observe errors upon loading icons for the app.

**Expected behavior**
Either a warning and/or discarding or sanitizing of related items

**System information:**
 - Distro: Arch
- Compositor: Hyprland
 - Ironbar version: `Ironbar version 0.12.1` via `ironbar --version` on `aur/ironbar-git`

**Configuration**
https://f.cloudninja.pw/ironbar_config.json

**Screenshot**
![SuccinctIncompetentKoala](https://user-images.githubusercontent.com/5809177/236894723-ab739d9c-fbf1-490e-b0c4-a046ae755309.png)

JakeStanger:
Will improve the error handling around this to try and rule out some of the non-errors, reduce missing images to warnings, and avoid printing the full stack.

TheSunCat:
When focusing an app for which it cannot find an icon (though Anyrun does find the icon), ironbar does indeed just print a warning, however it still renders the icon for the last focused app.

For example, if I move focus from Firefox to Supersonic (AUR package `supersonic-desktop`), ironbar fails to find Supersonic's icon and instead renders Firefox's icon.

JakeStanger:
Thanks for reporting. I've opened an issue to track that and should get it sorted soon.

PR #476, opened 2024-02-26
feat(launcher): item.name as fallback for app icon

slowsage:
When using spotify in wayland with hyprland, the spotify icon is not pulled up correctly by the launcher module:
`spotify --enable-features=UseOzonePlatform,WaylandWindowDecorations --ozone-platform=wayland`
The class/app_id is "" while the name/title is "Spotify Premium". This uses item.name as a fallback which ensure spotify icon shows up correctly under wayland. Spotify under X11 does not have this problem.

Refs: #228, #146

### DIFF
--- a/src/modules/launcher/item.rs
+++ b/src/modules/launcher/item.rs
@@ -166,8 +166,12 @@ impl ItemButton {
 
         if appearance.show_icons {
             let gtk_image = gtk::Image::new();
-            let image =
-                ImageProvider::parse(&item.app_id.clone(), icon_theme, true, appearance.icon_size);
+            let input = if item.app_id.is_empty() {
+                item.name.clone()
+            } else {
+                item.app_id.clone()
+            };
+            let image = ImageProvider::parse(&input, icon_theme, true, appearance.icon_size);
             if let Some(image) = image {
                 button.set_image(Some(&gtk_image));
                 button.set_always_show_image(true);


JakeStanger:
Cheers for the PR, seems like a good fix

src/modules/launcher/item.rs
```diff
@@ -166,8 +166,12 @@ impl ItemButton {
 
         if appearance.show_icons {
             let gtk_image = gtk::Image::new();
-            let image =
-                ImageProvider::parse(&item.app_id.clone(), icon_theme, true, appearance.icon_size);
+            let input = if item.app_id.is_empty() {
+                item.name.clone()
+            } else {
+                item.app_id.clone()
+            };
```
slowsage, line 173: I tried changed your suggestion by changing `cloned()` to `clone()` but I get:
```
no method named `unwrap_or_else` found for struct `std::string::String` in the current scope
method not found in `String`
```
Best I seem to be able to do (without failing clippy) is:

```suggestion
            let input = item
                .app_id
                .is_empty()
                .then(|| item.name.clone())
                .unwrap_or_else(|| item.app_id.clone());
```

I think the if is clearer but will defer to whatever you prefer - not familiar with the rest of the code. The issue is app_id is not None but the empty string in this case.

JakeStanger: Oh sorry, I misread it and thought it was an `Option` not an empty string. In that case, I think it's best as you originally had it.

JakeStanger (approved):
All LGTM, I'll try and give this a test in the next couple of days then merge

PR #481, opened 2024-03-04
Fix(music): mpris - handle no active mpris player on launch & hide label when player list is empty.

slowsage:
Sometimes, I see the NoActivePlayer DBusError when no music players are running.

```
2024-03-03T21:41:53.536515Z ERROR ironbar::logging: 48: The application panicked (crashed).
Message:  Failed to connect to D-Bus: DBusError(TransportError(D-Bus error: No player is being controlled by playerctld (com.github.altdesktop.playerctld.NoActivePlayer)))
Location: src/clients/music/mpris.rs:40
```

The first commit prevents the panic so the song shows when the player is started after ironbar was launched.

The second commit checks if all the players have been cleared out and sends None to hide the label in this case. With this change, the music label does not stay after the app is closed.

I tested by
1.start ironbar
2. then spotify,
3. then another player
4. and closed both
5. and restarted spotify.

The show/hide behavior is as expected: 1. none, 2. song1, 3. song2, 4. song2->song1->None, 5. song1

It's entirely possible I missed the actual mechanism to hide the label when no players active. Please let me know if there is a better way. Thank you!

### DIFF
--- a/src/clients/music/mpris.rs
+++ b/src/clients/music/mpris.rs
@@ -18,6 +18,11 @@ pub struct Client {
     _rx: broadcast::Receiver<PlayerUpdate>,
 }
 
+const NO_ACTIVE_PLAYER: &str = "com.github.altdesktop.playerctld.NoActivePlayer";
+const NO_REPLY: &str = "org.freedesktop.DBus.Error.NoReply";
+const NO_SERVICE: &str = "org.freedesktop.DBus.Error.ServiceUnknown";
+const NO_METHOD: &str = "org.freedesktop.DBus.Error.UnknownMethod";
+
 impl Client {
     pub(crate) fn new() -> Self {
         let (tx, rx) = broadcast::channel(32);
@@ -35,44 +40,48 @@ impl Client {
                 // D-Bus gives no event for new players,
                 // so we have to keep polling the player list
                 loop {
-                    let players = player_finder
-                        .find_all()
-                        .expect("Failed to connect to D-Bus");
+                    // mpris-rs does not filter NoActivePlayer errors, so we have to do it ourselves
+                    let players = player_finder.find_all().unwrap_or_else(|e| match e {
+                        mpris::FindingError::DBusError(DBusError::TransportError(
+                            transport_error,
+                        )) if transport_error.name() == Some(NO_ACTIVE_PLAYER)
+                            || transport_error.name() == Some(NO_REPLY) =>
+                        {
+                            Vec::new()
+                        }
+                        _ => panic!("Failed to connect to D-Bus"),
+                    });
+                    // Acquire the lock of current_player before players to avoid deadlock.
+                    // There are places where we lock on current_player and players, but we always lock on current_player first.
+                    // This is because we almost never need to lock on players without locking on current_player.
+                    {
+                        let mut current_player_lock = lock!(current_player);
 
-                    let mut players_list_val = lock!(players_list);
-                    for player in players {
-                        let identity = player.identity();
+                        let mut players_list_val = lock!(players_list);
+                        for player in players {
+                            let identity = player.identity();
 
-                        if !players_list_val.contains(identity) {
-                            debug!("Adding MPRIS player '{identity}'");
-                            players_list_val.insert(identity.to_string());
+                            if current_player_lock.is_none() {
+                                debug!("Setting active player to '{identity}'");
+                                current_player_lock.replace(identity.to_string());
 
-                            let status = player
-                                .get_playback_status()
-                                .expect("Failed to connect to D-Bus");
-
-                            {
-                                let mut current_player = lock!(current_player);
-
-                                if status == PlaybackStatus::Playing || current_player.is_none() {
-                                    debug!("Setting active player to '{identity}'");
-
-                                    current_player.replace(identity.to_string());
-                                    if let Err(err) = Self::send_update(&player, &tx) {
-                                        error!("{err:?}");
-                                    }
+                                if let Err(err) = Self::send_update(&player, &tx) {
+                                    error!("{err:?}");
                                 }
                             }
+                            if !players_list_val.contains(identity) {
+                                debug!("Adding MPRIS player '{identity}'");
+                                players_list_val.insert(identity.to_string());
 
-                            Self::listen_player_events(
-                                identity.to_string(),
-                                players_list.clone(),
-                                current_player.clone(),
-                                tx.clone(),
-                            );
+                                Self::listen_player_events(
+                                    identity.to_string(),
+                                    players_list.clone(),
+                                    current_player.clone(),
+                                    tx.clone(),
+                                );
+                            }
                         }
                     }
-
                     // wait 1 second before re-checking players
                     sleep(Duration::from_secs(1));
                 }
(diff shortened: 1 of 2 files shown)


slowsage:
Addendum: It's entirely possible I missed the actual mechanism to hide the label when no players active. Please let me know if there is a better way. Thank you!

slowsage:
Change made. Thank you!

slowsage:
Please hold off on committing it. There might be an issue. Want to test for a few days more. Thank you. I will comment here once confirmed.

slowsage:
Ready for review. Added more checks as I found them.

Possible issue: I removed the status playing check on new player added since it complicated the code. It seems to work fine without it since the active player gets set on the playing/progress event. But this might be a function of my app. Can add it back if you prefer.

Apart from the above issue, it seems to handle all edge cases on my setup with at most a second of delay after a player is terminated.

Please take a look and possibly even run this for a few days/weeks if you have time. Do not want to rush this since user environments vary and concurrency is involved.

Thank you!!!

src/clients/music/mpris.rs
```diff
@@ -46,33 +58,26 @@ impl Client {
...
-                            );
                         }
-                    }
 
+                        Self::listen_player_events(
+                            identity.to_string(),
+                            players_list.clone(),
+                            current_player.clone(),
+                            tx.clone(),
+                        );
```
slowsage, line 79: Can this be outside the block with the current_player lock? Not sure why it was inside.

src/clients/music/mpris.rs
```diff
@@ -35,34 +39,36 @@ impl Client {
...
-                        .expect("Failed to connect to D-Bus");
+                    // mpris-rs does not filter NoActivePlayer errors, so we have to do it ourselves
+                    let players = player_finder.find_all().unwrap_or_else(|e| match e {
+                        mpris::FindingError::DBusError(DBusError::TransportError(
+                            transport_error,
+                        )) if transport_error.name() == Some(NO_ACTIVE_PLAYER)
+                            || transport_error.name() == Some(NO_REPLY) =>
+                        {
+                            Vec::new()
+                        }
```
JakeStanger, line 50: This can be squashed down a bit:

```suggestion
                        )) if transport_error.name() == Some(NO_ACTIVE_PLAYER)
                            || transport_error.name() == Some(NO_REPLY) => vec![]
```

JakeStanger (approved):
Only 1 tiny nit if you get a chance, but otherwise lgtm. Thanks for your work on this!

I'll give this a few tests myself over the weekend, and if all looks good I'll merge. If anybody following master does find any bugs we can get it sorted before next release.

src/clients/music/mpris.rs
slowsage: This was a bit finicky. Tried you suggestion and other changes for a bit but cargo fmt keeps adding the extra lines with {}. Tried shortening the previous line using:
```
use mpris::EventError::DBusError as EventDBusError;
use mpris::FindingError::DBusError as FindingDBusError;
```
but wont pass cargo fmt unless I make it something like EDE and FDE for the above errors which sacrifices clarity for brevity.
Please lmk if you have something that passes cargo fmt. Thank you!

JakeStanger: Ah fair enough, I wonder if it's just because of the pattern complexity it demands a block. I'm not particularly bothered by that either way so happy to mark resolved.

slowsage:
Apologies. Just ran into one more edge case.

I can get back after a month of testing. If you want to close this till then, please do so.

JakeStanger:
No rush, take your time. You can convert this to a draft if you want, but either way just give me a ping when you're happy with it.

slowsage:
This is fairly stable (crossing fingers). Biggest new addition: made the lock acquisition order consistent with your earlier shutdown of player (current_player, then players). Please take a look at the locking code changes.
The only possible downgrade in behavior I can think of is the following sequence:

- player1.play() [shows player1]
- player2.play() [shows player2]
- player2.pause() [shows player2 while player 1 is playing].

It switches to player1 if you either
1. pause and play player1 
2. close player2.

Either way, I think its minor and can be a separate pull request once this is stable and merged in.

src/clients/music/mpris.rs
```diff
@@ -35,35 +40,38 @@ impl Client {
...
+                    // This is because we almost never need to lock on players without locking on current_player.
+                    let mut current_player_lock = lock!(current_player);
 
                     let mut players_list_val = lock!(players_list);
                     for player in players {
                         let identity = player.identity();
 
+                        if current_player_lock.is_none() {
+                            debug!("Setting active player to '{identity}'");
+                            current_player_lock.replace(identity.to_string());
```
JakeStanger, line 65: It's better practise to hold a lock for only as long as you need it to minimise the risk of deadlock, eg:

```rs
lock!(current_player).replace(identity.to_string());
```

Does this work here and remove the need for the 'special' lock above?

slowsage: I think doing it that way can cause a deadlock.

True, it will hold the lock for longer now, but it will eventually be released. I was trying to avoid something like this:
https://www.ibm.com/docs/en/configurepricequote/10.0?topic=caching-sort-order-deadlocks.

It should not deadlock if we keep the order the same, i.e. current_player before players - which we need to do in listen_player_events where we always need to lock on current_player but we need to lock players only when a player crashes.
Another way to work around it would be to only hold one lock at a time but that seems more complex.
Please let me know if you disagree. Thanks!

JakeStanger: In that case let's not risk it, and leave it as it is because that's been tested, cheers for confirming

JakeStanger:
Changes LGTM, thanks. Agreed that 'downgrade' is very minor, so happy to let that slide for now.

I'll give this a test over the next couple of days myself then merge.

JakeStanger:
I have just given this a test through, and I've found a regression: there's a second delay between player some updates (such as play/pause) and Ironbar reflecting that, even if toggled through the Ironbar popup.

There is also a pre-existing bug I've found where if you close a player it remains the 'active' one, then trying to play it throws an error.

slowsage:
> I have just given this a test through, and I've found a regression: there's a second delay between player some updates (such as play/pause) and Ironbar reflecting that, even if toggled through the Ironbar popup.

Verified. Missed this regression. Thank you for noticing this. Will figure out why and get back.

> 
> There is also a pre-existing bug I've found where if you close a player it remains the 'active' one, then trying to play it throws an error. I'm not sure how easy this is to fix, but it might be worth including a fix here as it's in scope. If not, I'll log a separate bug for it.

I cannot reproduce this. Which player? My guess is yet another error is being thrown in the `Err(err) => error!("{err:?}"),` match in `listen_player_events` that should be caught and handle_shutdown should be called in that case as well. Do you have a log of any errors?

JakeStanger:
> I cannot reproduce this. Which player?

I tested with Ryhthmbox:

1. Play video on YouTube via Firefox
2. Pause video
3. Open Rhythmbox
4. Start playing a song from Rhythmbox
5. Pause Rhythmbox
6. Close Rhythmbox
7. MPRIS widget continues to show Rhythmbox as player. Clicking play from widget causes error.

JakeStanger:
I think you're right, I was a commit behind. Sorry about that! I have just tested again on the latest commit and everything seems solid.

The commits at the moment are possibly a bit technical and separated - do you think you could squash them down to reflect each bug please, as each commit will form an entry in the changelog? Once that's done I think it's merge time :)

JakeStanger:
Cheers for all your work on this!

PR #505, opened 2024-03-20
Native modules inside custom modules

### LINKED ISSUE #131: Support for normal module types in custom modules.

JakeStanger:
Adds ability to put native modules inside custom modules.

This works apart from a few (relatively) edge case. The code also needs refactoring.

Sticking this up for now so it's available - will revisit to get across the line in the near future.

Resolves #131.


JakeStanger:
Edge cases sorted. I want to spend some more time polishing the code & update the docs then I think this is ready.

Issue #511, opened 2024-03-27
Setting icons at the widget level


nyadiia:
**Is your feature request related to a problem? Please describe.**
The battery widget icon currently blends into my bar. I'd like to be able to change the icon(s) without changing the whole bar's icon theme.

**Describe the solution you'd like**
Some way to set an icon per widget, like how you're able to set a [ramp in polybar](https://github.com/polybar/polybar/wiki/Module:-battery#additional-formatting) or how you're able to [set multiple icons in the format in waybar](https://github.com/Alexays/Waybar/wiki/Module:-Battery#example).

**Describe alternatives you've considered**
I could just switch the bar's icon theme however I would like to use a different theme per widget. Doing it with text as well ( like polybar and waybar ) would allow it to have the same color as the font for a more cohesive look.

JakeStanger:
All icons should be customisable, ideally with support for both images and textual icons. This is inconsistent at the moment and I know I've made it worse with the introduction of the volume and notification modules. Because of that I'll look to address this as part of the next update.

JakeStanger:
Moving this back to 0.16.0 because I've been working on a library to assist with both this and the upcoming sysinfo changes, so I think it makes sense to tackle both at once. 

This will be towards the top of the pile in the next update.

Issue #148, opened 2023-05-09
Network module


JakeStanger:
**Is your feature request related to a problem? Please describe.**
Currently this is missing functionality that has to be scripted. 

**Describe the solution you'd like**
A new module for managing network connections that connects to the `NetworkManager` service. This should primarily focus on WiFi connections, but Ethernet and VPN connections are also on the table.

**Describe alternatives you've considered**
The custom module can currently be used to achieve basic functionality but this is sub-par. 

**Additional context**
Crate that looks hopeful for this: https://crates.io/crates/network-manager

yavko:
Problem is all the network manager bindings I could find use the `dbus` crate, and not `zbus` :/

JakeStanger:
How much of a problem is that? Obviously pure rust is better and I'd rather avoid bringing *even more* deps into the mix, but if it works it works right?

yavko:
> How much of a problem is that? Obviously pure rust is better and I'd rather avoid bringing _even more_ deps into the mix, but if it works it works right?

I'm pretty sure the `dbus` crate doesn't support async

JakeStanger:
Not ideal, but that's okay - you can use tokio's `spawn_blocking` to delegate it to its own task and avoid blocking the main thread

yavko:
Thats true, also seems https://crates.io/crates/network-manager is only updated on the github, and not on crates.io

JakeStanger:
Mm yeah didn't spot that. Looks like they've been working towards [libnm-rs](https://github.com/balena-io-modules/libnm-rs) which suffers from the same thing, and has a lot of git dependencies. 

I guess that isn't a complete blocker, but I'd like to try and resolve that before this can be called complete.

yavko:
Is using zbus on its own hard?

yavko:
Forking libnm is also a possibility, seems to only use git deps because of glib, maybe a libnm-ironbar package could be published

JakeStanger:
> Is using zbus on its own hard?

From my (little) experience, it's not *easy* but it's not too bad. It depends a bit on the interface, but is certainly nicer than working with Wayland protocols. 

> Forking libnm is also a possibility

Not against it. I'd rather not have to maintain another library myself if I don't have to but as a last resort, sure.

madushan1000:
I noticed that you're already using Gio, Gio also has decent dbus support with async if you want to write something from scratch.

JakeStanger:
That's good to know, cheers :)

[...]

PR #518, opened 2024-03-30
Add NetworkManager module

Zedfrigg:
Implements a module displaying (very) basic network connectivity from NetworkManager.

First step towards resolving #148.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ default = [
     "ipc",
     "launcher",
     "music+all",
+    "network_manager",
     "notifications",
     "sys_info",
     "tray",
@@ -60,6 +61,8 @@ music = ["regex"]
 "music+all" = ["music", "music+mpris", "music+mpd"]
 "music+mpris" = ["music", "mpris"]
 "music+mpd" = ["music", "mpd-utils"]
+
+network_manager = ["futures-lite", "futures-signals", "zbus"]
 
 notifications = ["zbus"]
 
@@ -136,6 +139,9 @@ chrono = { version = "0.4.38", optional = true, default-features = false, featur
 mpd-utils = { version = "0.2.1", optional = true }
 mpris = { version = "2.0.1", optional = true }
 
+# network_manager
+futures-signals = { version = "0.3.33", optional = true }
+
 # sys_info
 sysinfo = { version = "0.29.11", optional = true }
 
@@ -154,11 +160,11 @@ hyprland = { version = "0.4.0-alpha.2", features = ["silent"], optional = true }
 futures-util = { version = "0.3.30", optional = true }
 
 # shared
-futures-lite = { version = "2.3.0", optional = true } # workspaces, upower
+futures-lite = { version = "2.3.0", optional = true } # network_manager, upower, workspaces
 regex = { version = "1.10.6", default-features = false, features = [
   "std",
 ], optional = true } # music, sys_info
-zbus = { version = "3.15.2", default-features = false, features = ["tokio"], optional = true } # notifications, upower
+zbus = { version = "3.15.2", default-features = false, features = ["tokio"], optional = true } # network_manager, notifications, upower
 
 # schema
 schemars = { version = "0.8.21", optional = true }
--- a/src/clients/networkmanager.rs
+++ b/src/clients/networkmanager.rs
@@ -1,0 +1,169 @@
+use std::sync::Arc;
+
+use color_eyre::Result;
+use futures_signals::signal::{Mutable, MutableSignalCloned};
+use tracing::error;
+use zbus::blocking::fdo::PropertiesProxy;
+use zbus::blocking::Connection;
+use zbus::{
+    dbus_proxy,
+    names::InterfaceName,
+    zvariant::{ObjectPath, Str},
+};
+
+use crate::{register_fallible_client, spawn_blocking};
+
+const DBUS_BUS: &str = "org.freedesktop.NetworkManager";
+const DBUS_PATH: &str = "/org/freedesktop/NetworkManager";
+const DBUS_INTERFACE: &str = "org.freedesktop.NetworkManager";
+
+#[derive(Debug)]
+pub struct Client {
+    client_state: Mutable<ClientState>,
+    interface_name: InterfaceName<'static>,
+    dbus_connection: Connection,
+    props_proxy: PropertiesProxy<'static>,
+}
+
+#[derive(Clone, Debug)]
+pub enum ClientState {
+    WiredConnected,
+    WifiConnected,
+    CellularConnected,
+    VpnConnected,
+    WifiDisconnected,
+    Offline,
+    Unknown,
+}
+
+#[dbus_proxy(
+    default_service = "org.freedesktop.NetworkManager",
+    interface = "org.freedesktop.NetworkManager",
+    default_path = "/org/freedesktop/NetworkManager"
+)]
+trait NetworkManagerDbus {
+    #[dbus_proxy(property)]
+    fn active_connections(&self) -> Result<Vec<ObjectPath>>;
+
+    #[dbus_proxy(property)]
+    fn devices(&self) -> Result<Vec<ObjectPath>>;
+
+    #[dbus_proxy(property)]
+    fn networking_enabled(&self) -> Result<bool>;
+
+    #[dbus_proxy(property)]
+    fn primary_connection(&self) -> Result<ObjectPath>;
+
+    #[dbus_proxy(property)]
+    fn primary_connection_type(&self) -> Result<Str>;
+
+    #[dbus_proxy(property)]
+    fn wireless_enabled(&self) -> Result<bool>;
+}
+
+impl Client {
+    fn new() -> Result<Self> {
+        let client_state = Mutable::new(ClientState::Unknown);
+        let dbus_connection = Connection::system()?;
+        let interface_name = InterfaceName::from_static_str(DBUS_INTERFACE)?;
+        let props_proxy = PropertiesProxy::builder(&dbus_connection)
+            .destination(DBUS_BUS)?
+            .path(DBUS_PATH)?
+            .build()?;
+
+        Ok(Self {
+            client_state,
+            interface_name,
+            dbus_connection,
+            props_proxy,
+        })
+    }
+
+    fn run(&self) -> Result<()> {
+        let proxy = NetworkManagerDbusProxyBlocking::new(&self.dbus_connection)?;
+
+        let mut primary_connection = proxy.primary_connection()?;
+        let mut primary_connection_type = proxy.primary_connection_type()?;
+        let mut wireless_enabled = proxy.wireless_enabled()?;
+
+        self.client_state.set(determine_state(
+            &primary_connection,
+            &primary_connection_type,
+            wireless_enabled,
+        ));
+
+        for change in self.props_proxy.receive_properties_changed()? {
+            let args = change.args()?;
+            if args.interface_name != self.interface_name {
+                continue;
+            }
+
+            let changed_props = args.changed_properties;
+            let mut relevant_prop_changed = false;
+
+            if changed_props.contains_key("PrimaryConnection") {
+                primary_connection = proxy.primary_connection()?;
+                relevant_prop_changed = true;
+            }
+            if changed_props.contains_key("PrimaryConnectionType") {
+                primary_connection_type = proxy.primary_connection_type()?;
+                relevant_prop_changed = true;
+            }
+            if changed_props.contains_key("WirelessEnabled") {
+                wireless_enabled = proxy.wireless_enabled()?;
+                relevant_prop_changed = true;
+            }
+
+            if relevant_prop_changed {
+                self.client_state.set(determine_state(
+                    &primary_connection,
+                    &primary_connection_type,
+                    wireless_enabled,
+                ));
+            }
+        }
+
+        Ok(())
+    }
+
+    pub fn subscribe(&self) -> MutableSignalCloned<ClientState> {
+        self.client_state.signal_cloned()
+    }
+}
+
+pub fn create_client() -> Result<Arc<Client>> {
+    let client = Arc::new(Client::new()?);
+    {
+        let client = client.clone();
+        spawn_blocking(move || {
+            if let Err(error) = client.run() {
+                error!("{}", error);
+            };
+        });
+    }
+    Ok(client)
+}
+
+fn determine_state(
+    primary_connection: &str,
+    primary_connection_type: &str,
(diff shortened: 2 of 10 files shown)


docs/modules/Networkmanager.md
```diff
@@ -0,0 +1,74 @@
+Displays the current network connection state of NetworkManager. Supports wired ethernet, wifi,
+cellular data and VPN connections among others.
+
+This module uses NetworkManager's so-called primary connection, and therefore inherits its
+limitation of only being able to display the "top-level" connection. For example, if we have a VPN
+connection over a wifi connection it will only display the former, until it is disconnected, at
+which point it will display the latter. A solution to this is currently in the works.
```
JakeStanger, line 7: It might be better to use a note block here, just so it's immediately clear to anybody reading the page.

```suggestion
> [!NOTE]
> This module uses NetworkManager's so-called primary connection, 
> and therefore inherits its limitation of only being able to display the "top-level" connection. 
> For example, if we have a VPN connection over a wifi connection it will only display the former, until it is disconnected, at which point it will display the latter. 
> 
> A solution to this is currently in the works.
```

src/config/mod.rs
```diff
@@ -14,6 +14,8 @@ use crate::modules::label::LabelModule;
 use crate::modules::launcher::LauncherModule;
 #[cfg(feature = "music")]
 use crate::modules::music::MusicModule;
+#[cfg(feature = "networkmanager")]
+use crate::modules::networkmanager::NetworkmanagerModule;
```
JakeStanger, line 18: Can struct/enum casing be changed throughout please to `NetworkManager` to match the program name. The feature flag and module type can stay as-is though I think.

src/modules/networkmanager.rs
```diff
@@ -0,0 +1,169 @@
...
+        context: &WidgetContext<NetworkmanagerState, ()>,
+        _: Receiver<()>,
+    ) -> Result<()> {
+        let tx = context.tx.clone();
+
+        spawn(async move {
+            /* TODO: This should be moved into a client à la the upower module, however that
+               requires additional refactoring as both would request a PropertyProxy but on
+               different buses. The proper solution will be to rewrite both to use trait-derived
+               proxies. */
```
JakeStanger, line 61: I think this TODO needs resolving before anything is merged. A controller is spawned for every module instance, which means every instance (across every bar) creates another connection to DBus.

Agreed long-term the right solution is to use the trait-based approach. I'm not sure I understand your reasoning as of current though. 

It should be sufficient to wrap the client code as-is in its own struct, and just provide a `subscribe` method which returns a `broadcast::Receiver`. That's the pattern I've used elsewhere so there should be a good few examples if you need a pointer.

src/modules/networkmanager.rs
```diff
@@ -0,0 +1,169 @@
...
+            let nm_proxy = {
+                let dbus = zbus::Connection::system().await?;
+                PropertiesProxy::builder(&dbus)
+                    .destination("org.freedesktop.NetworkManager")?
+                    .path("/org/freedesktop/NetworkManager")?
```
JakeStanger, line 66: These should be `const` values at the top of the func/module please

src/modules/networkmanager.rs
```diff
@@ -0,0 +1,169 @@
...
+                NetworkmanagerState::Cellular => "network-cellular-symbolic",
+                NetworkmanagerState::Offline => "network-wireless-disabled-symbolic",
+                NetworkmanagerState::Unknown => "dialog-question-symbolic",
+                NetworkmanagerState::Vpn => "network-vpn-symbolic",
+                NetworkmanagerState::Wired => "network-wired-symbolic",
+                NetworkmanagerState::Wireless => "network-wireless-symbolic",
+                NetworkmanagerState::WirelessDisconnected => "network-wireless-acquiring-symbolic",
+            };
+            ImageProvider::parse(icon_name, &icon_theme, false, self.icon_size)
+                .map(|provider| provider.load_into_image(icon.clone()));
```
JakeStanger, line 121: Heads-up I do want to refactor how this works to resolve #511. For now though, this is fine.

src/modules/networkmanager.rs
```diff
@@ -0,0 +1,169 @@
...
+
+async fn get_network_state(
+    nm_proxy: &PropertiesProxy<'_>,
+    device_interface_name: &InterfaceName<'_>,
+) -> Result<NetworkmanagerState> {
+    let properties = nm_proxy.get_all(device_interface_name.clone()).await?;
+
+    let primary_connection_path = properties["PrimaryConnection"]
+        .downcast_ref::<ObjectPath>()
+        .unwrap();
```
JakeStanger, line 136: Unwrap should be avoided in production code. Generally you should either handle the error, or use `expect` if you know it absolutely should not fail to state *why* it should not fail.

In this case, since you're returning `color_eyre`'s magic report, you can just use the `?` operator to propagate any errors up. Either that or fall back to a safe default (ie assume disconnected or unknown)

JakeStanger (changes requested):
Thanks for your hard work on this so far. There's a few changes needed here, but generally the code looks solid & this should be a good foundation for the module.

If you can also resolve the formatting issue, and squash commits down into one please that'd be great. 

Any questions let me know.

src/modules/networkmanager.rs
Zedfrigg: Alright, I've now rewritten the module to use a client. It overall matches the pattern you mentioned, with the exception that I've used a signal from `futures_signals` instead of a Tokio's `broadcast` to provide reactivity from the client to the module controller. Because a single `ClientState` instance contains everything the UI needs to render, there is no need to process previous values, just the latest.

Let me know what you think of this!

src/config/mod.rs
Zedfrigg: Implemented :+1:

docs/modules/Networkmanager.md
Zedfrigg: Implemented :+1:

src/modules/networkmanager.rs
Zedfrigg: This code has now been moved to `clients/networkmanager.rs`, where it now uses the `error!` plus `return` pattern used elsewhere.

src/modules/networkmanager.rs
Zedfrigg: Implemented :+1:

src/clients/networkmanager.rs
```diff
@@ -0,0 +1,187 @@
...
+            return;
+        };
+        let Ok(changed_props_stream) = props_proxy.receive_properties_changed() else {
+            error!("Failed to create NetworkManager D-Bus changed properties stream");
+            return;
+        };
+        let Ok(props) = props_proxy.get_all(interface_name.clone()) else {
+            error!("Failed to get NetworkManager D-Bus properties");
+            return;
+        };
```
Zedfrigg, line 68: The error handling here is uhhh a bit verbose, but I couldn't think of a good way to replace it. If you have any suggestions for this I'd love to hear them.

JakeStanger: If you make `run` return a `Result<()>` (you can either use the ZBus Result type or `color_eyre::Result` if that proves a pain), you should be able to just use the question mark operator. 

```rs
let dbus_connection = Connection::system()?
let builder = PropertiesProxy::builder(&dbus_connection);
let builder = builder.destination(DBUS_BUS)?;

// etc...
```


I recently refactored some of the client code so that they can fail on startup - you should be able to see examples in the other dbus-based clients.

src/modules/networkmanager.rs
JakeStanger: I've not seen `futures_signals` before. It looks powerful but I'm interested as to what it brings in that a broadcast channel doesn't. Does `futures_signals` always return the current state on the first call to `next()`, in order to initialise?

Zedfrigg: Yes, basically if the `Mutable` in the client is updated more than once, followed by a read of its signal through `next()`, that read will only see the most recent value. This is of course fairly useless for communicating events (essentially changes to state), because we need to process all of them in order to arrive at the new state. But because it's much simpler to just communicate the entire new state (`ClientState`) to the module instances, we don't need a queue, just the latest value. Also see [the docs](https://docs.rs/futures-signals/latest/futures_signals/tutorial/index.html).

Of course, if you don't want to pull in the extra dependency I can rewrite it in terms of a `broadcast` queue, at the cost of some performance.

As for your second question, yes, all the module instances receive the update - as long as they call `subscribe` of course.

src/clients/networkmanager.rs
Zedfrigg: Excellent timing with that refactor :ok_hand: 

I've rewritten things accordingly, and it's looking a lot nicer now, thanks.

Zedfrigg:
@JakeStanger let me know if you think the code is ready to merge in its current state :)

src/modules/networkmanager.rs
```diff
@@ -2,7 +2,7 @@ use color_eyre::Result;
 use futures_lite::StreamExt;
 use futures_signals::signal::SignalExt;
 use gtk::prelude::ContainerExt;
-use gtk::{Image, Orientation};
+use gtk::{Box, Image, Orientation};
```
JakeStanger, line 5: I'm not a fan of importing `gtk::Box` because it makes it harder to use the stdlib Box type.

JakeStanger (changes requested):
It's getting there generally, but there's still a few things pending  & I'm still not 100% sure about the signals crate - if you can get back to my Qs pls just to clear up some potential concerns.

I need to give this a proper test too, so I'll try and do that as soon as I can.

src/modules/networkmanager.rs
Zedfrigg: Fair enough, I aliased them as `GtkBox` now

src/clients/networkmanager.rs
```diff
@@ -0,0 +1,187 @@
...
+
+        let mut primary_connection = {
+            let Some(primary_connection) = props["PrimaryConnection"].downcast_ref::<ObjectPath>()
+            else {
+                error!("PrimaryConnection D-Bus property is not a path");
+                return;
+            };
+            primary_connection.to_string()
+        };
```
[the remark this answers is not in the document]
Zedfrigg: Yes, see [here](https://people.freedesktop.org/~lkundrak/nm-docs/gdbus-org.freedesktop.NetworkManager.html)

src/modules/networkmanager.rs
Zedfrigg: Yeah, we could either choose to put them in various structs that then get nested in each other, or have multiple `Mutable`s, each for a distinct part of the state. Either approach will work. For communication back from the widget to the client, we'll stick with the events on a `tokio::sync::broadcast` model.

src/clients/networkmanager.rs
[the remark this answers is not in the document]
JakeStanger: I hate DBus :) Got to be the worst syntax for showing type info ever conceived. 

I was initially wondering if a macro would help reduce the boilerplate here, but not sure if it will actually?

src/modules/networkmanager.rs
JakeStanger: Okay, that sounds good to me then, happy to stick with that.

src/clients/networkmanager.rs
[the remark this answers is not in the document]
Zedfrigg: What will definitely help with that is switching to using proxy traits. I should get around to doing that refactor...

As a tangent, I also like how the zbus crate does it. They take a different approach and just infer everything from the Rust type of the variable you assign it to.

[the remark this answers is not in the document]
Zedfrigg: Update: I've now done that refactor and the code is indeed now looking a lot nicer.

Cargo.toml
```diff
@@ -136,6 +139,12 @@ chrono = { version = "0.4.38", optional = true, default_features = false, featur
 mpd-utils = { version = "0.2.1", optional = true }
 mpris = { version = "2.0.1", optional = true }
 
+# networkmanager
+futures-signals = { version = "0.3.33", optional = true }
+
+# schema
+schemars = { version = "0.8.21", optional = true }
```
JakeStanger, line 146: I intentionally added this one to the bottom as it's not a 'standard' module feature

src/modules/networkmanager.rs
```diff
@@ -0,0 +1,87 @@
...
+
+use crate::clients::networkmanager::{Client, ClientState};
+use crate::config::CommonConfig;
+use crate::gtk_helpers::IronbarGtkExt;
+use crate::image::ImageProvider;
+use crate::modules::{Module, ModuleInfo, ModuleParts, ModuleUpdateEvent, WidgetContext};
+use crate::{glib_recv, module_impl, send_async, spawn};
+
+#[derive(Debug, Deserialize, Clone)]
+pub struct NetworkManagerModule {
```
JakeStanger, line 17: The introduction of JSON schema support annoyingly means every type used in the config needs to derive the `JsonSchema` type when the feature is enabled:

```suggestion
#[derive(Debug, Deserialize, Clone)]
#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
pub struct NetworkManagerModule {
```

JakeStanger (changes requested):
Thanks for the work on the refactor, and apologies I haven't inputted a lot on this one for a while.

A couple of very small changes now, but outside of those the code lgtm now. I'll give this some tests and report back.

Is there anything else you want to add at the PR at this stage, or do you reckon we're go for the initial version?

JakeStanger:
> I'll give this some tests and report back.

Looks like it's all working as intended. I've tested Ethernet and Wifi and toggling connections in both cases causes instant updates.

Cargo.toml
Zedfrigg: Oh I see! Moved it back now :+1:

src/modules/networkmanager.rs
Zedfrigg: Added it :+1:

Zedfrigg:
> Is there anything else you want to add at the PR at this stage, or do you reckon we're go for the initial version?

No as far as I'm concerned we're good to go!

JakeStanger:
Sorry for delay on this (again). There's a merge conflict outstanding - if you can resolve that I'll add the docs and merge :)

Zedfrigg:
Merged and re-tested :+1:

The `devShell` CI workflow is failing, but I don't think it's related to this change?

JakeStanger (approved):
Apologies for the huge delay on this, and thanks again for all your work. I've finally got round to putting the finishing touches on it.

FYI, I've made some basic refactors and renamed the module/feature flag to `network_manager`.

I'll merge this in a few mins and it'll be available in the git build.

PR #530, opened 2024-04-04
feat: Add orientation support for clock, button, label, and sys_info

ClaireNeveu:
After making the switch to hyprland ironbar is by far the best dock solution I've found. Most wayland dockbars that aren't built into a big DE have very poor support for vertical docks _or_ lack a proper program list/launcher. The one big issue I've run into is that the clock is incorrectly rotated when using a vertical bar. In [this discussion](https://github.com/JakeStanger/ironbar/discussions/294) you mentioned you wanted to keep the current behavior as an option.

I went ahead and added orientation support for clock as well as label and button which some people were asking for. I also created an enum for orientation and moved deserialization into serde and out of an ad-hoc function.

I also added a test config where you can see that all variations (default, horizontal, and vertical along with shorthand) are functioning properly.

Please let me know if there's anything you need on my end to get this merged, I didn't see any contribution guidelines or standards in the repo.


ClaireNeveu:
Just noticed there are a few module's I'm not using that currently still rotate based on bar position (upower, music, sysinfo) so I'd be happy to add orientation support for those as well if you like.

src/modules/custom/label.rs
```diff
@@ -12,6 +13,7 @@ pub struct LabelWidget {
     name: Option<String>,
     class: Option<String>,
     label: String,
+    orientation: Option<ModuleOrientation>,
```
JakeStanger, line 16: Perhaps it'd better to provide a `#serde[default]` attr here, so we can remove the `Option` and additional logic that comes with that

test-configs/orientation.yaml
JakeStanger, line 1: Certainly not against including this, as some test configs would be good, but just wondering what your thinking is here?

JakeStanger (changes requested):
Hey, thanks for the PR. I really appreciate the work you've put in here. Vertical bars have needed some love for a very long time so I'm very glad to finally see them get some.

The code generally lgtm, there's a few picky points I've raised but nothing of actual concern.

src/modules/clock.rs
```diff
@@ -97,10 +101,15 @@ impl Module<Button> for ClockModule {
...
-            .build();
+        let label = if matches!(self.orientation, ModuleOrientation::Vertical) {
+            Label::builder()
+                .angle(90.0)
+                .use_markup(true)
+        } else {
+            Label::builder()
+                .use_markup(true)
+            
+        }.build();
```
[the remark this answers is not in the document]
ClaireNeveu: Yep, didn't go back and change this after I added the function.

src/config/common.rs
```diff
@@ -40,6 +40,36 @@ pub enum TransitionType {
...
+    Vertical,
+}
+
+impl ModuleOrientation {
+    pub const fn to_gtk(&self) -> Orientation {
+        match self {
+            Self::Horizontal => Orientation::Horizontal,
+            Self::Vertical => Orientation::Vertical
+        }
+    }
```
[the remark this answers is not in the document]
ClaireNeveu: Thanks, I don't write rust so I'm unfamiliar with the idioms.

src/modules/custom/label.rs
ClaireNeveu: Good call.

test-configs/orientation.yaml
ClaireNeveu: My first thought was that I wanted some unit tests to make sure the various config options I was adding parsed correctly. Secondly I wanted to make sure they actually _displayed_ correctly as well. Ideally there would be some integration tests but I don't know nearly enough about rust or wayland to make that happen so I just added a config I could check manually. 

I found this very helpful when developing and as a reviewer it provides you an easy way to confirm my code works. but if you'd rather I exclude it I can remove it easily.

ClaireNeveu:
I addressed the PR feedback and I added orientation support for sysinfo. I took a stab at upower and music but I got hung up on rotating the icons (if you know how to do this I'd be happy to do the rest). Notably right now upower at least doesn't really have proper behavior on vertical bars since only the label changes when the whole button needs to change.

IMO it'd be better to just remove line 184 there, you'd get a more functional widget in vertical mode.

src/modules/sysinfo.rs
```diff
@@ -21,6 +21,11 @@ pub struct SysInfoModule {
     #[serde(default = "Interval::default")]
     interval: Interval,
 
+    #[serde(default)]
+    orientation: ModuleOrientation,
```
ClaireNeveu, line 25: I actually added two options here. Although I don't use sysinfo personally my expectation here for a vertical bar would be horizontal text in a vertical layout which can now be done via `orientation: horizontal, layout: vertical`.

src/config/common.rs
```diff
@@ -40,6 +40,34 @@ pub enum TransitionType {
...
+            Self::Vertical => 90.0,
+        }
+    }
+}
+
+impl From<ModuleOrientation> for Orientation {
+    fn from(o: ModuleOrientation) -> Self {
+        match o {
+            ModuleOrientation::Horizontal => Orientation::Horizontal,
+            ModuleOrientation::Vertical => Orientation::Vertical,
```
JakeStanger, line 66: You can use the `Self` alias here:

```suggestion
            ModuleOrientation::Horizontal => Self::Horizontal,
            ModuleOrientation::Vertical => Self::Vertical,
```

src/modules/custom/box.rs
```diff
@@ -20,9 +20,7 @@ impl CustomWidget for BoxWidget {
         let container = build!(self, Self::Widget);
 
         if let Some(orientation) = self.orientation {
-            container.set_orientation(
-                try_get_orientation(&orientation).unwrap_or(Orientation::Horizontal),
-            );
+            container.set_orientation(Orientation::from(orientation));
```
JakeStanger, line 23: Implementing `From` gives you a free `Into` impl, and Rust is clever enough to infer the types. Sometimes using From is better because it's more explicit, but I think here it's obvious enough the concise syntax is preferable: 

```suggestion
            container.set_orientation(orientation.into());
```

(Same story in `slider` and `sysinfo`)

src/modules/sysinfo.rs
JakeStanger: I think the two options is a great idea, but I'm not sure on the naming yet. My concern is that we're using `orientation` throughout, the `tray` module already has a `direction` option, and this'd add a third `layout` option which doesn't feel tremendously consistent.

It might be that the tray option should be renamed (unsure). Alternatively I'm wondering if we should stick to `orientation` here for both, as in:

- `orientation` - controls the box (aka `layout` currently)
- `label_orientation` controls the labels

Very open to ideas on this one so let me know what you think

test-configs/orientation.yaml
JakeStanger: > Ideally there would be some integration tests but I don't know nearly enough about rust or wayland to make that happen so I just added a config I could check manually.

Agreed we could really do with some more automated tests. The difficult part is ideally the tests *shouldn't* test against the actual GTK bar, as that'd be testing GTK and the compositor rather than our logic. Somehow the tests need to run against our own logic, which is the tricky bit and why I've not got round to it yet :p  

I suspect some refactoring will be required there, but that's my problem.

I do also have an ever-mutating test config that I've never found a home for, which I think this finally gives it. Hopefully this will give me reason to standardise it and put more in there so there's at least good manual candidates.

My only issue is YAML. I know I'm biased but I'd rather the test configs were in Corn, as that way they can easily be converted to JSON, YAML, TOML as desired/required. Also I hate YAML lol

JakeStanger (changes requested):
A few more small changes please, and then if you can squash your commits down to remove the PR-related noise I think we're getting there.

> I took a stab at upower and music but I got hung up on rotating the icons (if you know how to do this I'd be happy to do the rest).

Sure, looking at re-doing the icon code is actually next on my agenda as part of https://github.com/JakeStanger/ironbar/issues/511 anyway so I can pick that up. 

Feel free to remove that line from UPower in the meantime.

test-configs/orientation.yaml
ClaireNeveu: I'd be happy to rewrite it to corn, it's just YAML because that's my preference.

src/modules/custom/box.rs
ClaireNeveu: Thanks.

ClaireNeveu:
@JakeStanger Resolved all your feedback. Is there a particular reason you want me to squash my commits manually instead of using Github's squash merge?

src/modules/sysinfo.rs
ClaireNeveu: I think `direction` is correct.

JakeStanger:
> Is there a particular reason you want me to squash my commits manually instead of using Github's squash merge feature?

GH's squash merge will reduce everything into a single commit, I just want to filter out the PR-related commits :) There's still several changes this PR makes so I'd like to keep them separated so it's clear in the changelog:

- Introduction of orientation option
- Change of UPower
- Introduction of test file
- Anything else I've missed

ClaireNeveu:
Gotcha, I'll try to get it down to relevant commits. It does mean that the commits themselves won't be strictly what is mentioned in their commit messages since interactive rebase can only squash into adjacent commits.

JakeStanger:
> interactive rebase can only squash into adjacent commits.

You can actually squash into any commit! You can re-order the lines in the editor to change what they're adjacent to, and it should work it out. If you want a hand with that let me know.

JakeStanger (commented):
Code all LGTM - if you can just fix formatting too pls, and then we're there. Cheers for your hard work on this

ClaireNeveu:
I don't have a separate commit for actually introducing the test config but I'll preserve the commit where I converted it to corn.

ClaireNeveu:
Ok, I reworked my commits. I've gotten used to squash merges but in the future I'll try to maintain better commit hygiene on PRs for ironbar.

JakeStanger (approved):

ClaireNeveu:
@JakeStanger Thanks for all the help and the quick merge!

PR #547, opened 2024-04-21
Fix #546: Convert from ARGB32 to RGBA32 formats for Pixmaps

### LINKED ISSUE #546: Tray icon color channels are being incorrectly rendered
**Describe the bug**
The tray module is incorrectly rendering the color channels of the pixmaps provided by applications with tray icons.

**To Reproduce**

1. Update the config to include the tray module (very large icons make it easier to see)
   ```yaml
   type: tray
   icon_size: 128
   prefer_theme_icons: false
   ```
2. Observe the tray icon in detail, and notice that it does not appear the same as the original icon viewed in an image viewer.

**Expected behavior**
The tray icon should be rendered the same way an image viewer would.

**System information:**
- Distro: Arch Linux
- Compositor: Hyprland
 - Ironbar version: f4e1e81

**Configuration**
```yaml
start: []
center: []
end:
  - type: tray
    icon_size: 128
    prefer_theme_icons: true
```

**Additional context**
The GTK documentation [here](https://docs.gtk.org/gdk-pixbuf/class.Pixbuf.html#image-data) shows that the internal representation of their pixel data is RGBA, instead of the ARGB data that's being handled by `modules/tray/icon.rs::get_image_from_pixmap`

**Screenshots**
Rendered Icon:
2024-04-21T19:02:39,174565917-04:00

Original Icon (Before it was scaled to 22x22 and back):
logo

rdnelson:
Simple change to swap the channel order of the incoming pixmap data from ARGB to RGBA before creating the pixbuf.

Using the same icon as the one I added in the issue, it now renders like this, which is much more accurate to the original.

2024-04-21T19:12:29,362148653-04:00

Also I'm still very new to rust, so if there's a cleaner way to do this please let me know.

Fixes #546

### DIFF
--- a/src/modules/tray/icon.rs
+++ b/src/modules/tray/icon.rs
@@ -79,6 +79,8 @@ fn get_image_from_icon_name(item: &TrayMenu, icon_theme: &IconTheme, size: u32) 
 ///
 /// The pixmap is supplied in ARGB32 format,
 /// which has 8 bits per sample and a bit stride of `4*width`.
+/// The Pixbuf expects RGBA32 format, so some channel shuffling
+/// is required.
 fn get_image_from_pixmap(item: &TrayMenu, size: u32) -> Result<Image> {
     const BITS_PER_SAMPLE: i32 = 8;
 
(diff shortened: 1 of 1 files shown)


JakeStanger:
Thanks for the fix. This is a very efficient solution as it mutates the vec in-place, not sure you could have done it better.

PR #555, opened 2024-04-24
Update workspaces module wiki

nyadiia:
The toml example currently doesn't work. This changes it so it does


JakeStanger:
Good spot, thanks. 

This does raise a need for configs to be tested, so I'll open an issue for that. In theory, it shouldn't be too difficult to pull from `test-configs` and auto-gen the tests.

PR #564, opened 2024-04-29
ci(binary): use release tag instead of github.ref_name

slowsage:
Sorry, one oversight: The zip file uses github.ref_name which resolves to master. This uses the release tag instead. Verified [here](https://github.com/slowsage/ironbar/releases/tag/v0.15.4).
You can run  again using `workflow_dispatch` to fix the name for `v0.15.0`.


slowsage:
Yes. This should generate new files. The old ones should be removed manually.

PR #671, opened 2024-07-20
Feat: sway-mode module

Rodrigodd:
Add a new module, similar to the waybar "sway/module" module. It shows the current mode under sway. For that, it needs to communicate with the Sway IPC. In this PR, the module is called "sway-mode" because "sway/mode" didn't work in my YAML config file (it would probably work if I surrounded it with quotes, but I didn't test).

This PR has two commits: the first one implements the module using a new sway connection. The second one refactors the sway `Client`, under `clients::compositor::sway`, to allow dynamically registering events, instead of only handling Workspace update ones. This second commit avoid creating a new socket for each module that needs to communicate with the Sway IPC, not sure if that is a problem to begin with.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -76,8 +76,10 @@ volume = ["libpulse-binding"]
 
 workspaces = ["futures-lite"]
 "workspaces+all" = ["workspaces", "workspaces+sway", "workspaces+hyprland"]
-"workspaces+sway" = ["workspaces", "swayipc-async"]
+"workspaces+sway" = ["workspaces", "sway"]
 "workspaces+hyprland" = ["workspaces", "hyprland"]
+
+sway = ["swayipc-async"]
 
 schema = ["dep:schemars"]
 
--- a/docs/modules/Sway-mode.md
+++ b/docs/modules/Sway-mode.md
@@ -1,0 +1,74 @@
+Displays the current sway mode.
+
+## Configuration
+
+> Type: `sway-mode`
+
+| Name                  | Type                                        | Default | Description                                                                                                                                           |
+| --------------------- | ------------------------------------------- | ------- | ----------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `truncate`            | `'start'` or `'middle'` or `'end'` or `Map` | `null`  | The location of the ellipses and where to truncate text from. Leave null to avoid truncating. Use the long-hand `Map` version if specifying a length. |
+| `truncate.mode`       | `'start'` or `'middle'` or `'end'`          | `null`  | The location of the ellipses and where to truncate text from. Leave null to avoid truncating.                                                         |
+| `truncate.length`     | `integer`                                   | `null`  | The fixed width (in chars) of the widget. Leave blank to let GTK automatically handle.                                                                |
+| `truncate.max_length` | `integer`                                   | `null`  | The maximum number of characters before truncating. Leave blank to let GTK automatically handle.                                                      |
+
+<details>
+<summary>JSON</summary>
+
+```json
+{
+  "end": [
+    {
+      "type": "sway-mode",
+      "truncate": "start"
+    }
+  ]
+}
+```
+
+</details>
+
+<details>
+<summary>TOML</summary>
+
+```toml
+[[end]]
+type = "sway-mode"
+truncate = "start"
+```
+
+</details>
+
+<details>
+<summary>YAML</summary>
+
+```yaml
+end:
+  - type: "sway-mode"
+    truncate: "start"
+```
+
+</details>
+
+<details>
+<summary>Corn</summary>
+
+```corn
+{
+  end = [
+    {
+      type = "sway-mode"
+      truncate = "start"
+    }
+  ]
+}
+```
+
+</details>
+
+## Styling
+
+| Selector | Description  |
+| -------- | ------------ |
+| `.label` | Label widget |
+
+For more information on styling, please see the [styling guide](styling-guide).
--- a/docs/modules/Sway-mode.md
+++ b/docs/modules/Sway-mode.md
@@ -1,4 +1,8 @@
-Displays the current sway mode.
+Displays the current sway mode in a label. If the current sway mode is
+"default", nothing is displayed.
+
+> [!NOTE]
+> This module only works under the [Sway](https://swaywm.org/) compositor.
 
 ## Configuration
 
@@ -67,8 +71,8 @@ end:
 
 ## Styling
 
-| Selector | Description  |
-| -------- | ------------ |
-| `.label` | Label widget |
+| Selector     | Description            |
+| ------------ | ---------------------- |
+| `.sway_mode` | Sway mode label widget |
 
 For more information on styling, please see the [styling guide](styling-guide).
--- a/src/clients/compositor/sway.rs
+++ b/src/clients/compositor/sway.rs
@@ -59,8 +59,9 @@ impl Client {
         }
 
         // Only the task and self have a reference to listeners, and we just abort the task. This
-        // is the only reference to listeners, so we can safely unwrap.
-        let listeners_mut = Arc::get_mut(listeners).unwrap();
+        // is the only reference to listeners, so we can safely get a mutable reference.
+        let listeners_mut = Arc::get_mut(listeners)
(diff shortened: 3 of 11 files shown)


src/clients/compositor/sway.rs
```diff
@@ -155,3 +207,18 @@ impl From<WorkspaceEvent> for WorkspaceUpdate {
         }
     }
 }
+
+fn sway_event_to_event_type(event: &Event) -> EventType {
```
Rodrigodd, line 211: This function could eventually be replaced by https://github.com/JayceFayne/swayipc-rs/pull/55

Cargo.toml
```diff
@@ -73,9 +73,11 @@ volume = ["libpulse-binding"]
 
 workspaces = ["futures-lite"]
 "workspaces+all" = ["workspaces", "workspaces+sway", "workspaces+hyprland"]
-"workspaces+sway" = ["workspaces", "swayipc-async"]
+"workspaces+sway" = ["workspaces", "sway"]
```
JakeStanger, line 76: I'd actually rather this *was* duplicated, as it keeps it explicit & if the sway module deps change in the future it avoids bringing in to the workspace.

```suggestion
"workspaces+sway" = ["workspaces", "swayipc-async"]
````

File: docs/modules/Sway-mode.md [hunk 3]
```diff
@@ -0,0 +1,74 @@
+Displays the current sway mode.
+
```
JakeStanger, line 2: Should be obvious, but could you add a note block here to make it explicit that the module is for the Sway compositor only pls

docs/modules/Sway-mode.md
```diff
@@ -0,0 +1,74 @@
...
+}
+```
+
+</details>
+
+## Styling
+
+| Selector | Description  |
+| -------- | ------------ |
+| `.label` | Label widget |
```
JakeStanger, line 72: I think this section needs updating - assuming it was copy/pasted from the label docs? :P

src/clients/compositor/sway.rs
```diff
@@ -4,15 +4,31 @@ use color_eyre::{Report, Result};
...
+}
+
 pub struct Client {
     client: Arc<Mutex<Connection>>,
-    workspace_tx: Sender<WorkspaceUpdate>,
-    _workspace_rx: Receiver<WorkspaceUpdate>,
+    task_state: Mutex<TaskState>,
+}
+
+impl std::fmt::Debug for Client {
```
JakeStanger, line 25: What's the reason for the manual debug impl here?

src/clients/compositor/sway.rs
```diff
@@ -21,36 +37,59 @@ impl Client {
...
+            let _ = handle.await;
+        }
 
-        {
-            // create 2nd client as subscription takes ownership
-            let client = Connection::new().await?;
-            let workspace_tx = workspace_tx.clone();
+        // Only the task and self have a reference to listeners, and we just abort the task. This
+        // is the only reference to listeners, so we can safely unwrap.
+        let listeners_mut = Arc::get_mut(listeners).unwrap();
```
JakeStanger, line 63: `.unwrap` should be [avoided](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap) in prod code. In part, it's because it is unclear that the situation *is* safe, without relying on a comment above. Generally if you're unwrapping because the code would have to get into an "impossible" state you should use `expect` instead.

In this case since you're returning a `Result` anyway, it would be better to return an error either with `?` or explicitly, and handle the situation at the call.

src/clients/compositor/sway.rs
JakeStanger: Cool, happy with that for now

src/modules/sway/mode.rs
```diff
@@ -0,0 +1,94 @@
...
+            let client = context.ironbar.clients.borrow_mut().sway()?;
+            client
+                .add_listener(
+                    EventType::Mode,
+                    Box::new(move |event| {
+                        trace!("event: {:?}", event);
+                        let Event::Mode(mode) = event else {
+                            unreachable!()
+                        };
+                        let _ = await_sync(tx.send(ModuleUpdateEvent::Update(mode.clone())));
```
JakeStanger, line 51: This can be simplified using `crate::try_send`:
```suggestion
                        try_send!(tx, ModuleUpdateEvent::Update(mode.clone()))
```

(I think I counted those brackets right...)

src/modules/sway/mode.rs
```diff
@@ -0,0 +1,94 @@
...
+        await_sync(async move {
+            let client = context.ironbar.clients.borrow_mut().sway()?;
+            client
+                .add_listener(
+                    EventType::Mode,
+                    Box::new(move |event| {
+                        trace!("event: {:?}", event);
+                        let Event::Mode(mode) = event else {
+                            unreachable!()
+                        };
```
JakeStanger, line 50: I'm not a fan of requiring this when the type is (sorta) already specified in the call. I can't see how to avoid it nicely though. Any ideas?

JakeStanger (changes requested):
Hey, thanks for this. It looks very good, just a few nits. 

Are you also able to provide some evidence of this working please? I can't get Sway to run at the moment :(

src/clients/compositor/sway.rs
Rodrigodd: `WorkspaceClient` requires `Debug`, and `TaskState::listeners: Arc<Vec<(EventType, Box<SyncFn>)>>,` does not implement it, so I need to make a manual implementation, at least for `TaskState`. I could make the debug implementation list the events being listened, but I preferred to avoid locking `TaskState` when debug printing and just printing `<...>` instead. Let me know if you prefer otherwise.

Cargo.toml
Rodrigodd: Tried to do this, but `create_workspace_client` uses `sway::Client`, which is also is used by the `sway-mode` module. I can implement this change by carefully spreading `cfg(any(feature = "workspaces+sway", feature = "sway"))` in a couple of places, but I not sure if that is a win.

Rodrigodd: Actually, currently `sway::Client` is under `clients::compositor`, which is gated by `workspaces`. I first need to move the `sway::Client` out of there, it does not need to be gated by `workspaces`.

src/clients/compositor/sway.rs
JakeStanger: Cool nope that makes sense & that's good with me. Figured it'd be something like that, just wnated to check.

src/modules/sway/mode.rs
Rodrigodd: I created a `SwayIpcEvent` trait with a associated const `EVENT_TYPE: EventType` and a `fn from_event(&Event) -> &Self` method, implemented to all event types, and used that to fix this issue.

Rodrigodd:
I believe I have addressed all points. Here is a video of the sway mode working:


https://github.com/user-attachments/assets/20a3d3f2-2844-4956-a352-67d1ca1b6897

JakeStanger (commented):
Code all lgtm, cheers for making those changes & cheers for the vid.

One last thing before I approve - could you update `docs/sidebar.md` to add an entry for the docs page pls

Rodrigodd:
@JakeStanger done.

JakeStanger (approved):

src/clients/sway.rs
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    pub async fn add_listener<T: SwayIpcEvent>(
+        &self,
+        f: impl Fn(&T) + Sync + Send + 'static,
+    ) -> Result<()> {
+        self.add_listener_type(
+            T::EVENT_TYPE,
+            Box::new(move |event| {
+                let event = T::from_event(event).unwrap();
```
JakeStanger, line 58: Oops no sorry, one more actually - can we remove this `unwrap` please - either `expect` if it should never fail, or safely handled otherwise

Rodrigodd: Replace it by a `expect`. The type is enum variant is checked in `add_listener_type`, write after passing it to the closure.

Rodrigodd:
@JakeStanger all done. (just in case my last comment didn't trigger a notification)

JakeStanger:
Cheers!

PR #683, opened 2024-07-30
feat: add shell completions

donovanglover:
Auto-generated with clap.

Note: May be ideal to replace some of the sub-commands with flags to decrease the amount of generated man pages.

Edit: Draft while I think of something better to do here

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,53 @@
+// Importing from Ironbar modules brings in lots of things not used by the build script
+// we can just globally suppress those.
+#![allow(unused, dead_code)]
+
+#[path = "src/cli.rs"]
+mod cli;
+
+#[path = "src/error.rs"]
+mod error;
+
+#[path = "src/ipc"]
+mod ipc {
+    #[path = "commands.rs"]
+    mod commands;
+
+    #[path = "responses.rs"]
+    mod responses;
+
+    pub use commands::Command;
+    pub use responses::Response;
+}
+
+use clap::Command;
+use clap::CommandFactory;
+use clap_complete::generate_to;
+use clap_complete::Shell::{Bash, Fish, Zsh};
+use cli::Args;
+use std::fs;
+use std::path::PathBuf;
+
+const NAME: &str = "ironbar";
+
+fn generate_shell_completions(mut cmd: Command) -> std::io::Result<()> {
+    const MANIFEST_DIR: &str = env!("CARGO_MANIFEST_DIR");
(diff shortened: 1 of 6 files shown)


donovanglover:
Upstream issue: https://github.com/clap-rs/clap/issues/5337

Could add shell completions now and think about man pages later.

Long-term could auto-complete bar names and other information as well.

build.rs
```diff
@@ -0,0 +1,69 @@
...
+use std::fs;
+use std::path::PathBuf;
+
+static NAME: &str = "ironbar";
+
+fn generate_man_pages(cmd: Command) {
+    let man_dir = PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("target/man");
+    let mut buffer = Vec::default();
+
+    Man::new(cmd.clone()).render(&mut buffer).unwrap();
```
JakeStanger, line 19: I appreciate this is just a build script, but I'd still prefer to keep `unwrap`s out of the code pls. I'll be more lenient than usual towards using `expect` everywhere so long as it's clear why it *shouldn't* crash.

src/cli.rs
JakeStanger, line 1: What's the reason for the restructuring here and moving `handle_response`? It looks like the diff has gone funny, so I'm assuming this isn't entirely intentional?

JakeStanger (changes requested):
Thanks for your work here, this will be a great QoL improvement to have. Looking like it's most of the way there already.

I'll be looking to merge this as part of the 0.17.0 release as 0.16.1 will be bugfixes only.

src/cli.rs
donovanglover: I think this was to avoid unused code warnings from `build.rs`

donovanglover:
Unfortunately I doubt I'll have time to work on this anytime soon but you're free to change anything as needed. Should also avoid duplicate code from `src/ipc/commands.rs`.

JakeStanger:
No worries, thanks for the heads up. I'm happy to take it from here, appreciate the work you've done to get it this far.

JakeStanger:
I'm going to drop manpage support for now because I see it having pretty limited use, especially when `ironbar --help` provides a much more ergonomic interface to see the same information. Having every subcommmand as its own page sorta sucks, and I don't like the idea of changing the CLI interface to fit it.

The completion is already useful though, so I'll merge that.

donovanglover:
Sounds good to me. Thanks for working on this!

JakeStanger (approved):

JakeStanger:
Thanks for getting it started :)

Issue #632, opened 2024-06-04
[Tray] Some icons not appearing on startup


ToxicMushroom:
I didn't find much when searching the wiki and repo for info on how to launch ironbar other than https://github.com/JakeStanger/ironbar/issues/11
Currently I am launching it via swaybar_command but I have multiple programs that autostart and rarely get picked up by the tray (seafile-applet, keepassxc). 

If I kill said programs and relaunch them they do consistently appear, indicating some sort of race condition between the bar loading and the programs wanting to appear in the tray (I've no idea on how the appindicator protocol works)

So my question is if I'm not launching it correctly or if this is a bug. And if we solved this we can maybe add documentation to the wiki on starting ironbar.

Fwiw I didn't have these issues with swaybar

**System information:**
 - Distro: Arch Linux
 - Compositor: Sway 1.9
 - Ironbar version: 0.16.0-pre

**Configuration**
<details><summary>Sway config</summary>

```ini
# Default config for sway
#
# Copy this to ~/.config/sway/config and edit it to your liking.
#
# Read `man 5 sway` for a complete reference.

### Variables
#
# Logo key. Use Mod1 for Alt.
set $mod Mod4
# Home row direction keys, like vim
set $left h
set $down j
set $up k
set $right l
# Your preferred terminal emulator
set $term alacritty
# Your preferred application launcher
# Note: pass the final command to swaymsg so that the resulting window can be opened
# on the original workspace that the command was run on.
set $menu fuzzel
set $explorer nemo

exec "systemctl --user import-environment DISPLAY XAUTHORITY WAYLAND_DISPLAY XDG_CURRENT_DESKTOP"
### Output configuration
#
# Default wallpaper (more resolutions are available in /usr/share/backgrounds/sway/)
output * bg /home/merlijn/Seafile/private/images/desktop-wallpapers/hollow-knight-1.jpg fill 
output DP-2 {
    mode 1920x1080@144Hz
    pos 1920 0
}
output HDMI-A-1 {
    mode 1920x1080@60Hz
    pos 0 0
}

# Autostart apps

    # Night mode
    exec gammastep

    # Notification server
    exec swaync

    # Clipboard history manager
    exec env copyq

    exec seafile-applet

    # Password manager
    exec keepassxc

    # Autostart noisetorch
    exec /home/merlijn/.bin/noisetorch.sh 
#
#   output HDMI-A-1 resolution 1920x1080 position 1920,0
#
[...]

PR #685, opened 2024-08-04
fix(tray): crash when provided empty pixmap

### LINKED ISSUE #675: [Tray] Segfaults when starting certain applications
**Describe the bug**
Ironbar segfaults when starting Mullvad VPN with the tray module enabled

**To Reproduce**
1. Start ironbar with the tray module
2. Start Mullvad VPN

**Expected behavior**
No crash

**System information:**
 - Distro: NixOS unstable
- Compositor: Hyprland
 - Ironbar version: https://github.com/JakeStanger/ironbar/commit/58190ab079d00dd53babb72346f1da6e1cc9ac72

**Additional context**
Likely caused by one of the `unsafe` blocks involving cairo since the segfault mentions `libcairo.so.2`

JakeStanger:
Fixes #675

### DIFF
--- a/src/modules/tray/icon.rs
+++ b/src/modules/tray/icon.rs
@@ -90,6 +90,10 @@ fn get_image_from_pixmap(item: &TrayMenu, size: u32) -> Result<Image> {
         .and_then(|pixmap| pixmap.first())
         .ok_or_else(|| Report::msg("Failed to get pixmap from tray icon"))?;
 
+    if pixmap.width == 0 || pixmap.height == 0 {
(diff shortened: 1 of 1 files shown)


donovanglover:
This fixes the crash, however now it shows an empty tray icon until ironbar is restarted. I think the icon is being updated after the application starts.

JakeStanger:
I suspected something like that might happen. There's definitely some odd behaviours with the tray I still need to work out, especially around images.

I'll merge this as it does fix the issue. There's a list of other applications affected by no icon showing here: https://github.com/JakeStanger/ironbar/issues/632

PR #690, opened 2024-08-04
One icon per connection type in NetworkManager module

Zedfrigg:
Draft PR to track the status of the implementation. Will watch all NM connections and figure out if they should be categorised as ethernet, wifi, cellular, VPN or other. If there is at least one connection in a category, an icon will be shown for it indicating its status in more detail.


Zedfrigg:
Commit `48493C6` here serves as a summary of the changes so far, I forgot that Github includes parent commits even if they weren't made on the PR branch.

Before I'll mark this PR as ready for review I will implement support for showing the wifi signal strength, and consider what the configuration should look like.

PR #699, opened 2024-08-10
Update flake.nix to match adwaita-icon-theme being moved to top-level

GameDungeon:
The icon theme was moved in the nixpkgs repo and this fixes a warning message that change causes


JakeStanger:
Cheers

PR #726, opened 2024-09-08
Support For Niri Workspaces

### LINKED ISSUE #650: Support Niri out of the box
> Currently workspaces in Niri WM are not supported. Niri does have IPC. 

Support Niri workspaces out of the box. What would be required to do this?

anant-357:
Hi, please let me know if I've missed something. Resolves #650.


src/clients/compositor/niri.rs
```diff
@@ -0,0 +1,190 @@
...
+    }
+}
+
+impl WorkspaceClient for Client {
+    fn focus(&self, name: String) -> Result<()> {
+        await_sync(async {
+            let mut conn = Connection::connect().await.unwrap();
+
+            let command = Request::Action(Action::FocusWorkspace {
+                reference: WorkspaceReferenceArg::from_str(name.as_str()).unwrap(),
```
YaLTeR, line 30: Since most workspaces don't have names, this should instead focus by Id (not Idx). There's `WorkspaceReferenceArg::Id(u64)`.

src/clients/compositor/niri.rs
```diff
@@ -0,0 +1,190 @@
...
+
+    fn subscribe_workspace_change(
+        &self,
+    ) -> tokio::sync::broadcast::Receiver<super::WorkspaceUpdate> {
+        let (tx, rx) = channel(32);
+
+        await_sync(async {
+            let mut workspace_state = Vec::new();
+            let mut conn = Connection::connect().await.unwrap();
+            match conn.send(Request::Workspaces).await.unwrap().0.unwrap() {
```
YaLTeR, line 45: You shouldn't need this. The event stream always sends the full initial workspaces state at the start.

src/clients/compositor/niri.rs
```diff
@@ -0,0 +1,190 @@
...
+                            new_workspaces.sort_by_key(|w| w.id);
+                            workspace_state.sort_by_key(|w| w.id);
+                            let mut old_index = 0;
+                            let mut new_index = 0;
+                            while old_index < workspace_state.len()
+                                && new_index < new_workspaces.len()
+                            {
+                                let old_workspace = &workspace_state[old_index];
+                                let new_workspace = &new_workspaces[new_index];
+                                match old_workspace.id.cmp(&new_workspace.id) {
```
YaLTeR, line 79: I'm not entirely sure if this code relies on this, but you probably shouldn't rely on workspace id increasing incrementally and never wrapping back. That is an implementation detail subject to change if needed.

YaLTeR (commented):
Hi, only commenting on the niri IPC parts. I also clarified these in niri-ipc docs.

src/clients/compositor/niri.rs
anant-357: The focus method in the ```WorkspaceClient``` trait sends a name. So I would need some way to get the id from name. When converting from Niri Workspace struct to Ironbar Workspace struct, if the workspace has no name I set the id as name. Should I change the ```FromStr for WorkspaceReferenceArg``` to parse id instead of idx?

YaLTeR: It depends on what ironbar gives you, which I'm not familiar with. Index will work only if the correct monitor is focused when you send this request.

src/clients/compositor/niri.rs
anant-357: What would be a better way of checking for all changes? I am also sorting both workspace states before comparing ids.

anant-357: Ironbar has events like ```WorkspaceUpdate::Add(Workspace)``` and ```WorkspaceUpdate::Remove(id)``` so I cannot directly replace the workspace state. I will have to track changes from previous state.

src/clients/compositor/niri.rs
JakeStanger: I think it uses name because that was necessary for one of Sway or Hyprland. Using ID is definitely preferable. If necessary, I'd be okay splitting into a `FocusById` and `FocusByName` and letting each compositor impl pick the best case.

src/clients/compositor/niri.rs
JakeStanger: What does the `WorkspacesChanged` event signify in Niri? Is it just any addition/removal/rename/move?

JakeStanger:
Appreciate the hard work here, big thanks for taking this on. Cheers YaLTeR for reviewing too, means a lot to see the actual dev provide input. 

I'm a little busy at the moment so responses from me won't be as fast as I like, but ping me as soon as you need anything from me and I'll look as soon as I can. I'll give it a full review as soon as you're ready.

src/clients/compositor/niri.rs
YaLTeR: It signifies the new complete workspaces state. Anything could change. Basically an escape hatch event for changes not covered by other events, or when multiple things need to change at once. In practice, workspace additions/removals/reorders all use this because it would be too complex to handle them as separate events atomically.

JakeStanger: Right okay. So the cheap and dirty way of handling that on Ironbar's side would be to slightly adjust and reuse the `WorkspaceUpdate::Init` event, which takes the full state. The issue I have with that is that it's very expensive to dump and re-render the UI every time anything changes.

Tracking state changes on Ironbar's side *is* preferable, if not possible in Niri. At least for the cases where we can.

YaLTeR: In my Waybar PR I did a sequence of remove obsolete widgets, add new widgets, reorder all to update positions on a vector of workspaces pre-sorted by idx.

anant-357: The way I am handling this here is by keeping a previous state of workspaces and comparing with the new state for changes. To do this, first I sort both(old and new) workspace vectors based on id. Then I do a linear scan on the states at together. If the id is missing, then I send a ```WorkspaceUpdate::Remove``` event. If there is a new id then I send a ```WorkspaceUpdate::Add``` event. For workspaces with the same id, if the name of the workspace is different, ```WorkspaceUpdate::Rename``` is sent, if the name of the monitor is different then ```WorkspaceUpdate::Move``` is sent.

src/clients/compositor/niri.rs
```diff
@@ -0,0 +1,186 @@
...
+                        }
+                    }
+                    Event::Other => {
+                        vec![]
+                    }
+                };
+                for event in events {
+                    send!(tx, event);
+                }
+                std::thread::sleep(Duration::from_millis(30));
```
JakeStanger, line 181: As you are inside an async block, you should use Tokio's sleep to avoid blocking the thread.

src/clients/niri.rs
JakeStanger, line 1: As this is all compositor-focused I would prefer if all this code was under the `compositor` module for now. Happy to have a `niri` module with a couple of files in there.

src/clients/niri.rs
```diff
@@ -0,0 +1,126 @@
...
+        let stream = raw_stream;
+        Ok(Self(stream))
+    }
+
+    pub async fn send(
+        &mut self,
+        request: Request,
+    ) -> io::Result<(Reply, impl FnMut() -> io::Result<Event> + '_)> {
+        let Self(stream) = self;
+        let mut buf = serde_json::to_string(&request).unwrap();
```
JakeStanger, line 107: In cases like this, if you return a `color_eyre::Result` you should be able to automatically use the `?` operator despite mixing error types.

JakeStanger (changes requested):
I'm yet to test the PR, but generally this is looking good. There's a few points after an initial scan I'd like addressed, but nothing major. Thanks again for your hard work.

src/clients/compositor/niri/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+        Self
+    }
+}
+
+impl WorkspaceClient for Client {
+    fn focus(&self, name: String) -> Result<()> {
+        await_sync(async {
+            if let Ok(mut conn) = Connection::connect().await {
+                let command = Request::Action(Action::FocusWorkspace {
+                    reference: WorkspaceReferenceArg::from_str(name.as_str()).unwrap(),
```
JakeStanger, line 28: Please can this unwrap be changed. By the looks of it this is infallible so just need to say it should always be valid in an `expect`.

src/clients/compositor/niri/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+                                    // if focuesd is true then focus has changed => find old focused workspace. set it to inactive and set current
+                                    match workspace_state.iter().position(|w| w.id == id as i64) {
+                                        Some(new_index) => {
+                                            if focused {
+                                                match workspace_state
+                                                    .iter()
+                                                    .position(|w| w.visibility.is_focused())
+                                                {
+                                                    Some(old_index) => {
+                                                        workspace_state[new_index].visibility =
```
JakeStanger, line 146: There are a lot of lookups to to `workspace_state[new_index]` and `workspace_state[old_index]` throughout this block. Could these just be referenced as `new` and `old` variables pls to clean up a bit.

src/clients/compositor/niri/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+                                        }
+                                    }
+                                }
+                                Event::Other => {
+                                    vec![]
+                                }
+                            };
+                            for event in events {
+                                send!(tx, event);
+                            }
```
JakeStanger, line 208: Is there a risk of filling up the channel buffer here by sending too many events at once? Some stress testing may be required to check. If it is a potential problem, a 'batch' event might solve it.

JakeStanger (changes requested):
Looking like the code is almost there, thanks for your work & sorry it's taken me so long to have a look. I'll try and test soon so that we can get this in relatively quick for 0.17.0.

src/clients/compositor/niri/mod.rs
anant-357: I cannot have 2 mutable borrows from the ```workspace_state``` . Is there a particular way of doing this?

calops: I tried to address this point on my side and it really messes up the code for no clear benefit. I think it's better to leave it like this.

calops:
Hi there! Thanks for the work, this PR is exactly what I need. Is there anything I can do to help move it forward?

src/clients/compositor/niri/mod.rs
JakeStanger: Thought that'd probs be the case, happy to leave as-is then. Cheers for checking on behalf

JakeStanger:
Hey, the following are outstanding:

- [ ]  Any unresolved comments here
- [ ] Documentation updates
  - [ ] README - we now offer first-class support for Niri
  - [ ] Workspaces - module now supports Niri
- [ ] Testing

If you're able to offer a hand with any of the above at all, that'd be great. 

The plan is to ship 0.16.1 this weekend, assuming the tray doesn't find new ways to break before then. Once that's out, I'll merge this as soon as it's ready.

anant-357:
There is an issue. If I remove a workspace from the middle, somehow when I move to another workspace, the focused workspace on the bar does not change. I am unable to pin point why this is happening. I also do not have multiple monitors, so it would be helpful if anyone could test for that.

calops:
I'd love to tackle these issues, but I don't know the best way to do it on a PR that comes from a fork that I don't own.

And for the comment about stress testing, I'm also not sure how to go about this.

JakeStanger:
You may be able to open a PR on the fork, targeting the branch used for this PR. If that gets merged, it'll update here. If not you can just send a file.

Spamming the Niri IPC (aka a while true loop) with workspace changes and ensuring Ironbar can keep up is sufficient for stress testing.

JakeStanger:
I'm closing in favour of #871, so I can rebase and continue development to get this across the line, as this being on the master branch with merge commits makes that difficult. You'll still be credited.

As it stands, that PR is identical to this one, albeit rebased onto the latest master commit and squashed.

Issue #731, opened 2024-09-19
Iron bar crash on clicking a specific workspace 3 times in a row 


RedStealthDev:
**Describe the bug**
> A clear and concise description of what the bug is.
in my bar i use the workspace module if if a work space is clicked 3 times in a row fast it crashes ironbar

**To Reproduce**
1. have a bar with work spaces
2. click a work space button 3 times in a row fast
3. observe the crash

**Expected behavior**
> A clear and concise description of what you expected to happen.
continues working

**System information:**
 - Distro: arch linux 6.10.10-zen
- Compositor: Hyprland v0.43.0
 - Ironbar version: 0.16.1-pre

**Configuration**
> Share your bar configuration and stylesheet as applicable:

<details><summary>Config</summary>

```position: "top"
height: 42
start:
  - type: workspaces
    all_monitors: false
center:
end:
  - type: music
    player_type: mpris
  - type: sys_info
    interval:
      memory: 30
      cpu: 1
      temps: 5
    format:
      - 'mem: {memory_used}/{memory_total}gb'
      - 'cpu: {cpu_percent}%|{temp_c:k10temp-Tctl}°C'
    direction: v
  - type: volume
    format: '{icon}{percentage}%'
    max_volume: 100
    icons:
      volume_high: 🔊
      volume_medium: 🔉
      volume_low: 🔈
      muted: 🔇
  - type: label
    name: bright
    label: '☀️{{poll:1000:echo "($(brightnessctl g) / $(brightnessctl m)) * 100" | bc -l | sed "s/\..*//"}}'
  - type: upower
    format: '{percentage}%'
  - type: clock

```
</details>

<details><summary>Styles</summary>

```.background{
	background-color: #fcd5c3;
	border: 3px solid #3ea49f;
	border-radius: 5px;
	opacity: 1;
}
* {
	text-shadow: none;
	box-shadow: none;
}
.workspaces .item.visible{
	background-color: #989b94;
}
.workspaces .item{
	background-color: #b8bbb4;
	border: 3px solid #3ea49f;
	border-radius: 5px;
	color: #b8bbb4;
	background-image: none;
}
.sysinfo {
	background-color: #b8bbb4;
	border: 3px solid #3ea49f;
	border-radius: 5px;
}
.sysinfo .item:first-child{
	margin-bottom: -10px;
}
.sysinfo .item{
	margin: 1px;
}
.volume {
	background-color: #b8bbb4;
	border: 3px solid #3ea49f;
[...]

PR #733, opened 2024-09-23
fix(workspaces): prevent crash when clicking current workspace

Leshuguita:
Should solve https://github.com/JakeStanger/ironbar/issues/731 by printing a warning instead of stopping the receiving loop when focusing a workspace.


JakeStanger:
Thanks for the fix. 

I'm going to merge this, but leave the issue open for now as I'd like to investigate *why* the channel is filling up and causing the problem in the first place. For now though, avoiding the crash is better than nothing.

PR #735, opened 2024-09-26
feat: add `.urgent` workspace css class

Rodrigodd:
Allow highlighting a workspace which contains a window that triggered a urgent event, i.e., requests attention. In my case, I see it happens when I clink a link and it opens in a new tab on firefox, and when displaying a `BEL` character in some terminal emulators (at least in alacritty).

You can test it by running

```sh
sleep 1 && tput bel
```

in alacritty, and switching to another workspaces before the sleep expires.

I implemented it for both Sway and Hyprland. Sway keep track of a `urgent` flag for each client, and notify ironbar about changes. Hyprland or the other hand only contains a urgent event that triggers once, so I clean the urgent state of a workspace whenever it is focused.

### DIFF
--- a/src/clients/compositor/hyprland.rs
+++ b/src/clients/compositor/hyprland.rs
@@ -166,10 +166,42 @@ impl Client {
             }
 
             {
+                let tx = tx.clone();
+                let lock = lock.clone();
+
                 event_listener.add_workspace_destroy_handler(move |data| {
                     let _lock = lock!(lock);
                     debug!("Received workspace destroy: {data:?}");
                     send!(tx, WorkspaceUpdate::Remove(data.workspace_id as i64));
+                });
+            }
+
+            {
+                event_listener.add_urgent_state_handler(move |address| {
+                    let _lock = lock!(lock);
+                    debug!("Received urgent state: {address:?}");
+
+                    let clients = match hyprland::data::Clients::get() {
+                        Ok(clients) => clients,
+                        Err(err) => {
+                            error!("Failed to get clients: {err}");
+                            return;
+                        }
+                    };
+                    clients.iter().find(|c| c.address == address).map_or_else(
+                        || {
+                            error!("Unable to locate client");
+                        },
+                        |c| {
+                            send!(
+                                tx,
+                                WorkspaceUpdate::Urgent {
+                                    id: c.workspace.id as i64,
+                                    urgent: true,
+                                }
+                            );
+                        },
+                    );
                 });
             }
 
@@ -191,6 +223,14 @@ impl Client {
             WorkspaceUpdate::Focus {
                 old: prev_workspace.take(),
                 new: workspace.clone(),
+            }
+        );
+
+        send!(
+            tx,
+            WorkspaceUpdate::Urgent {
+                id: workspace.id,
+                urgent: false,
             }
         );
 
--- a/src/clients/compositor/sway.rs
+++ b/src/clients/compositor/sway.rs
@@ -109,6 +109,16 @@ impl From<WorkspaceEvent> for WorkspaceUpdate {
             WorkspaceChange::Move => {
                 Self::Move(event.current.expect("Missing current workspace").into())
             }
+            WorkspaceChange::Urgent => {
+                if let Some(node) = event.current {
+                    Self::Urgent {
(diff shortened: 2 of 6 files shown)


src/clients/compositor/hyprland.rs
```diff
@@ -166,13 +166,45 @@ impl Client {
...
                     let _lock = lock!(lock);
                     debug!("Received workspace destroy: {data:?}");
                     send!(tx, WorkspaceUpdate::Remove(data.workspace_id as i64));
                 });
             }
 
+            {
+                event_listener.add_urgent_state_handler(move |address| {
+                    let _lock = lock!(lock);
+                    debug!("Received workspace destroy: {address:?}");
```
yavko, line 182: This string is wrong

Rodrigodd: @yavko Fixed

src/modules/workspaces.rs
```diff
@@ -413,6 +413,17 @@ impl Module<gtk::Box> for WorkspacesModule {
                             }
                         }
                     }
+                    WorkspaceUpdate::Urgent{id, urgent} => {
+                        let button = button_map.get(&id);
+                        if let Some(item) = button {
+                            if urgent  {
+                            // if fav_names.contains(&workspace) {
```
yavko, line 420: Prob should uncomment

Rodrigodd: @yavko done

src/clients/compositor/sway.rs
```diff
@@ -109,6 +109,17 @@ impl From<WorkspaceEvent> for WorkspaceUpdate {
             WorkspaceChange::Move => {
                 Self::Move(event.current.expect("Missing current workspace").into())
             }
+            WorkspaceChange::Urgent => {
+                dbg!(&event);
```
yavko, line 113: Sorry for all the comments, but r u sure this should stay?

Rodrigodd: Yeah, no, it should not stay (I should have revised everything before publishing, sorry).

I will only be able to fix it in 9 hours. Feel free to fix it yourselve, if you prefer.

src/modules/workspaces.rs
```diff
@@ -413,6 +413,16 @@ impl Module<gtk::Box> for WorkspacesModule {
                             }
                         }
                     }
+                    WorkspaceUpdate::Urgent { id, urgent } => {
+                        let button = button_map.get(&id);
+                        if let Some(item) = button {
+                            if urgent {
+                                item.style_context().add_class("urgent");
```
JakeStanger, line 420: I have an extension trait which provides a method to shorten this a bit to `item.add_class()`
https://github.com/JakeStanger/ironbar/blob/e08027fe6992756f8cb4800ec19782024543b19b/src/gtk_helpers.rs#L30-L33

JakeStanger:
Thanks for this. This looks good, I'll test it when I can and will approve once I've confirmed it working as expected.

src/modules/workspaces.rs
Rodrigodd: Yeah, in an earlier iteration I had added a `remove_class` to the extension trait, but I reverted it because it was too unrelated to this PR (I would need to update all uses of `remove_class`, and it appears you weren't using the `add_class` extension everywhere).

~~I may have introduced the `style_context` here by accident in the process~~. Actually, the match arm above also doesn't use it either.

yavko (approved):
Looks fine to me

src/modules/workspaces.rs
JakeStanger: There's only a few instances of `remove_class` in the codebase which is why I never bothered. There's a few dangling `style_context.add_class` references which I've never switched out but I try to avoid introducing new ones, and switch out the old ones when I see them.

Rodrigodd: Understandable. Pushed a fix for that.

JakeStanger (approved):

JakeStanger:
Cheers, this looks good. 

The next release will be bugfix-only, so I'm going to hold off merging for the time being. Once #748 merges, this will need to be slightly reworked, which is poor planning on my part. I'll pick that up to avoid creating you too much more work.

PR #747, opened 2024-10-16
Github Action for automatically deploying to the AUR

Punff:
it's been bugging me for a while that the AUR package is so behind so here it is. It requires adjustment of the AUR credentials and adding the AUR ssh key as a secret. It's my first pull request so I hope I haven't done something horribly wrong. Cheers


JakeStanger:
Hey, thanks for this but this isn't necessary. `-git` packages on the AUR will always pull the latest commit, the version number only gets bumped when the PKGBUILD changes.

If you're using an AUR helper (eg yay or paru) you can configure these to automatically track and update git packages for you to ensure you're always on the latest commit.

PR #748, opened 2024-10-18
Workspaces module rewrite

### LINKED ISSUE #705: Favourite-Workspace that was focused during start of ironbar disappears when switching to a different workspace
**Describe the bug**
When adding favorites in the workspace module, the workspaces that were active when ironbar starts disappear when empty or emptied and I'm switching to a different workspace on that screen.
They _stay_ hidden when trying to switch back.

**To Reproduce**
1.
  ```toml
  [[start]]
  type = "workspaces"
  favorites = [ "1", "2" ]
  ```
2. switch to workspace 1 
3. Open something on ws-1
4. start ironbar
 => ws-1 stays visible because something is open
5. close whatever is on workspace 1
6. switch to another workspace on that screen
=> workspace 1 disappears even though it is listed in favourites
7. switch back to empty workspace 1 and open something
=> workspace 1 _stays_ hidden, even though it should not

**Expected behavior**
All workspaces listed in the favorites array should be visible in the bar at all times, even when empty.

**System information:**
 - Distro: Arch Linux
 - Compositor: Sway
 - Ironbar version: 0.16.1-pre

**Configuration**
> Share your bar configuration and stylesheet as applicable:

```toml
position = "top"
height = 24

[[start]]
type = "workspaces"
# all_monitors = false
# favorites = { "DVI-D-1"=[ "1", "2", "3", "4" ] }
favorites = [ "1", "2" ]

center=[]

[[end]]
type = "tray"
direction = "right_to_left"

[[end]]
type = "music"
format = "{title}"

[[end]]
type = "clock"
format = "%d/%b/%Y | %H:%M:%S"

[[end]]
type = "notifications"
show_count = true

```
</details>

issue persists even when using no stylesheet

**Additional context**

**Screenshots**

JakeStanger:
fix(workspaces): rewrite module to fix several small issues

Rewrites the module code to be better structured, in a similar pattern to the launcher. The code is now more robust and more maintainable, yay!

Fixes #705

Fixes an issue with moving favourite workspaces.

Fixes an issue with workspace visible state being incorrect.

Fixes an issue where the `inactive` class looked at hidden instead of closed favourites.


JakeStanger:
No idea why the build is failing, seems GH have messed up their task runner

chmanie:
I'm adding a question here as this also seems to fix issues that are not (yet) in GitHub: I'm on Hyprland and I'm having the issue that after a while the focused workspace is "stuck" at a certain number and won't change until I issue `ironbar reload`. Everything else (like hover) seems to still work fine.

src/modules/workspaces/mod.rs
```diff
@@ -0,0 +1,378 @@
...
+
+/// Re-orders the container children alphabetically,
+/// using their widget names.
+///
+/// Named workspaces are always sorted before numbered ones.
+fn reorder_workspaces(container: &gtk::Box) {
+    let mut buttons = container
+        .children()
+        .into_iter()
+        .map(|child| (child.widget_name().to_string(), child))
```
JakeStanger, line 143: Will need updating to `child.label()` (see https://github.com/JakeStanger/ironbar/pull/799#issuecomment-2513110838)

PR #769, opened 2024-11-10
feat: Minimize focused window option for launcher

### LINKED ISSUE #658: [Launcher] Focused windows should minimize when ironbar launcher icon is clicked
Maybe there's an option to make this function as I expect it to. Most docks seem to operate this way:

* Click application icon to focus window
* Click same icon again to minimize window

Is it possible to do this currently?

pachliopta:
This PR adds a new config option `minimize_focused` to the launcher module, which when set to true will minimize the currently focused window if its icon is clicked. If there is more than a single window open of that particular application the normal focus behavior will occur.

I haven't written much rust before, but hopefully it's relatively problem free!

Resolves  #658

### DIFF
--- a/src/modules/launcher/mod.rs
+++ b/src/modules/launcher/mod.rs
@@ -56,6 +56,12 @@ pub struct LauncherModule {
     #[serde(default = "crate::config::default_false")]
     reversed: bool,
 
+    /// Whether to minimize a window if it is focused when clicked.
+    ///
+    /// **Default**: `true`
+    #[serde(default = "crate::config::default_true")]
+    minimize_focused: bool,
+
     // -- common --
     /// Truncate application names on the bar if they get too long.
     /// See [truncate options](module-level-options#truncate-mode).
@@ -111,6 +117,7 @@ pub enum ItemEvent {
     FocusItem(String),
     FocusWindow(usize),
     OpenItem(String),
+    MinimizeItem(String),
 }
 
 enum ItemOrWindow {
@@ -290,6 +297,7 @@ impl Module<gtk::Box> for LauncherModule {
         });
 
         // listen to ui events
+        let minimize_focused = self.minimize_focused;
         let wl = context.client::<wayland::Client>();
         spawn(async move {
             while let Some(event) = rx.recv().await {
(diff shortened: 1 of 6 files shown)


src/modules/launcher/mod.rs
```diff
@@ -54,6 +54,12 @@ pub struct LauncherModule {
     #[serde(default = "crate::config::default_false")]
     reversed: bool,
 
+    /// Whether to minimize a window if it is focused when clicked.
+    ///
+    /// **Default**: `false`
+    #[serde(default = "crate::config::default_false")]
```
JakeStanger, line 60: I reckon this should actually be on by default

src/modules/launcher/mod.rs
```diff
@@ -292,19 +300,26 @@ impl Module<gtk::Box> for LauncherModule {
                     );
                 } else {
                     send_async!(tx, ModuleUpdateEvent::ClosePopup);
+                    let mut minimize_window = false;
```
JakeStanger, line 303: I think it'd be slightly cleaner to calculate this in-place, like:

```suggestion
                    let mut minimize_window = false;
                    let minimize_window = matches!(event, ItemEvent::MinimizeItem(_));
```

(untested but I think that should work)

src/modules/launcher/mod.rs
```diff
@@ -292,19 +300,26 @@ impl Module<gtk::Box> for LauncherModule {
...
-                        ItemEvent::FocusItem(app_id) => {
-                            lock!(items).get(&app_id).and_then(|item| {
-                                item.windows
-                                    .iter()
-                                    .find(|(_, win)| !win.open_state.is_focused())
-                                    .or_else(|| item.windows.first())
-                                    .map(|(_, win)| win.id)
-                            })
-                        }
+                        ItemEvent::FocusItem(app_id) => search_items(app_id),
```
JakeStanger, line 316: Based on the prev comment, that would allow you to combine the focus/minimize branches here:

```suggestion
                        ItemEvent::FocusItem(app_id) => search_items(app_id),
                        ItemEvent::FocusItem(app_id) | ItemEvent::MinimizeItem => search_items(app_id),
```

JakeStanger (changes requested):
Thanks for this, this all looks great. I've raised a few minor points if you could have a look at those please.

I will be looking at merging this for 0.17, as soon as 0.16.1 is released.

src/modules/launcher/mod.rs
pachliopta: Makes sense! I wanted them on the same branch but I preoccupied myself hoping for some esoteric solution that just doing it in place never crossed my mind

pachliopta:
Changes should be addressed now, thank you for taking a look!

JakeStanger (commented):
lgtm, cheers. I'll give a quick test when I can & approve when confirmed working.

JakeStanger (approved):

PR #777, opened 2024-11-12
Allow IPC to act on multiple bars by the same name

### LINKED ISSUE #770: `ironbar bar <name> hide` doesn't hide all bars by the same name
Hi, glad to see this project has evolved so well :)
I'm finally trying to move from swaybar, but:

With a simple config on a multimonitor setup:

```toml
name = "bar"
start = []
center = []
end = []
```

invoking the cli to hide the bar that is on multiple monitors:

`ironbar bar bar hide`

only hides a bar on a single monitor, instead of all.

zeroeightysix:
Closes #770

### DIFF
--- a/src/ipc/server/bar.rs
+++ b/src/ipc/server/bar.rs
@@ -13,7 +13,7 @@ pub fn handle_command(command: BarCommand, ironbar: &Rc<Ironbar>) -> Response {
         return Response::error("Invalid bar name");
     }
 
-    let mut responses = bars
+    let responses = bars
         .into_iter()
         .map(|bar| match &command.subcommand {
             Show => set_visible(&bar, true),
--- a/src/ipc/server/bar.rs
+++ b/src/ipc/server/bar.rs
@@ -8,18 +8,21 @@ use std::rc::Rc;
 pub fn handle_command(command: BarCommand, ironbar: &Rc<Ironbar>) -> Response {
     use BarCommandType::*;
 
-    let bars = ironbar.bar_by_name(&command.name);
+    let bars = ironbar.bars_by_name(&command.name);
     if bars.is_empty() {
         return Response::error("Invalid bar name");
     }
 
-    let mut responses = bars.into_iter()
+    let mut responses = bars
+        .into_iter()
         .map(|bar| match &command.subcommand {
             Show => set_visible(&bar, true),
             Hide => set_visible(&bar, false),
             SetVisible { visible } => set_visible(&bar, *visible),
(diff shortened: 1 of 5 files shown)


JakeStanger (changes requested):
Thanks for this. I've flagged some changes I think should go in first. Please also ensure the styling is fixed.

Documentation will also need updating - are you happy to take this on?

zeroeightysix:
> Documentation will also need updating - are you happy to take this on?

Sure! To make sure I don't miss anything, where are the places I should be looking?

JakeStanger:
It should just be the [Controlling Ironbar](https://github.com/JakeStanger/ironbar/wiki/controlling-ironbar) page

src/ipc/server/bar.rs
```diff
@@ -8,48 +8,68 @@ use std::rc::Rc;
...
-            bar.set_exclusive(exclusive);
+            GetPopupVisible => Response::OkValue {
+                value: bar.popup().visible().to_string(),
+            },
+            SetExclusive { exclusive } => {
+                bar.set_exclusive(*exclusive);
+                Response::Ok
+            }
+        })
+        .collect::<Vec<_>>();
```
JakeStanger, line 55: As you're immediately iterating over the responses, it'd be better to avoid collecting & allocating a vec, and instead tag the below `reduce` on straight after the map

src/ipc/server/bar.rs
```diff
@@ -8,48 +8,68 @@ use std::rc::Rc;
...
+            (Response::OkValue { value: v1 }, Response::OkValue { value: v2 }) => Response::Multi {
+                values: vec![v1, v2],
+            },
+            (Response::Multi { mut values }, Response::OkValue { value: v }) => {
+                values.push(v);
+                Response::Multi { values }
+            }
+            _ => unreachable!(),
+        })
+        .unwrap()
```
JakeStanger, line 72: Please avoid `unwrap`. It looks like here there's no possible path for the program to panic, so instead use `expect` & explain why

JakeStanger (changes requested):
Cheers for changes, just a couple of little bits more when you can pls

src/ipc/server/bar.rs
zeroeightysix: Although slightly complicating the flow, I might as well use `unwrap_or` and remove the length check at the beginning

JakeStanger (commented):
Perfect, thanks, code LGTM. I'll test shortly & approve if all seems good!

JakeStanger (approved):

PR #799, opened 2024-12-02
sway: support workspace rename events

quietvoid:
Simple enough to propagate the event of the rename.
However I'm not sure how to fix the order on rename, it doesn't seem to work if I just add
```rust
if self.sort == SortOrder::Alphanumeric {
    reorder_workspaces(&container);
}
```

### DIFF
--- a/src/clients/compositor/sway.rs
+++ b/src/clients/compositor/sway.rs
@@ -109,6 +109,16 @@ impl From<WorkspaceEvent> for WorkspaceUpdate {
             WorkspaceChange::Move => {
                 Self::Move(event.current.expect("Missing current workspace").into())
             }
+            WorkspaceChange::Rename => {
+                if let Some(node) = event.current {
+                    Self::Rename {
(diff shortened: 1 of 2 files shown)


quietvoid:
It seems the ordering relies on `widget_name` while we're only changing the label.
Not sure which should be done of:
- Also changing `widget_name` and reorder
- Reorder based on label

JakeStanger:
Hm, I can't think why it would need to be tied to the widget name. Sorting by label instead should be fine.

quietvoid:
Here we go. The reorder prioritizes the label if the widget is indeed a button, otherwise still uses the widget name.
I'm not too sure what other widgets are present in the container and if this could have a negative effect.
I think the workspace box is all buttons but I may be wrong.

JakeStanger (approved):
Thanks, LGTM. Everything in the container should always be a button, and should always have a label, but having the fallback is definitely good.

PR #802, opened 2024-12-07
Workspace ordering allowing for names of kind <number>:<_rest> to be deterministic. 

PoutineSyropErable:
…back to alphanumeric sorting then, which is locale dependant?? Works differently on differnt machine. (Workspace 10:<icon> is first on one machine, and last one another. I prefer last . So, I added some code to check if it can be n:<rest> and then order with that n. It should try to order following <n>, <n>:_rest, lexixographical fallback.

for stuff like that:

		name_map.1 = "1: "  
		name_map.2 = "2: "  
		name_map.3 = "3: "  
		name_map.4 = "4: "  
		name_map.5 = "5:󰄨 "  
		name_map.6 = "6: "  
		name_map.7 = "7: "  
		name_map.8 = "8: "  
		name_map.9 = "9:󰯂 "  
		name_map.10 = "10: " 
		
Now, 10 is consistently at the end.

### DIFF
--- a/src/modules/workspaces.rs
+++ b/src/modules/workspaces.rs
@@ -160,6 +160,16 @@ fn create_button(
     button
 }
 
+// Function to extract numeric prefix and remaining string
+fn extract_numeric_and_rest(label: &str) -> (Option<i32>, &str) {
+    if let Some((prefix, rest)) = label.split_once(':') {
+        if let Ok(number) = prefix.trim().parse::<i32>() {
+            return (Some(number), rest); // Successfully parsed number
+        }
+    }
+    (None, label) // Return None if no numeric prefix is found
+}
+
 fn reorder_workspaces(container: &gtk::Box) {
     let mut buttons = container
         .children()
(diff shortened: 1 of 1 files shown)


JakeStanger:
While I think having some form of solution for this is good, I'm going to close this PR because the implementation is specific to your configuration. Any other format won't work with it.

This looks like it may have come about due to the switch from sorting by name to sorting by label here https://github.com/JakeStanger/ironbar/pull/799? If that's the case then perhaps a better solution would be to add a new sorting option, perhaps `alphanumeric_name` that uses the name from the compositor instead of the `name_map` label.

PoutineSyropErable:
How yeah, i just quickly glued it together. Idk how the comment was incomplete. 
Basically, I have the names be <number>:<icon>, which isn't a number so it falls back to lexicographical ordering. 
Weirdly enough, on both machine, the lexi ordering differed. 

On my laptop, where I had US_en locale from the beginning, it worked correctly. 
But on my desktop, where I had french locale it was different, and even after having the exact same locale and then resinstalling it didn't work. I don't know what other config file was different. 

It worked on the same pc before I had to reinstall everything due to filesystem corruption. 

For myself, I quickly implemented <number>:<rest> (where the cutoff is :) using chatgpt, since I'm not really a rust user for now. 


Anyway, as such, Other sorting order like (just take the numbers in the name and assume its one number), might be useful. 

example: 1aabb2 -> 112 (just numbers)
 

or just take the leftmost or rightmost number

11aabb2 -> 11 (right)
or 
11aabb2 -> 2 (left)

Just a feature request.

Issue #809, opened 2024-12-09
vlc tray icon not respecting `icon_size`


salahoued:
**Describe the bug**
> A clear and concise description of what the bug is.

I'm not sure if this is an `ironbar` issue or a vlc one, but when vlc (flatpak) is launched, its tray icon is shown in `ironbar` too big which causes the bar height to grow big

**To Reproduce**
1. open vlc
2. observe the tray icon

**Expected behavior**
> A clear and concise description of what you expected to happen.

tray icons should respect the `icon_size` option in the config, or if not declared should respect the bar `height`

**System information:**
 - Distro: openSUSE Tumblweed
- Compositor: Hyprland
 - Ironbar version: 0.16.1

**Configuration**
> Share your bar configuration and stylesheet as applicable:

<details><summary>Config</summary>

```
let {
  $position = "top"
  $height = 20
  $margin = {
    top = 0
    left = 0
    right = 0
  }
  
  $clock = {
    type = "clock"
    name = "time-widget"
    format = "<span font-weight='bold'>%H:%M</span>, %a %d-%m-%y"
  }

  $launcher = {
    type = "launcher"
    show_names = false
    show_icons = true
    icon_size = 16
    reversed = false
  }

  $focused = {
    name = "focused-window"
    type = "focused"
    show_icon = true
    show_title = true
    icon_size = 16
    truncate = "middle"
  }

  $sys_info = {
    type = "sys_info"

    interval.memory = 10
    interval.cpu = 1
    interval.temps = 5
    interval.networks = 3


    format = [
      "C: {cpu_percent}%"
      "R: {memory_percent}%"
      "T: {temp_c:coretemp-Package-id-0}°C"
    ]
  }

  $upower = {
    type = "upower"
    icon_size = 16
  }

  $clipboard = {
      type = "clipboard"
      icon = "icon:edit-paste"
      icon_size = 16
      max_items = 10
      truncate.mode = "end"
      truncate.length = 50
    }

  $tray = {
    type = "tray"
    icon_size = 16
  }

  $notifs = {
    type = "notifications"
    show_count = true

    icons.closed_none = "󰍥"
    icons.closed_some = "󱥂"
    icons.closed_dnd = "󱅯"
    icons.open_none = "󰍡"
    icons.open_some = "󱥁"
    icons.open_dnd = "󱅮"

  }

  $logimage = {
    type = "image"
    src = "system-shutdown"
    name = "logimage"
    size = 20
  }

  $logbutton = {
[...]

PR #812, opened 2024-12-13
Force scaling icon

ogios:
fix https://github.com/JakeStanger/ironbar/issues/809

### DIFF
--- a/src/modules/tray/icon.rs
+++ b/src/modules/tray/icon.rs
@@ -1,11 +1,12 @@
 use crate::image::ImageProvider;
 use crate::modules::tray::interface::TrayMenu;
+use cairo::ImageSurface;
 use color_eyre::{Report, Result};
 use glib::ffi::g_strfreev;
 use glib::translate::ToGlibPtr;
 use gtk::ffi::gtk_icon_theme_get_search_path;
 use gtk::gdk_pixbuf::{Colorspace, InterpType, Pixbuf};
-use gtk::prelude::IconThemeExt;
+use gtk::prelude::{GdkContextExt, IconThemeExt};
 use gtk::{IconLookupFlags, IconTheme, Image};
 use std::collections::HashSet;
 use std::ffi::CStr;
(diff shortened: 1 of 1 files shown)


ogios:
and also, should we find a proper size in pixmap function instead of just use the first one?
https://github.com/JakeStanger/ironbar/blob/8a06ec04b098d662188426d7eaf0e9d108725cc5/src/modules/tray/icon.rs#L90

JakeStanger:
Thanks for this, haven't given a full review yet but looks good.

I think it would make sense to pick the next biggest size and scale that down. That'd be cheaper than always scaling the largest.

ogios:
> I think it would make sense to pick the next biggest size and scale that down.   

i've added the functionality for this if i understand correctly.

One thing bothering me is that, whether the `Vec<IconPixmap>` is sorted with height, if so then we can use a better way to speed up the search.

JakeStanger:
> One thing bothering me is that, whether the Vec<IconPixmap> is sorted with height, if so then we can use a better way to speed up the search.

The spec is so incredibly vague and has so many incorrect implementations that I'd keep away from trying to do anything like that. The good news is in reality there will only be a few entries on rare occasions so the search will be fast enough always.

src/modules/tray/icon.rs
```diff
@@ -44,17 +46,42 @@ pub fn get_image(
...
+
+    Ok(image)
+}
+
+fn scale_image_to_height(pixbuf: Pixbuf, size: i32) -> ImageSurface {
+    let scale = size as f64 / pixbuf.height() as f64;
+    let width = (pixbuf.width() as f64 * scale).ceil() as i32;
+    let height = (pixbuf.height() as f64 * scale).ceil() as i32;
+
+    let surf = ImageSurface::create(cairo::Format::ARgb32, width, height).unwrap();
```
JakeStanger: Please don't use unwrap in prod code - this should either use `expect` if there is no scenario in which it should realistically go wrong, or return a result if it can. Same with every case pls

src/modules/tray/icon.rs
```diff
@@ -81,13 +106,15 @@ fn get_image_from_icon_name(item: &TrayMenu, icon_theme: &IconTheme, size: u32)
...
-fn get_image_from_pixmap(item: &TrayMenu, size: u32) -> Result<Image> {
+fn get_image_from_pixmap(item: &TrayMenu, size: u32) -> Result<Pixbuf> {
     const BITS_PER_SAMPLE: i32 = 8;
 
     let pixmap = item
         .icon_pixmap
         .as_ref()
-        .and_then(|pixmap| pixmap.first())
+        // The vec is sorted(ASC) with size(width==height) most of the time,
+        // but we can not be sure that it'll always sorted by `height`
```
JakeStanger, line 117: There's a slight catch here that ideally needs to be incorporated. I'm hoping it shouldn't be too hard as it's just a switch on the value to check, but for vertical bars the bar "height" property actually refers to its width. We'd need to perform all operations here off the width instead.

JakeStanger (changes requested):
I'm really sorry, going through old PRs and I didn't realise I left this review in draft. Let me know if you're still up for picking up the changes, if not I'll get it finished up and merged

src/modules/tray/icon.rs
ogios: changed in 455545e5d673

src/modules/tray/icon.rs
ogios: changed to use width in find_approx_size 455545e5d673

JakeStanger:
Rebasing this onto the GTK4 rewrite proved difficult so I've manually picked out your `find_approx_size` function and tests and created a new commit. You're co-author in the commit.

So sorry this PR didn't get my full attention and slipped past me a few times, but many thanks for sending in the fix.

Issue #813, opened 2024-12-15
Launcher: Middle click to open new window


BowDown097:
**Is your feature request related to a problem? Please describe.**
Opening a window when one is already open is currently pretty inconvenient, as there is no way to do so other than manually launching from outside of the launcher. This feature would be the most effective way to tackle this problem.

**Describe the solution you'd like**
An implementation of this feature - middle clicking an icon opening a new window, whether or not one is already open.

**Describe alternatives you've considered**
Launching from another source like rofi is the only way to do this currently. Another way this could be done is to have it as an action in a right-click menu, which I see has already been suggested in #247, but is not as convenient.

**Additional context**
This is a feature present in many other launchers.

BowDown097:
Okay, actually I've just gone and made a PR.

PR #814, opened 2024-12-15
feat(launcher): open new window on middle-click

BowDown097:
See the (now closed) #813 for more info.

### DIFF
--- a/src/modules/launcher/item.rs
+++ b/src/modules/launcher/item.rs
@@ -7,6 +7,7 @@ use crate::modules::launcher::{ItemEvent, LauncherUpdate};
 use crate::modules::ModuleUpdateEvent;
 use crate::{read_lock, try_send};
 use glib::Propagation;
+use gtk::gdk::{BUTTON_MIDDLE, BUTTON_PRIMARY};
 use gtk::prelude::*;
 use gtk::{Button, IconTheme, Image, Label, Orientation};
 use indexmap::IndexMap;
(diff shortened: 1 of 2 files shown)


JakeStanger:
Agree this makes sense to add. Could you just add a quick sentence to the docs to explain the new behaviour please?

src/modules/launcher/item.rs
```diff
@@ -201,20 +202,27 @@ impl ItemButton {
...
-                        try_send!(tx, ItemEvent::MinimizeItem(app_id.clone()));
+
+            button.connect_button_release_event(move |button, event| {
+                if event.button() == BUTTON_PRIMARY {
+                    // lazy check :| TODO: Improve this
+                    let style_context = button.style_context();
+                    if style_context.has_class("open") {
+                        let menu_state = read_lock!(menu_state);
+
+                        if style_context.has_class("focused") && menu_state.num_windows == 1 {
```
BowDown097, line 213: I've also noticed this - is clicking to minimize having no effect if a program has multiple instances an intentional design choice? Seems a bit odd, but if it's intentional, it's fine.

JakeStanger: The current (intended) behaviour is that clicking when there are multiple windows cycles them instead, so minimising in that case doesn't make much sense.

BowDown097: Oh, well that doesn't happen for me (in this release version too, nothing introduced by this PR), it just focuses the first window in the group. Maybe that's something worth a separate issue. That's all I have to say otherwise. I updated the docs page, hope it's fine.

JakeStanger: Yes, sounds like a separate bug to me. I'll open an issue for that

JakeStanger:
LGTM, I'll give this a quick test when I can & approve if all working. Cheers

JakeStanger:
There are some bugs around the behaviour of this, but they're related to gtk-launch and long-standing issues with the launcher. I'll need to address those at a later date, but I think as it stands this PR works as well as it can within the system so I'll merge. Cheers again.

JakeStanger (approved):

PR #823, opened 2024-12-28
refactor: replace channel macros with ext trait methods

JakeStanger:
Big refactor. Replaces all the macros related to channels (send/receive) with traits and extension methods, which is a little bit cleaner and should give the tooling a better shot.


JakeStanger:
There are potentially some weird teething issues around race conditions or channels misbehaving with this, so keeping an eye for a few days before merging.

PR #829, opened 2025-01-01
feat: justify property for clock and custom label

BowDown097:
Provides a way for the user to align text in these modules. Particularly useful with multi-line text. Equivalent to this would be CSS's ``text-align`` property (which, for God knows what reason, GTK CSS does not provide).

These seemed to be the best fitting modules to add this property to, but I'm open to adding it to more if there are any others that seem deserving of it to someone else.

### DIFF
--- a/src/modules/clock.rs
+++ b/src/modules/clock.rs
@@ -58,10 +58,6 @@ pub struct ClockModule {
     #[serde(default)]
     orientation: ModuleOrientation,
 
-    /// See [common options](module-level-options#common-options).
-    #[serde(flatten)]
-    pub common: Option<CommonConfig>,
-
     /// The justification (alignment) of the date/time shown on the bar.
     ///
     /// **Valid options**: `left`, `right`, `center`, `fill`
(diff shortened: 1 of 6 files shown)


BowDown097:
Here's a before/after adding ``justify: "center"`` to a multi-line clock:
Screenshot_20250101_123115
Screenshot_20250101_123128

JakeStanger:
Cheers, this seems like a good addition to me :)

src/modules/clock.rs
```diff
@@ -61,6 +61,14 @@ pub struct ClockModule {
...
     #[serde(flatten)]
     pub common: Option<CommonConfig>,
+
+    /// The justification (alignment) of the date/time shown on the bar.
+    ///
+    /// **Valid options**: `left`, `right`, `center`, `fill`
+    /// <br>
+    /// **Default**: `left`
+    #[serde(default)]
+    justify: ModuleJustification,
```
JakeStanger, line 71: Can this be moved above `common` please? I like to keep that last as it's a "special" property

JakeStanger (changes requested):

JakeStanger (approved):

JakeStanger:
Whoops I shoulda squashed that. Ah well, enjoy multiple changelog entries lol

Issue #575, opened 2024-05-08
Language/keyboard layout module


madfanat:
**Request**
It would be great to have a built-in module similar to Waybar’s "sway/language" or "hyprland/language".

JakeStanger:
Could you expand on the request please, as I don't have the context on what Waybar's module does, or what you potentially want this one to do differently?

madfanat:
Well, shortly, the main feature of both these modules is the display of the current keyboard layout/language (qwerty, dvorak, ENG, RUS, etc whatever you want). Also, it’s nice to have a popup box with these settings.

yavko:
> Well, shortly, the main feature of both these modules is the display of the current keyboard layout/language (qwerty, dvorak, ENG, RUS, etc whatever you want). Also, it’s nice to have a popup box with these settings.

For hyprland I think the layoutchanged event in hyprland-rs can be used, but I'm not totally sure

JakeStanger:
Is there any standardised protocol/service this should be using, or is it down to the compositor to expose over its IPC?

yavko:
I'm pretty sure it's compositor specific

bluebyt:
I tried to do a script for Ironbar to do that, I can change the current layout with this command:
```
hyprctl switchxkblayout logitech-craft next
```
But I cannot get the current layout like explained in this tread: 
https://github.com/hyprwm/Hyprland/discussions/2616

quietvoid:
I'm using a `script` module for this in Sway but it doesn't seem possible for the layout to update automatically without having to set a low interval.

PR #836, opened 2025-01-10
feat: keyboard layout module

kuzy000:
Displays the current keyboard layout. Clicking switches to the next layout. Supports images. See `docs/` for details.

Resolves https://github.com/JakeStanger/ironbar/issues/575.

Current concerns:
1. `keyboard_layout` is a clear name, but maybe it's too verbose?
2. Is it okay to add yet another channel in `hyprland::Client`?
3. Maybe I should rename `layout_map` to `name_map` to be consistent with workspaces module?

### DIFF
--- a/src/clients/compositor/hyprland.rs
+++ b/src/clients/compositor/hyprland.rs
@@ -1,7 +1,11 @@
-use super::{Visibility, Workspace, WorkspaceClient, WorkspaceUpdate};
+use super::{
+    KeyboardLayoutClient, KeyboardLayoutUpdate, Visibility, Workspace, WorkspaceClient,
+    WorkspaceUpdate,
+};
 use crate::{arc_mut, lock, send, spawn_blocking};
 use color_eyre::Result;
-use hyprland::data::{Workspace as HWorkspace, Workspaces};
+use hyprland::ctl::switch_xkb_layout;
+use hyprland::data::{Devices, Workspace as HWorkspace, Workspaces};
 use hyprland::dispatch::{Dispatch, DispatchType, WorkspaceIdentifierWithSpecial};
 use hyprland::event_listener::EventListener;
 use hyprland::prelude::*;
@@ -13,15 +17,21 @@ use tracing::{debug, error, info};
 pub struct Client {
     workspace_tx: Sender<WorkspaceUpdate>,
     _workspace_rx: Receiver<WorkspaceUpdate>,
+
+    keyboard_layout_tx: Sender<KeyboardLayoutUpdate>,
+    _keyboard_layout_rx: Receiver<KeyboardLayoutUpdate>,
 }
 
 impl Client {
     pub(crate) fn new() -> Self {
         let (workspace_tx, workspace_rx) = channel(16);
+        let (keyboard_layout_tx, keyboard_layout_rx) = channel(4);
 
         let instance = Self {
             workspace_tx,
             _workspace_rx: workspace_rx,
+            keyboard_layout_tx,
+            _keyboard_layout_rx: keyboard_layout_rx,
         };
 
         instance.listen_workspace_events();
@@ -32,6 +42,7 @@ impl Client {
         info!("Starting Hyprland event listener");
 
         let tx = self.workspace_tx.clone();
+        let keyboard_layout_tx = self.keyboard_layout_tx.clone();
 
         spawn_blocking(move || {
             let mut event_listener = EventListener::new();
@@ -178,6 +189,9 @@ impl Client {
             }
 
             {
+                let tx = tx.clone();
+                let lock = lock.clone();
+
                 event_listener.add_urgent_state_handler(move |address| {
                     let _lock = lock!(lock);
                     debug!("Received urgent state: {address:?}");
@@ -203,6 +217,55 @@ impl Client {
                             );
                         },
                     );
+                });
+            }
+
+            {
+                let tx = keyboard_layout_tx.clone();
+                let lock = lock.clone();
+
+                event_listener.add_keyboard_layout_change_handler(move |layout_event| {
+                    let _lock = lock!(lock);
+
+                    let layout = if layout_event.layout_name.is_empty() {
+                        // FIXME: This field is empty due to bug in `hyprland-rs_0.4.0-alpha.3`. Which is already fixed in last betas
+
+                        // The layout may be empty due to a bug in `hyprland-rs`, because of which the `layout_event` is incorrect.
+                        //
+                        // Instead of:
+                        // ```
+                        // LayoutEvent {
+                        //     keyboard_name: "keychron-keychron-c2",
+                        //     layout_name: "English (US)",
+                        // }
+                        // ```
+                        //
+                        // We get:
+                        // ```
+                        // LayoutEvent {
+                        //     keyboard_name: "keychron-keychron-c2,English (US)",
+                        //     layout_name: "",
+                        // }
+                        // ```
+                        // 
+                        // Here we are trying to recover `layout_name` from `keyboard_name`
+
+                        let layout = layout_event.keyboard_name.as_str().split(",").nth(1);
+                        let Some(layout) = layout else {
+                            error!(
+                                "Failed to get layout from string: {}. The failed logic is a workaround for a bug in `hyprland 0.4.0-alpha.3`", layout_event.keyboard_name);
+                            return;
+                        };
+
+                        layout.into()
+                    }
+                    else {
+                        layout_event.layout_name
+                    };
+
+                    debug!("Received layout: {layout:?}");
+
+                    send!(tx, KeyboardLayoutUpdate(layout));
                 });
             }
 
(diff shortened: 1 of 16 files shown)


JakeStanger:
Hey, thanks for this! My thoughts below:

1. You've hit the same problem I have with the new `keys` module (which is a name I'm still not happy with). I'm wondering if it's actually a good idea to combine to the two and then it can just be called `keyboard`.
2. Yep, do whatever's necessary. I think that's the best way here.
3. Yes I think so. `name_map` seems like an applicable enough name, and I'd rather the names be consistent than exactly spot on.

src/modules/keys.rs
```diff
@@ -149,22 +187,49 @@ impl Module<gtk::Box> for KeysModule {
...
+            });
+        }
+
+        // Change keyboard layout
+        spawn(async move {
+            trace!("Setting up keyboard_layout UI event handler");
+
+            while let Some(()) = rx.recv().await {
+                if let Err(e) = client.next_keyboard_layout() {
+                    warn!("Couldn't change keyboard layout: {e:#}");
```
JakeStanger, line 226:
```suggestion
                    error!("Couldn't change keyboard layout: {e:#}");
```

src/modules/keys.rs
```diff
@@ -197,31 +266,50 @@ impl Module<gtk::Box> for KeysModule {
             container.add(&*scroll);
         }
 
+        if self.show_layout {
+            layout.add_class("key");
```
JakeStanger, line 270: This class can be removed (see my comment in docs)

JakeStanger (changes requested):
Looking mostly good, just a handful of little bits if you can tweak those please. Thanks for your work on this!

src/clients/compositor/hyprland.rs
```diff
@@ -206,6 +220,32 @@ impl Client {
...
             }
 
+            {
+                let tx = keyboard_layout_tx.clone();
+                let lock = lock.clone();
+
+                event_listener.add_keyboard_layout_change_handler(move |layout_event| {
+                    let _lock = lock!(lock);
+
+                    // FIXME: This field is empty due to bug in `hyprland-rs`. Which is already fixed in last betas
```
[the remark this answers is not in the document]
kuzy000: Both cases are now supported.

The bug is in hyprland-rs crate, not in Hyprland itself. Btw, in the newer version, they decided to rename/break half of the API, so it's not that useful

src/clients/compositor/hyprland.rs
```diff
@@ -300,6 +340,45 @@ impl WorkspaceClient for Client {
...
+                .iter()
+                .find(|k| k.main)
+                .map(|k| k.active_keymap.clone());
+
+            if let Some(layout) = layout {
+                send!(tx, KeyboardLayoutUpdate(layout));
+            } else {
+                error!("Failed to get current keyboard layout hyprland");
+            }
+        }
```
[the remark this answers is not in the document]
kuzy000: I also removed it for `WorkspaceClient`

src/clients/compositor/hyprland.rs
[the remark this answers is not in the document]
JakeStanger: Ah okay, well no harm in supporting both for now I guess. I do want to move away from that crate eventually because it is very slow moving and very unstable.

src/clients/compositor/hyprland.rs
```diff
@@ -206,6 +220,55 @@ impl Client {
...
+                        // Here we are trying to recover `layout_name` from `keyboard_name`
+
+                        let layout = layout_event.keyboard_name.as_str().split(",").nth(1);
+                        if layout.is_none() {
+                            error!(
+                                "Failed to get layout from string: {}. The failed logic is a workaround for a bug in `hyprland 0.4.0-alpha.3`", layout_event.keyboard_name);
+                            return;
+                        }
+
+                        layout.unwrap().into()
```
JakeStanger, line 260: You can use a `let else` binding to avoid the `unwrap`:

```suggestion
                        let Some(layout) = layout else {
                            error!(
                                "Failed to get layout from string: {}. The failed logic is a workaround for a bug in `hyprland 0.4.0-alpha.3`", layout_event.keyboard_name);
                            return;
                        }

                        layout.into()
```

src/clients/compositor/sway.rs
```diff
@@ -133,3 +136,75 @@ impl From<WorkspaceEvent> for WorkspaceUpdate {
         }
     }
 }
+
+impl KeyboardLayoutClient for Client {
+    fn next_keyboard_layout(&self) -> Result<()> {
+        await_sync(async move {
```
JakeStanger, line 142: I'm not sure what my reasoning was for using `await_sync` for the workspaces here, but I'm pretty sure `crate::spawn` would work (and wouldn't block) so would be better. Worth trying...

JakeStanger (changes requested):

JakeStanger:
> I'm not sure I can rename it without breaking something.

RE this, it'll break configs of anybody already using `keys`, but it's not in a stable release so I'm not worried about that. If you change the config enum variant, feature name, and string in the macro at the top of the module, that'll do it. If you'd rather not let me know and I'll pick that bit up.

kuzy000:
Renamed to `keyboard`

src/clients/compositor/sway.rs
kuzy000: Probably to pass an error outside. But it is just logged immediately after the call, so I think it can be logged right in place.

src/clients/compositor/hyprland.rs
kuzy000: Didn't know about that, neat

src/clients/compositor/sway.rs
```diff
@@ -1,18 +1,23 @@
...
-            let mut client = self.connection().lock().await;
-            client.run_command(format!("workspace {id}")).await
-        })?;
-        Ok(())
+    fn focus(&self, id: String) {
+        let client = self.connection().clone();
+        spawn(async move {
+            let mut client = client.lock().await;
+            if let Err(e) = client.run_command(format!("workspace {id}")).await {
+                warn!("Couldn't focus workspace '{id}': {e:#}");
```
JakeStanger, line 18:
```suggestion
                error!("Couldn't focus workspace '{id}': {e:#}");
```

src/clients/compositor/hyprland.rs
```diff
@@ -264,36 +327,73 @@ impl Client {
...
+    fn focus(&self, id: String) {
         let identifier = id.parse::<i32>().map_or_else(
             |_| WorkspaceIdentifierWithSpecial::Name(&id),
             WorkspaceIdentifierWithSpecial::Id,
         );
 
-        Dispatch::call(DispatchType::Workspace(identifier))?;
-        Ok(())
+        if let Err(e) = Dispatch::call(DispatchType::Workspace(identifier)) {
+            warn!("Couldn't focus workspace '{id}': {e:#}");
```
JakeStanger, line 337:
```suggestion
            error!("Couldn't focus workspace '{id}': {e:#}");
```

src/modules/keyboard.rs
```diff
@@ -149,22 +187,47 @@ impl Module<gtk::Box> for KeysModule {
...
+                            let event = KeyEvent {
+                                key,
+                                state: client.get_state(key),
+                            };
+                            module_update!(tx, KeyboardUpdate::Key(event));
                         }
                     }
                     Event::Key(ev) => {
-                        send_async!(tx, ModuleUpdateEvent::Update(ev));
+                        send_async!(tx, ModuleUpdateEvent::Update(KeyboardUpdate::Key(ev)));
```
JakeStanger, line 198:
```suggestion
                        module_update!(tx, KeyboardUpdate::Key(ev));
```

src/modules/keyboard.rs
```diff
@@ -149,22 +187,47 @@ impl Module<gtk::Box> for KeysModule {
...
+            let client = client.clone();
+            let tx = context.tx.clone();
+            spawn(async move {
+                let mut srx = client.subscribe_keyboard_layout_change();
+
+                trace!("Set up keyboard_layout subscription");
+
+                while let Ok(payload) = srx.recv().await {
+                    debug!("Received update: {payload:?}");
+                    send_async!(tx, ModuleUpdateEvent::Update(KeyboardUpdate::Layout(payload)));
```
JakeStanger, line 215:
```suggestion
                    module_update!(tx, KeyboardUpdate::Layout(payload));
```

src/clients/compositor/mod.rs
```diff
@@ -146,10 +171,20 @@ pub enum WorkspaceUpdate {
...
 }
 
 register_fallible_client!(dyn WorkspaceClient, workspaces);
+
+pub trait KeyboardLayoutClient: Debug + Send + Sync {
+    /// Switches to the next layout.
+    fn next_keyboard_layout(&self);
+
+    /// Creates a new to keyboard layout event receiver.
+    fn subscribe_keyboard_layout_change(&self) -> broadcast::Receiver<KeyboardLayoutUpdate>;
```
JakeStanger, line 187: I think these method names can be improved as they're a bit long and partially redundant. `subscribe` matches the other client names. `set_next_active` I think is a better descriptor as `next_keyboard_layout` doesn't necessarily convey it's switching (sounds more like a getter than setter)

```suggestion
    /// Switches to the next layout.
    fn set_next_active(&self);

    /// Creates a new to keyboard layout event receiver.
    fn subscribe(&self) -> broadcast::Receiver<KeyboardLayoutUpdate>;
```

src/modules/workspaces/mod.rs
```diff
@@ -213,9 +213,7 @@ impl Module<gtk::Box> for WorkspacesModule {
             trace!("Setting up UI event handler");
 
             while let Some(name) = rx.recv().await {
-                if let Err(e) = client.focus(name.clone()) {
-                    warn!("Couldn't focus workspace '{name}': {e:#}");
-                };
+                client.focus(name.clone());
```
JakeStanger, line 216: What's the reasoning here for moving the error handing from the module to the clients?

src/clients/compositor/mod.rs
kuzy000: > `subscribe` matches the other client names. 

`WorkspaceClient` has `subscribe_workspace_change`. I changed it too.

Potentially this may cause problems when using `sway::Client` directly (not through `dyn WorkspaceClient` or `dyn KeyboardLayoutClient`), because it now has two `subscribe` methods from different traits.

src/modules/workspaces/mod.rs
kuzy000: I can't return an error from `focus` without either making it `async` or using a `await_sync` inside. The same with `set_next_active`. 

See: https://github.com/JakeStanger/ironbar/pull/836#discussion_r1924342340

src/clients/compositor/mod.rs
```diff
@@ -54,6 +54,26 @@ impl Compositor {
         }
     }
 
+    pub fn create_keyboard_layout(
```
JakeStanger, line 57: Should be renamed to match existing method naming

```suggestion
    pub fn create_keyboard_layout_client(
```

src/modules/keyboard.rs
```diff
@@ -149,22 +187,47 @@ impl Module<gtk::Box> for KeysModule {
...
                     }
                     Event::Key(ev) => {
-                        send_async!(tx, ModuleUpdateEvent::Update(ev));
+                        module_update!(tx, KeyboardUpdate::Key(ev));
                     }
                 }
             }
         });
 
+        let client = context.ironbar.clients.borrow_mut().keyboard_layout()?;
```
JakeStanger, line 204: As this client takes no parameters, you should be able can use the shorthand here:

```suggestion
        let client = context.client::<compositor::KeyboardLayoutClient>()?;
```

(might need a `dyn` on that, or it might not work because it's a trait I'm not 100%)

JakeStanger (changes requested):
Just a couple more very small changes, thanks for being so patient. I should be able to get this merged over the next few days now

src/modules/keyboard.rs
kuzy000: It's `dyn` + `try_client`

JakeStanger (approved):

JakeStanger:
Thanks again!

PR #837, opened 2025-01-11
feat: icon overrides

BowDown097:
This allows the user to override an app's icon in the launcher with a specified icon name. The only alternative to this currently is to dig through your icon theme's folder and replace existing entries / manually add entries yourself, which is rather inconvenient.

I thought about instead adding functionality to look for an associated desktop file and use the icon from that if one is found, but I'm not sure how I would address the problem of app IDs not having to match the desktop file name, or anything in the file.

Example:
```toml
icon_overrides = { legcord = "discord" }
```

<p>&nbsp;&nbsp;&nbsp;&nbsp;Before&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;After</p>


JakeStanger:
Thanks for this. I'm wondering if this would actually be better as a top-level config option, rather than implemented per module. That would allow for a single set of overrides to be defined and take place everywhere an icon is used, rather than eg just within the Launcher module.

BowDown097:
I thought about doing that, but I wasn't sure where else it could be used. Only other place I could think of is the system tray but I'm not sure how that would work.

JakeStanger:
A lot of the modules have an `icons` option, or some other ability to configure its icons. Most of those support image icons, which includes icons straight from the system theme. I was originally thinking they'd make use of it but I guess that'd be useless actually since you're already controlling the displayed icon.

`launcher`, `focused` and `tray` would be the 3 big ones, but yeah you're probably right that it wouldn't be very useful for the tray. 

It would definitely make sense to have in `focused` for the same reason as the launcher, but probably not enough to warrant a top-level option. I think if you're up for it, just duplicating the feature into the focused module would actually be enough.

BowDown097:
:+1: I think it would probably be good to have it for future-proofing purposes too. I'll get to it soon enough.

docs/Configuration guide.md
```diff
@@ -309,6 +309,7 @@ The following table lists each of the bar-level bar config options:
 | `exclusive_zone`  | `boolean`                                      | `true` unless `start_hidden` is enabled. | Whether the bar should reserve an exclusive zone around it.                                                                |
 | `popup_gap`       | `integer`                                      | `5`                                      | The gap between the bar and popup window.                                                                                  |
 | `icon_theme`      | `string`                                       | `null`                                   | Name of the GTK icon theme to use. Leave blank to use default.                                                             |
+| `icon_overrides`  | `map<string, string>`                          | `null`                                   | Map of app IDs (or classes) to icon names, overriding the app's default icon.                                              |
```
JakeStanger, line 312:
```suggestion
| `icon_overrides`  | `Map<string, string>`                          | `null`                                   | Map of app IDs (or classes) to icon names, overriding the app's default icon.                                              |
```

src/config/mod.rs
```diff
@@ -290,6 +290,12 @@ pub struct BarConfig {
     /// **Default**: `null`
     pub icon_theme: Option<String>,
 
+    /// Map of app IDs (or classes) to icon names,
+    /// overriding the app's default icon.
+    ///
+    /// **Default**; `null`
+    pub icon_overrides: Option<HashMap<String, String>>,
```
JakeStanger, line 297: This does not need to be an `Option`. Empty hashmaps do not allocate, so you can stick a `#[serde(default)]` attr on the field and simplify the code throughout

src/modules/launcher/mod.rs
```diff
@@ -149,24 +149,33 @@ impl Module<gtk::Box> for LauncherModule {
                 favorites
                     .iter()
                     .map(|app_id| {
+                        let icon_override = _info
```
JakeStanger, line 152: Since `info` is being used now, it should be renamed to drop the `_` prefix

JakeStanger (changes requested):

src/config/mod.rs
BowDown097: Oh, I thought it may have been some kind of stylistic choice since Option is used for HashMap in every other place in the config. Does definitely make life easier.

JakeStanger: So it is, not sure why I did that. With both the Maps and Vecs it's unnecessary; I'll tidy that up soon.

src/config/mod.rs
```diff
@@ -290,6 +290,13 @@ pub struct BarConfig {
     /// **Default**: `null`
     pub icon_theme: Option<String>,
 
+    /// Map of app IDs (or classes) to icon names,
+    /// overriding the app's default icon.
+    ///
+    /// **Default**; `{}`
```
JakeStanger, line 296:
```suggestion
    /// **Default**: `{}`
```

src/config/mod.rs
```diff
@@ -290,6 +290,13 @@ pub struct BarConfig {
...
+    #[serde(default)]
+    pub icon_overrides: HashMap<String, String>,
```
JakeStanger, line 298: This gives the ability to set overrides per-bar. I'm not sure I can see a use-case for that personally - it would probably be better to have this on the `Config` struct so that it applies application-wide. 

Happy to keep as-is if you think otherwise though?

src/modules/focused.rs
```diff
@@ -153,9 +154,15 @@ impl Module<gtk::Box> for FocusedModule {
 
         {
             let icon_theme = icon_theme.clone();
+            let icon_overrides = icon_overrides.clone();
```
JakeStanger, line 157: If this were to move to `Config` (see other comment)...I don't particularly like that we need to clone here, as it'll re-alloc for every focused/launcher instance. It'd be cheaper to wrap in an `Arc` (no `Mutex` needed since we're not mutating).

If it's kept at bar-level config though, I think this is fine since in most cases you'll only have 1 copy of the launcher or focused module at a time anyway.

src/modules/launcher/mod.rs
```diff
@@ -149,24 +149,32 @@ impl Module<gtk::Box> for LauncherModule {
...
                         )
                     })
                     .collect::<IndexMap<_, _>>()
             });
 
         let items = arc_mut!(items);
-
         let items2 = Arc::clone(&items);
 
+        let icon_overrides = arc_mut!(info.icon_overrides.clone());
```
JakeStanger, line 168: This does not need to be mutable. Removing the mutex will also simplify some of the code below as it removes the need to lock.

JakeStanger (changes requested):

src/modules/focused.rs
BowDown097: I'm not sure how icon_overrides could be used in the closure without performing a clone (letalone in an Arc). Wouldn't there be lifetime issues with the ModuleInfo reference, or with a borrow of context.ironbar.config outside the closure?

JakeStanger: You'd still need to clone, but if it were in an `Arc`, the clone would just increase a reference count rather than re-alloc so is much cheaper.

JakeStanger: I meant put the `Arc` on the `ModuleInfo` struct when it's first created, instead of the reference. You can then freely clone the `Arc`. The current implementation still fully clones the data, then puts the cloned result in an Arc that's used once.

src/bar.rs
```diff
@@ -268,6 +269,7 @@ impl Bar {
                     output_name: &self.monitor_name,
                     location: $location,
                     icon_theme: &icon_theme,
+                    icon_overrides: Arc::new(self.ironbar.config.borrow().icon_overrides.clone()),
```
JakeStanger, line 272: This still clones the data once per bar. Create the `Arc` at the top level, then pass it down into the bar creation func(s). The num of arguments is stacking up but I'll look to refactor that later.

JakeStanger: I guess you could do it on the struct itself if that's easier, but I meant just move it out of the config, wrap it in `Arc` and pass that to `create_bar` and `Bar::new`. So long as there's only one copy of the data...

JakeStanger (approved):
Cheers, thanks for your patience on this one.

PR #838, opened 2025-01-13
feat: bluetooth module

kuzy000:
Bluetooth module via `bluer` crate. I think it is complete now.

ironbar_bluetooth

### DIFF
--- a/docs/modules/Bluetooth.md
+++ b/docs/modules/Bluetooth.md
@@ -1,0 +1,242 @@
+> [!NOTE]
+> For the battery percentage option the experimental features of BlueZ needs to be enabled by adding `Experimental = true` to the `[General]` section in `/etc/bluetooth/main.conf`
+
+Displays the current bluetooth status.
+Clicking on the widget opens a popout displaying list of available devices and connection controls.
+
+![Screenshot of bluetooth widget](https://f.jstanger.dev/github/ironbar/bluetooth.png)
+
+## Configuration
+
+> Type: `bluetooth`
+
+| Name                          | Type      | Default                                               | Description                                                                                                                                      |
+| ----------------------------- | --------- | ----------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `format.not_found`            | `string`  | ``                                                    | Format string to use for the widget button when bluetooth adapter not found.                                                                     |
+| `format.disabled`             | `string`  | ` Off`                                               | Format string to use for the widget button when bluetooth adapter is disabled.                                                                   |
+| `format.enabled`              | `string`  | ` On`                                                | Format string to use for the widget button when bluetooth adapter is enabled but no devices are connected.                                       |
+| `format.connected`            | `string`  | ` {device_alias}`                                    | Format string to use for the widget button when bluetooth adapter is enabled and a device is connected.                                          |
+| `format.connected_battery`    | `string`  | ` {device_alias} • {device_battery_percent}%`        | Format string to use for the widget button when bluetooth adapter is enabled, a device is connected and `{device_battery_percent}` is available. |
+| `popup.scrollable`            | `boolean` | `true`                                                | If true makes the popup scrollable, if false stretchable to show all of its content.                                                             |
+| `popup.header`                | `string`  | ` Enable Bluetooth`                                  | Format string to use for the header of popup window.                                                                                             |
+| `popup.disabled`              | `string`  | `{adapter_status}`                                    | Format string to use for the message that is displayed when the adapter is not found or disabled.                                                |
+| `popup.device.header`         | `string`  | `{device_alias}`                                      | Format string to use for the header of device box.                                                                                               |
+| `popup.device.header_battery` | `string`  | `{device_alias}`                                      | Format string to use for the header of device box when `{device_battery_percent}` is available.                                                  |
+| `popup.device.footer`         | `string`  | `{device_status}`                                     | Format string to use for the footer of device box.                                                                                               |
+| `popup.device.footer_battery` | `string`  | `{device_status} • Battery {device_battery_percent}%` | Format string to use for the footer of device box when `{device_battery_percent}` is available.                                                  |
+| `adapter_status.not_found`    | `string`  | `No Bluetooth adapters found`                         | The value of `{adapter_status}` formatting token when adapter not found.                                                                         |
+| `adapter_status.enabled`      | `string`  | `Bluetooth enabled`                                   | The value of `{adapter_status}` formatting token when adapter is enabled.                                                                        |
+| `adapter_status.enabling`     | `string`  | `Enabling Bluetooth...`                               | The value of `{adapter_status}` formatting token when adapter is enabling.                                                                       |
+| `adapter_status.disabled`     | `string`  | `Bluetooth disabled`                                  | The value of `{adapter_status}` formatting token when adapter is disabled.                                                                       |
+| `adapter_status.disabling`    | `string`  | `Disabling Bluetooth...`                              | The value of `{adapter_status}` formatting token when adapter is disabling.                                                                      |
+| `device_status.connected`     | `string`  | `Connected`                                           | The value of `{device_status}` formatting token when device is connected.                                                                        |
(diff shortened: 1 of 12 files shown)


JakeStanger:
Looks amazing so far, thanks so much for this. Looking forward to the final thing.

docs/modules/Bluetooth.md
```diff
@@ -0,0 +1,245 @@
+> [!NOTE]
+> For the battery percentage option the experimental features of BlueZ needs to be enabled by adding `Experimental = true` to the `[General]` section in `/etc/bluetooth/main.conf`
+
+Displays the current bluetooth status.
+Clicking on the widget opens a popout displaying list of available devices and connection controls.
+
+![Screenshot of bluetooth widget](TODO)
```
kuzy000, line 7: @JakeStanger not sure if I can use this link: https://github.com/user-attachments/assets/9650b2ff-c536-49d0-947e-92ebe147ebbd ? I think it's better to upload it somewhere else

JakeStanger: Yeah I don't know how long the GitHub CDN keeps files around and I don't fully trust it. I've uploaded it to my own fileshare alongside some of the others, so you can use this link:

https://f.jstanger.dev/github/ironbar/bluetooth.png

src/modules/bluetooth.rs
```diff
@@ -0,0 +1,838 @@
...
+            let mut device_map = HashMap::new();
+            let mut seq = 0u32;
+            let mut num_pinned = 0;
+
+            let mut handle_state = move |state: BluetoothState| {
+                if let Some(handle) = enable_handle.take() {
+                    header_switch.disconnect(handle);
+                }
+
+                devices.set_visible(matches!(state, BluetoothState::Enabled { .. }));
```
JakeStanger, line 716: Rather than relying on the `matches!` macro throughout, I think it'd be better to implement `PartialEq` (and I think `Eq`?) and then just use `==` or `!=`. 

For `Enabled` specifically you could abstract to a `is_enabled` method. I think you might need the macro there still since it's a struct variant.

src/modules/bluetooth.rs
```diff
@@ -0,0 +1,838 @@
...
+                    }));
+                }
+
+                if matches!(state, BluetoothState::Enabled { .. }) {
+                    let tx = tx.clone();
+                    enable_handle = Some(header_switch.connect_active_notify(move |switch| {
+                        try_send!(tx, BluetoothAction::Disable);
+                        switch.set_sensitive(false);
+                    }));
+                }
```
JakeStanger, line 768: Can these two be swapped so that enabled comes first please? You could also use an `else if` here.

src/modules/bluetooth.rs
```diff
@@ -0,0 +1,838 @@
...
+
+                if let BluetoothState::Enabled { devices } = state {
```
JakeStanger, line 770: Does this block need to be separate, or could it be moved into the `if state == enabled` above?

src/modules/bluetooth.rs
JakeStanger, line 1: This file is getting a bit long for my liking. Could you promote to a module (like music/launcher/etc) and split into two or more files as you see fit please? I'd be inclined to move the config and formatting stuff out.

JakeStanger (changes requested):
By far and away the highest quality and most feature complete PR I've received, so huge thank you for your work on this. The code seems sensible and logically sound to me, so no complaints there. 

I've just raised a few minor bits where I need some clarification or would like it tweaked, if you're able to have a look at those please.

src/clients/bluetooth.rs
```diff
@@ -0,0 +1,417 @@
...
+                                Self::patch_new_state(&mut new_state, old_state);
+                                *old_state = new_state;
+                            });
+                        }
+                        Err(err) => {
+                            error!("Bluetooth error: {}", err);
+                        }
+                    }
+
+                    tokio::time::sleep(Duration::from_secs(5)).await;
```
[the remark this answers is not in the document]
kuzy000: This is not the only place where the state is updated. The state is also updated after each response from `bluer`, for example, after `device.connect().await`. So this delay only affects the appearance of newly available devices, but not the actions of the user.

It is also needed to mitigate strange errors from the `bluer`. When, for example, it returns an error when connecting a device. But in fact, the device connects.

Ideally, this `loop` should be replaced by a subscription to `bluer` events. Unfortunately it is very unreliable and misses notifications for half of the events.

docs/modules/Bluetooth.md
```diff
@@ -0,0 +1,242 @@
...
+| `format.enabled`              | `string`  | ` On`                                                | Format string to use for the widget button when bluetooth adapter is enabled but no devices are connected.                                       |
+| `format.connected`            | `string`  | ` {device_alias}`                                    | Format string to use for the widget button when bluetooth adapter is enabled and a device is connected.                                          |
+| `format.connected_battery`    | `string`  | ` {device_alias} • {device_battery_percent}%`        | Format string to use for the widget button when bluetooth adapter is enabled, a device is connected and `{device_battery_percent}` is available. |
+| `popup.scrollable`            | `boolean` | `true`                                                | Whether to make the popup scrollable or stretchable to show all of its content.                                                                  |
```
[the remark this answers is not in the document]
kuzy000: Fixed

docs/modules/Bluetooth.md
```diff
@@ -0,0 +1,242 @@
...
+| `popup.header`                | `string`  | ` Enable Bluetooth`                                  | Format string to use for the header of popup window.                                                                                             |
+| `popup.status`                | `string`  | `{adapter_status}`                                    | Format string to use for the status string that is displayed when the adapter is not found or disabled.                                          |
```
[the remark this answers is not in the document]
kuzy000: Fixed

src/modules/bluetooth.rs
kuzy000: All `matches!` removed

src/modules/bluetooth.rs
kuzy000: Fixed

src/modules/bluetooth.rs
kuzy000: Fixed

src/modules/bluetooth.rs
kuzy000: Fixed

kuzy000:
Actual diff: https://github.com/JakeStanger/ironbar/compare/a23ae89b3d59c361e198e11f58ce3b4267c447ff..fe0f07b19a605b1d008248e8b21afe2068091f48

The rest is formatting and merging

P.S. Maybe i should start to push individual commits with fixes as GitHub PRs aren't very rebase-friendly.

docs/modules/Bluetooth.md
JakeStanger, line 1: Just realised you're missing an entry in `_sidebar.md` :)

src/clients/bluetooth.rs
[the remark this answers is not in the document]
JakeStanger: Hm okay, a little dodgy that the official bindings crate has that sort of issue but it doesn't surprise me to be honest. Happy enough with that for now.

Could you just stick a summary of the above in a comment please so that's clear in code.

docs/modules/Bluetooth.md
kuzy000: Added

src/clients/bluetooth.rs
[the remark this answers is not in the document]
kuzy000: Added a comment

JakeStanger:
Literally at the same pub again, 60 miles from home. Weird coincidence.

I'll review in the next couple of days, but reckon we're about there. Cheers

JakeStanger:
Build only failing due to clippy errors on master, ignoring

kuzy000:
Thank you for merge!

> It does have a couple of rough edges I've managed to hit in some edge cases, but it's hard to say whether that's just Bluetooth being Bluetooth. I don't really have a good test setup for it, so I'm happy to release this as-is and we'll refine any issues as they come in.

I have been using it for the last 3 months. It works... okay for me. 

I mean, it has roughly the same issues as `bluetoothctl`. Maybe it's because of the cheap usb bluetooth adapter I'm using. But on the other hand it works well on dual-booted windows. So, I think it's a linux bluetooth stack thing.

PR #846, opened 2025-01-16
refactor(keys): switch to `colpetto`

JakeStanger:
Switches from the `input` crate to new async `colpetto`. Less janky to work with, and comes with noticeable performance improvements.


JakeStanger:
@morr0ne :eyes:

PR #855, opened 2025-01-31
feat: rename sway_mode to bindmode and add Hyprland support

Rodrigodd:
I am currently migrating from sway to Hyprland, and so I added support for Hyprland in the sway_mode module. Because the module is no longer sway-only, I decided to rename to a more generic name. This feature does not have a consistent name across window managers, i3/Sway [calls it binding mode](https://i3wm.org/docs/userguide.html#binding_modes), while Hyprland calls it [Keybind submaps](https://wiki.hyprland.org/Configuring/Binds/#submaps). But "bindmode" appears to be a good enough name.

I made "sway_mode" be a alias for "bindmode" to avoid breaking configs that use the previous name, but I can remove that if you prefer.

### DIFF
--- a/docs/modules/Bindmode.md
+++ b/docs/modules/Bindmode.md
@@ -1,12 +1,12 @@
-Displays the current sway mode in a label. If the current sway mode is
-"default", nothing is displayed.
+> [!IMPORTANT]
+> This module is currently only available on Sway and Hyprland.
 
-> [!NOTE]
-> This module only works under the [Sway](https://swaywm.org/) compositor.
+Displays Sway's current binding mode or [Hyprland's current submap](https://wiki.hyprland.org/Configuring/Binds/#submaps)
+in a label. Nothing is displayed if no binding mode is active.
 
 ## Configuration
 
-> Type: `sway-mode`
+> Type: `bindmode`
 
 | Name                  | Type                                        | Default | Description                                                                                                                                           |
 | --------------------- | ------------------------------------------- | ------- | ----------------------------------------------------------------------------------------------------------------------------------------------------- |
@@ -22,7 +22,7 @@ Displays the current sway mode in a label. If the current sway mode is
 {
   "end": [
     {
-      "type": "sway-mode",
+      "type": "bindmode",
       "truncate": "start"
     }
   ]
@@ -36,7 +36,7 @@ Displays the current sway mode in a label. If the current sway mode is
 
 ```toml
 [[end]]
-type = "sway-mode"
+type = "bindmode"
 truncate = "start"
 ```
 
@@ -47,7 +47,7 @@ truncate = "start"
 
 ```yaml
 end:
-  - type: "sway-mode"
+  - type: "bindmode"
     truncate: "start"
 ```
 
@@ -60,7 +60,7 @@ end:
 {
   end = [
     {
-      type = "sway-mode"
+      type = "bindmode"
       truncate = "start"
     }
   ]
@@ -71,8 +71,8 @@ end:
 
 ## Styling
 
-| Selector     | Description            |
-| ------------ | ---------------------- |
-| `.sway_mode` | Sway mode label widget |
+| Selector    | Description            |
+| ----------- | ---------------------- |
+| `.bindmode` | Bind mode label widget |
 
 For more information on styling, please see the [styling guide](styling-guide).
--- a/src/clients/compositor/hyprland.rs
+++ b/src/clients/compositor/hyprland.rs
@@ -10,10 +10,21 @@ use hyprland::shared::{HyprDataVec, WorkspaceType};
 use tokio::sync::broadcast::{Receiver, Sender, channel};
 use tracing::{debug, error, info};
 
+#[cfg(feature = "bindmode+hyprland")]
+use super::{BindModeClient, BindModeUpdate};
+#[cfg(feature = "keyboard+hyprland")]
+use super::{KeyboardLayoutClient, KeyboardLayoutUpdate};
+
 #[derive(Debug)]
 struct TxRx<T> {
     tx: Sender<T>,
     _rx: Receiver<T>,
+}
+impl<T: Clone> TxRx<T> {
+    fn new() -> Self {
+        let (tx, rx) = channel(16);
+        Self { tx, _rx: rx }
+    }
 }
 
 #[derive(Debug)]
@@ -23,27 +34,20 @@ pub struct Client {
 
     #[cfg(feature = "keyboard+hyprland")]
     keyboard_layout: TxRx<KeyboardLayoutUpdate>,
+
+    #[cfg(feature = "bindmode+hyprland")]
+    bindmode: TxRx<BindModeUpdate>,
 }
 
 impl Client {
     pub(crate) fn new() -> Self {
-        #[cfg(feature = "workspaces+hyprland")]
-        let (workspace_tx, workspace_rx) = channel(16);
-
-        #[cfg(feature = "keyboard+hyprland")]
-        let (keyboard_layout_tx, keyboard_layout_rx) = channel(16);
-
         let instance = Self {
             #[cfg(feature = "workspaces+hyprland")]
-            workspace: TxRx {
-                tx: workspace_tx,
-                _rx: workspace_rx,
-            },
+            workspace: TxRx::new(),
             #[cfg(feature = "keyboard+hyprland")]
-            keyboard_layout: TxRx {
-                tx: keyboard_layout_tx,
-                _rx: keyboard_layout_rx,
-            },
+            keyboard_layout: TxRx::new(),
+            #[cfg(feature = "bindmode+hyprland")]
+            bindmode: TxRx::new(),
         };
 
         instance.listen_events();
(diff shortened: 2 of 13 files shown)


JakeStanger:
Thanks for this. 

There may be some conflicts with #836, which I'm looking to merge soon, which will relate to the client code. I'd also like to ensure that there's consistency between this, the workspaces and keyboard module so I will hold off a full review until that's merged.

JakeStanger:
#836 is now merged & has created the expected conflict

Rodrigodd:
@JakeStanger I rebased the PR over master.

My code has a very different style compared to the workspace and keyboard layout mode. I just query both compositors on module creation, choosing the first one that does not error out, while the other modules create a trait over the Compositors.  

I found the second approach too complex to be worth it in this case. I started implementing it anyway for consistency's sake, but I noticed that the `feature`s for the keyboard module are slightly broken (matches become non-exhaustive if you disable one of the features, and part of the code depends on "workspace," but it is not listed as a dependency).  

So for now, I will wait for your opinion.

JakeStanger:
> I just query both compositors on module creation, choosing the first one that does not error out, while the other modules create a trait over the Compositors.

The issue with your approach is that the module code is now coupled to, and responsible for, compositor-specific logic. This makes future changes more difficult, and could lead to the code getting more unwieldy if say support for another compositor was added. Eventually there may also be a Wayland protocol so that the bar does not need to rely on compositor-specific IPCs.

Ideally, the modules should have *no idea* what compositor or protocol or whatever the client is talking to. They should just say "get me the client for X", get the instance, and be able the blindly call the methods on the interface.

The other issue I have is that it duplicates some of the logic - there is already [a function](https://github.com/JakeStanger/ironbar/blob/68ccc7ac4d1ac3f5c70891973138acce5fb2a851/src/clients/compositor/mod.rs#L41) for getting the current compositor (which isn't exception based so is preferable).

tl;dr going the trait route avoids duplication and makes it easier to make changes in future.

As for the feature flags, that's an issue I missed. The two should be decoupled from each other, so I'll have a look at that.

Rodrigodd:
I rebased over master and refactored the implementation to follows the ones for `workspace` and `keyboard` modules. I think it is ready for a review.

docs/modules/Sway-mode.md
JakeStanger, line 1: I'd rather make the breaking change and remove the old module, rather than keeping it around for backwards compatibility. Can you remove references to `sway-mode` please?

docs/modules/Bindmode.md
```diff
@@ -0,0 +1,75 @@
+Displays Sway's current binding mode or [Hyprland's current submap](https://wiki.hyprland.org/Configuring/Binds/#submaps)
```
JakeStanger, line 1: Please add a header stating compatibility limitations:

```suggestion
> [!IMPORTANT]
> This module is currently only supported on Sway, Hyprland and Niri*

Displays Sway's current binding mode or [Hyprland's current submap](https://wiki.hyprland.org/Configuring/Binds/#submaps)
```

docs/modules/Bindmode.md
```diff
@@ -0,0 +1,75 @@
...
+in a label. If the current sway mode is "default", nothing is displayed.
```
JakeStanger, line 2: This sentence probably needs updating to be inclusive of Hyprland behaviour, if it applies here.

src/clients/compositor/hyprland.rs
```diff
@@ -23,42 +34,41 @@ pub struct Client {
...
+        let workspace_tx = self.workspace.tx.clone();
 
         #[cfg(feature = "keyboard+hyprland")]
         let keyboard_layout_tx = self.keyboard_layout.tx.clone();
 
+        #[cfg(feature = "bindmode+hyprland")]
+        let bindmode_tx = self.bindmode.tx.clone();
+
+        #[cfg(feature = "workspaces+hyprland")]
+        let active = Self::get_active_workspace()?;
```
JakeStanger, line 70: What's the reason for this addition and the `listen_workspace_events` method change? 

If that's a separate fix, can it be in a separate commit please.

src/clients/compositor/mod.rs
```diff
@@ -66,6 +66,32 @@ impl Compositor {
...
+    pub fn create_bindmode_client(
+        clients: &mut super::Clients,
+    ) -> ClientResult<dyn BindModeClient + Send + Sync> {
+        let current = Self::get_current();
+        debug!("Getting keyboard_layout client for: {current}");
+        match current {
+            #[cfg(feature = "bindmode+sway")]
+            Self::Sway => clients
+                .sway()
+                .map(|client| client as Arc<dyn BindModeClient + Send + Sync>),
```
JakeStanger, line 79: Is this cast required just on the first branch to make the `match` happy?

src/clients/compositor/mod.rs
```diff
@@ -66,6 +66,32 @@ impl Compositor {
...
+            #[cfg(feature = "bindmode+hyprland")]
+            Self::Hyprland => Ok(clients.hyprland()?),
+            #[cfg(feature = "niri")]
+            Self::Niri => Err(Report::msg("Unsupported compositor").note(
+                "Currently keyboard layout functionality are only supported by Sway and Hyprland",
```
JakeStanger, line 84: references to `keyboard` need changing to `bindmode` here and 2 cases below

src/config/mod.rs
```diff
@@ -83,8 +83,10 @@ pub enum ModuleConfig {
     Script(Box<ScriptModule>),
     #[cfg(feature = "sys_info")]
     SysInfo(Box<SysInfoModule>),
-    #[cfg(feature = "sway")]
-    SwayMode(Box<SwayModeModule>),
+    #[cfg(feature = "bindmode")]
+    SwayMode(Box<Bindmode>),
+    #[cfg(feature = "bindmode")]
+    Bindmode(Box<Bindmode>),
```
JakeStanger, line 89: Please keep these in alphabetical order. Same below

src/modules/bindmode.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    fn spawn_controller(
+        &self,
+        _info: &ModuleInfo,
+        context: &WidgetContext<Self::SendMessage, Self::ReceiveMessage>,
+        _rx: mpsc::Receiver<Self::ReceiveMessage>,
+    ) -> Result<()> {
+        info!("Bindmode module started");
+
+        let client = context.ironbar.clients.borrow_mut().bindmode()?;
```
JakeStanger, line 41:
```suggestion
        let client = let client = context.client::<clients::compositor::BindModeClient>();
```

src/modules/bindmode.rs
```diff
@@ -0,0 +1,90 @@
...
+            let label = label.clone();
+
+            if let Some(truncate) = self.truncate {
+                label.truncate(truncate);
+            }
+
+            let on_mode = move |mode: BindModeUpdate| {
+                trace!("mode: {:?}", mode);
+                label.set_use_markup(mode.pango_markup);
+                if mode.name == "default" {
```
JakeStanger, line 75: As with comment in docs, need to ensure this covers Hyprland behaviour too

JakeStanger (changes requested):
I'm happy with the implementation now, so thanks for taking the time to change that. Most of what I've raised below is the standard small changes stuff. 

The only biggish change is removing the sway-mode module as an alias etc throughout please. I think it's best to make the breaking change now, and keep things tidy to avoid any confusion or maintenance overhead.

There are also a few comments from the previous review, if you can address those please.

src/clients/compositor/hyprland.rs
Rodrigodd: It comes from the previous implementation, where I try to create a Hyprland client and if it fails, I try to create a sway client. So I had to make the Hyprland client return a `Result`.

I split that in a separated commit. This change does not make difference in practice because now the Hyprland client is only creating after checking that Hyprland is running.

src/clients/compositor/mod.rs
Rodrigodd: No, the implicit cast don't work there because of it being wrapped in a `Result` I believed, so it is being explicitly casted. But I rewrite that as `Ok(clients.sway()?)`, and the implicit cast works there. It also became more consistent with the Hyprland branch.

I kept it as a separated commit because I also change the other two compositor modules.

src/modules/bindmode.rs
Rodrigodd: This is only need for sway, I moved the check to the `impl` in the sway compositor module and added a comment.

src/modules/bindmode.rs
```diff
@@ -1,73 +1,64 @@
...
-        info: &ModuleInfo,
+        _info: &ModuleInfo,
     ) -> Result<ModuleParts<Label>> {
-        let label = Label::builder()
-            .use_markup(true)
-            .angle(self.layout.angle(info))
-            .justify(self.layout.justify.into())
-            .build();
+        let label = Label::new(None);
+        label.set_use_markup(true);
```
JakeStanger, line 61: This needs reverting to its original impl:

```suggestion
        let label = Label::builder()
            .use_markup(true)
            .angle(self.layout.angle(info))
            .justify(self.layout.justify.into())
            .build();
```

src/clients/compositor/hyprland.rs
JakeStanger: If it's not needed anymore then can it be removed then please? I'd rather not add more parameters to method signatures if not necessary.

JakeStanger (changes requested):

src/clients/compositor/hyprland.rs
Rodrigodd: The point was to avoid panics in the Hyprland client, event if they are currently unreachable. But no problem. I reverted that change.

JakeStanger (approved):
LGTM, thanks for all your hard work on this, I appreciate there was a fair amount of back and forth.

src/clients/compositor/hyprland.rs
JakeStanger: I'd argue that's more of an architecture problem, rather than something to monkey-patch over. I'd be happy to review that in more detail outside this PR, but for now yeah let's keep it out and keep this focused.

PR #863, opened 2025-02-06
Fixes clipboard deadlocking data sources

ToxicMushroom:
Wayland spec says clients should read the offers they request receive on.
Ironbar was failing to do this because the event loop it used could get deadlocked on reading a pipe.

This pr fixes that by switching to tokio's async readers 

Thanks @zeroeightysix

### DIFF
--- a/.idea/runConfigurations/Run__Live_Config_.xml
+++ b/.idea/runConfigurations/Run__Live_Config_.xml
@@ -11,7 +11,7 @@
     <option name="backtrace" value="SHORT" />
     <envs>
       <env name="PATH" value="/usr/local/bin:/usr/bin:$USER_HOME$/.local/share/npm/bin" />
-      <env name="RUST_LOG" value="debug" />
+      <env name="IRONBAR_LOG" value="debug" />
     </envs>
     <option name="isRedirectInput" value="false" />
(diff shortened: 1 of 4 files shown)


JakeStanger (approved):
Looks great, thanks so much both of you. I could never get anywhere trying to get it working with Tokio but who knows what I was doing wrong.

I wonder if this means the horrendous `epoll` and pipe size-increasing syscalls can be removed now...

JakeStanger:
Are you able to squash your commits down with the following message please:

```
fix(clipboard): possible deadlock when copying (#863)
```

I think if I squash from browser it'll lose the co-authoring but want to try and keep you both credited :)

zeroeightysix:
> I think if I squash from browser it'll lose the co-authoring but want to try and keep you both credited :)

I think all you need is to keep the `Co-authored-by: Ridan Vandenbergh <ridanvandenbergh@gmail.com>` at the bottom of the commit message (2 blank lines inbetween)

JakeStanger:
Let's see...

JakeStanger:
Cool okay yeah, it just previews it wrong. Thanks again guys!

ToxicMushroom:
you were faster :p cheers

PR #869, opened 2025-02-16
keyboard: fix panic when layout update channel lags behind

quietvoid:
Not sure of the implications of increasing capacity but I don't think it's really a problem.
Switched event loop to avoid ending on `RecvError::Lagged` response.

### DIFF
--- a/src/clients/compositor/hyprland.rs
+++ b/src/clients/compositor/hyprland.rs
@@ -25,7 +25,7 @@ pub struct Client {
 impl Client {
     pub(crate) fn new() -> Self {
         let (workspace_tx, workspace_rx) = channel(16);
-        let (keyboard_layout_tx, keyboard_layout_rx) = channel(4);
+        let (keyboard_layout_tx, keyboard_layout_rx) = channel(16);
 
         let instance = Self {
             workspace_tx,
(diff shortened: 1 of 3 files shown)


quietvoid:
FWIW on my system with the original channel capacity, I'd see these:
> Channel lagged behind by 4, this may result in unexpected or broken behaviour

Up to `lagged behind by 7`.

JakeStanger:
Cheers.

> Not sure of the implications of increasing capacity but I don't think it's really a problem.

Memory is allocated for `capacity * size(T)`, otherwise nothing.

> Up to lagged behind by 7.

This suggests either capacity is still too low, or simply messages are being processed faster than they're being consumed. If the latter is the case, increased capacity will likely only offset the problem and we'll need to look into that to properly fix.

I'm going to hold off merging this to see if we can get a better answer in the next couple of days. If not, I'll merge as-is as it's an improvement over crashing.

quietvoid:
> This suggests either capacity is still too low

The messages only show up with the old capacity of 4. With 16 there's no warnings.

JakeStanger:
Ah I see! In that case, happy to roll with this

JakeStanger (approved):

PR #872, opened 2025-02-21
refactor: update to rust 2024, fix strict clippy warnings

JakeStanger:
Happy new year


JakeStanger:
Faster than the CI image whoops

PR #890, opened 2025-03-10
Fixed volume module inability to rotate based on bar location

skelitheprogrammer:
Made a discussion #887 which show this problem.

This change based on [clock module](https://github.com/JakeStanger/ironbar/blob/master/src/modules/clock.rs#L132)

Tested horizontal and vertical, all good.
There is still a problem with the behavior of the text and the icon. Currently the icon is not trying to center itself. I can try to fix this in the this PR, but this is my first time using rust so don't at me xD.


JakeStanger:
Thanks for the fix. Unfortunately for you I was already working on #907 in the background which superseeds this, so going to close. Appreciated nonetheless

PR #911, opened 2025-03-24
refactor: sort niri workspaces before WorkspaceUpdate::Init

MapoMagpie:
Niri has a "[Named Workspaces](https://github.com/YaLTeR/niri/wiki/Configuration:-Named-Workspaces)" feature, 
where Workspace's id are assigned based on the order declared in the configuration file. 

However, when Ironbar initializes, the WorkspacesChanged event output by Niri is not sorted by ID. As a result, even when I declare `sort = "added"` in Ironbar's configuration file, the final sorting remains inconsistent. 

To achieve the expected behavior, we simply need to sort new_workspaces by ID before initializing workspaces in Ironbar. This change will not affect the original functionality.

This is my ironbar workspaces configuration
```
[[start]]
type = "workspaces"
sort = "added"
[start.name_map]
NETW = " NETW  "
TERM = " TERM  "
CODE = " CODE  "
CHAT = " CHAT  "
GAME = " GAME  "
6 = " TEMP  "
```
Before change
grim_20250324_210020
After change
grim_20250324_210036

### DIFF
--- a/src/clients/compositor/niri/mod.rs
+++ b/src/clients/compositor/niri/mod.rs
@@ -40,7 +40,9 @@ impl Client {
                         let mut updates: Vec<WorkspaceUpdate> = vec![];
 
                         if first_event {
-                            updates.push(WorkspaceUpdate::Init(new_workspaces.clone()));
+                            let mut new_workspaces = new_workspaces.clone();
+                            new_workspaces.sort_by_key(|w| w.id);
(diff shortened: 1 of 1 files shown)


JakeStanger:
Thanks for the fix. Would you mind just adding a comment to the code summarising what you've written above? It's not obvious why the sort is there otherwise

JakeStanger (approved):
Thanks

PR #112, opened 2023-04-21
The long road to GTK4

JakeStanger:
GTK4 introduces a lot of breaking changes, mostly involving event handling. This PR will track the progress of the port. It's going to take me a while to get through it all, and this will be secondary to continuing bar development.

You need gtk4 and gtk4-layer-shell installed. These are normally available through your package mananger.

Progress documented here: https://github.com/JakeStanger/ironbar/wiki/gtk4

Resolves #933 
Resolves #244
Resolves #900
Resolves #1091
Resolves #963 
Resolves #1042 
Resolves #1183 

Closes #791 

---

Huge shoutout to those that helped with the port:
- Thanks to @tglman for porting the tray, upower, volume and network manager modules.
- Thanks to @pinklifeart for porting the notifications module.

PR #913, opened 2025-03-27
fix(notifications): gtk4 port

pinklifeart:
Replaced the overlay_pass_through call following the migration notes doc: https://docs.gtk.org/gtk4/migrating-3to4.html#adapt-to-changes-in-gtkoverlay-api

Now it successfully compiles and on the first glance seems to be behaving the same as the gtk3 version, so hopefully a little bit of help towards #112 :crossed_fingers:

### DIFF
--- a/src/modules/notifications.rs
+++ b/src/modules/notifications.rs
@@ -184,7 +184,7 @@ impl Module<Overlay> for NotificationsModule {
     {
         let overlay = Overlay::new();
         let button = Button::with_label(&self.icons.closed_none);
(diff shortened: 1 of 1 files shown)


JakeStanger:
Awesome, thank you! Are you just able to show a quick screenshot or other evidence that it's working now pls just before I merge? Unable to test and the moment and obvs CI is broken.

I'll then merge and update the porting doc :)

JakeStanger:
Lovely, happy with that. Thanks again!

Issue #862, opened 2025-02-04
`workspace` and `keyboard` feature flags have some incorrect cross-dependency


JakeStanger:
**Describe the bug**
Enabling one of `workspace` or `keyboard` without the other fails to compile due to some incorrect/missing feature flags.

**Expected behavior**
You should be able to independently enable either feature flag.

**System information:**
 - Ironbar version: [68ccc7a](https://github.com/JakeStanger/ironbar/commit/68ccc7ac4d1ac3f5c70891973138acce5fb2a851)

**Additional context**

Raised here: https://github.com/JakeStanger/ironbar/pull/855#issuecomment-2632505664

PR #932, opened 2025-04-09
Fix compilation for many feature combinations

Rodrigodd:
This is a extended fix for the issue I pointed out in https://github.com/JakeStanger/ironbar/pull/855#issuecomment-2632505664 and tracked by the issue #862 which was not fully fixed by #910. This also fixes many issues with other features.

I use the following shell script to test the changes:

```sh
#!/bin/sh

check() {
    echo "Checking $1"
    cargo check --no-default-features --features "$@" 1>/dev/null 2>/dev/null
}

cargo check \
&& check config+json \
&& check config+yaml \
&& check config+toml \
&& check config+ron \
&& check config+all \
&& check config+yaml,keyboard+all \
&& check config+yaml,workspaces+all \
&& check config+yaml,workspaces+all,keyboard+all \
&& check config+yaml,workspaces+sway,keyboard+hyprland \
&& check config+yaml,workspaces+hyprland,keyboard+sway \
&& check config+yaml,workspaces+hyprland,keyboard+sway \
&& check config+yaml,workspaces,hyprland,keyboard,sway \
&& check config+yaml,workspaces \
&& check config+yaml,keyboard \
&& check config+yaml,sway \
&& check config+yaml,niri \
&& check config+yaml,hyprland \
&& check config+yaml,music+mpris \
&& check config+yaml,music+mpd \
&& check config+yaml,cli \
&& check config+yaml,cairo \
&& check config+yaml,clipboard \
&& check config+yaml,clock \
&& check config+yaml,config+all \
&& check config+yaml,custom \
&& check config+yaml,focused \
&& check config+yaml,http \
&& check config+yaml,ipc \
&& check config+yaml,keyboard+all \
&& check config+yaml,launcher \
&& check config+yaml,label \
&& check config+yaml,music+all \
&& check config+yaml,network_manager \
&& check config+yaml,notifications \
&& check config+yaml,script \
&& check config+yaml,sys_info \
&& check config+yaml,tray \
&& check config+yaml,upower \
&& check config+yaml,volume \
&& check config+yaml,workspaces+all \
&& check config+yaml,schema \
&& echo "Check passed" \
|| echo "Check failed"

# Fails to compile, cannot get a default for PlayerType
# && check config+yaml,music \

# universal-config fails to compile
# && check ''
# && check config+corn \
```

Some notes:
- I added `config+yaml` as a feature to all checks, because without it `universal-config` fails to compile (but that is fixed by https://github.com/JakeStanger/universal-config-rs/pull/8).
- I was careful with features like `keyboard+sway` and `workspace+hyprland` to only enable the implementation of the module for the given compositor when those specific features are enable, and not when only `keyboard` and `sway` are both enabled, for example.
- I replace cfg checks like `#[cfg(any(feature = "keyboard+hyprland", feature = "workspaces+hyprland"))]` with a feature for each compositor (like `#[cfg(feature = "hyprland")` in that case). This will scale better when more compositor aware modules are added (like my `bindmode+sway` and `bindmode+hyprland` in the near future).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,11 +30,11 @@ default = [
     "tray",
     "upower",
     "volume",
-    "workspaces+all"
+    "workspaces+all",
 ]
 
 cli = ["dep:clap", "ipc"]
-ipc = ["dep:serde_json"]
+ipc = ["dep:serde_json", "dep:clap"]
 
 http = ["dep:reqwest"]
 
@@ -61,7 +61,7 @@ custom = []
 
 focused = []
 
-keyboard = ["dep:colpetto", "dep:evdev-rs", "dep:rustix"]
+keyboard = ["dep:colpetto", "dep:evdev-rs", "dep:rustix", "futures-lite"]
 "keyboard+all" = ["keyboard", "keyboard+sway", "keyboard+hyprland"]
 "keyboard+sway" = ["keyboard", "sway"]
 "keyboard+hyprland" = ["keyboard", "hyprland"]
@@ -81,7 +81,7 @@ notifications = ["zbus"]
 
 script = []
 
-sys_info = ["dep:sysinfo"]
+sys_info = ["dep:sysinfo", "ipc"]
 
 tray = ["system-tray"]
 
@@ -93,9 +93,11 @@ workspaces = ["futures-lite"]
 "workspaces+all" = ["workspaces", "workspaces+sway", "workspaces+hyprland", "workspaces+niri"]
 "workspaces+sway" = ["workspaces", "sway"]
 "workspaces+hyprland" = ["workspaces", "hyprland"]
-"workspaces+niri" = ["workspaces"]
+"workspaces+niri" = ["workspaces", "niri"]
 
 sway = ["swayipc-async", "futures-lite"]
+
+niri = ["dep:serde_json"]
 
 schema = ["dep:schemars"]
 
--- a/src/clients/compositor/hyprland.rs
+++ b/src/clients/compositor/hyprland.rs
@@ -1,4 +1,4 @@
-use super::{KeyboardLayoutClient, KeyboardLayoutUpdate, Visibility, Workspace, WorkspaceUpdate};
+use super::{Visibility, Workspace, WorkspaceUpdate};
 use crate::{arc_mut, lock, send, spawn_blocking};
 use color_eyre::Result;
 use hyprland::ctl::switch_xkb_layout;
@@ -11,24 +11,39 @@ use tokio::sync::broadcast::{Receiver, Sender, channel};
 use tracing::{debug, error, info};
 
 #[derive(Debug)]
-pub struct Client {
-    workspace_tx: Sender<WorkspaceUpdate>,
-    _workspace_rx: Receiver<WorkspaceUpdate>,
+struct TxRx<T> {
+    tx: Sender<T>,
+    _rx: Receiver<T>,
+}
 
-    keyboard_layout_tx: Sender<KeyboardLayoutUpdate>,
-    _keyboard_layout_rx: Receiver<KeyboardLayoutUpdate>,
+#[derive(Debug)]
+pub struct Client {
+    #[cfg(feature = "workspaces+hyprland")]
+    workspace: TxRx<WorkspaceUpdate>,
+
+    #[cfg(feature = "keyboard+hyprland")]
+    keyboard_layout: TxRx<KeyboardLayoutUpdate>,
 }
 
 impl Client {
     pub(crate) fn new() -> Self {
+        #[cfg(feature = "workspaces+hyprland")]
         let (workspace_tx, workspace_rx) = channel(16);
+
+        #[cfg(feature = "keyboard+hyprland")]
         let (keyboard_layout_tx, keyboard_layout_rx) = channel(16);
 
         let instance = Self {
-            workspace_tx,
-            _workspace_rx: workspace_rx,
-            keyboard_layout_tx,
-            _keyboard_layout_rx: keyboard_layout_rx,
+            #[cfg(feature = "workspaces+hyprland")]
+            workspace: TxRx {
+                tx: workspace_tx,
+                _rx: workspace_rx,
+            },
+            #[cfg(feature = "keyboard+hyprland")]
+            keyboard_layout: TxRx {
+                tx: keyboard_layout_tx,
+                _rx: keyboard_layout_rx,
+            },
         };
 
         instance.listen_workspace_events();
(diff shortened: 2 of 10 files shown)


Cargo.toml
```diff
@@ -30,11 +30,11 @@ default = [
     "tray",
     "upower",
     "volume",
-    "workspaces+all"
+    "workspaces+all",
 ]
 
 cli = ["dep:clap", "ipc"]
```
JakeStanger, line 36: Can clap be removed from here please, as it's now provided by the required IPC feature

Cargo.toml
```diff
@@ -61,7 +61,7 @@ custom = []
 
 focused = []
 
-keyboard = ["dep:colpetto", "dep:evdev-rs", "dep:rustix"]
+keyboard = ["dep:colpetto", "dep:evdev-rs", "dep:rustix", "futures-lite"]
```
JakeStanger, line 64: Can the comment for the `futures-lite` declaration be updated to include `keyboard` please

Cargo.toml
```diff
@@ -81,7 +81,7 @@ notifications = ["zbus"]
 
 script = []
 
-sys_info = ["dep:sysinfo"]
+sys_info = ["dep:sysinfo", "ipc"]
```
JakeStanger, line 84: This module should not depend on `sys_info`. If there are IPC-specific features, they either need uncoupling or feature-gating accordingly.

Cargo.toml
```diff
@@ -93,10 +93,12 @@ workspaces = ["futures-lite"]
 "workspaces+all" = ["workspaces", "workspaces+sway", "workspaces+hyprland", "workspaces+niri"]
 "workspaces+sway" = ["workspaces", "sway"]
 "workspaces+hyprland" = ["workspaces", "hyprland"]
-"workspaces+niri" = ["workspaces"]
+"workspaces+niri" = ["workspaces", "niri"]
 
 sway = ["swayipc-async", "futures-lite"]
 
+niri = ["dep:serde_json"]
```
JakeStanger, line 100: Can `serde_json`'s declaration be moved from the IPC section to the shared section and commented accordingly please

src/clients/compositor/hyprland.rs
```diff
@@ -38,8 +53,11 @@ impl Client {
     fn listen_workspace_events(&self) {
```
JakeStanger, line 53: This method should have a feature flag, rather than everything inside of it

src/clients/mod.rs
```diff
@@ -7,7 +7,7 @@ use std::sync::Arc;
 
 #[cfg(feature = "clipboard")]
 pub mod clipboard;
-#[cfg(any(feature = "keyboard", feature = "workspaces"))]
+#[cfg(any(feature = "keyboard", feature = "workspaces", feature = "hyprland"))]
```
JakeStanger, line 10: What is the reasoning for adding this flag?

src/image/gtk.rs
```diff
@@ -9,7 +9,10 @@ use std::ops::Deref;
     feature = "clipboard",
     feature = "keyboard",
     feature = "music",
-    feature = "workspaces"
+    feature = "workspaces",
+    feature = "cairo",
+    feature = "clipboard",
+    feature = "launcher",
```
JakeStanger, line 15: When multiple flags are added like this, can they be preserved in alphabetical order please? It makes the lists easier to maintain. Same applies below in this file, and in `image/mod.rs`

JakeStanger (changes requested):
Thanks for this. Definitely will be good to have this sorted once and for all, and I think the changes here are generally good, and help tidy things up a bit. 

I've raised a few bits that ideally need sorting but nothing major.

As with the universal-config changes. I'll follow up the PR with CI changes to run the checks in a matrix. I've updated that crate so the `config+yaml` fix will no longer be required.

Cargo.toml
Rodrigodd: I just added a `#[cfg(feature = "ipc")]` in the code inside the sysinfo module that uses the IPC feature. Do you prefer that I add a `sysinfo+ipc` feature instead? That module is the only one using the ironvar feature currently as far as I see, so I don't have a example to guide me.

src/clients/compositor/hyprland.rs
Rodrigodd: That function have a single section for the "keyboard" feature, that is why I did that way. So now I splitted that function into `listen_workspace_events` and `listen_keyboard_events`, each one creating its own connection to Hyprland's IPC socket (I am currently doing the same in my bindmode module).

Thinking now, I could have kept the single connection, and split the part where the cfg repeated the most in a helper function (although there would still be a couple cfg's scattered inside the method). Let me know what you prefer.

src/clients/mod.rs
Rodrigodd: The Hyprland client (`compositor::hyprland::Client`) is inside the `compositor` module (as are the `keyboard_layout` and `workspaces` clients).

Rodrigodd:
Addressed all issues as best as an can. I kept each change in its own commit, let me know if you prefer that I squash them.

Cargo.toml
JakeStanger: Just `"ipc"` makes sense here. I think having flags for every part of the IPC would be overkill.

src/clients/compositor/hyprland.rs
JakeStanger: Ah right sorry, I didn't spot that before. 

I'd say if you say keep a single connection that would be preferable, and just do your best to split them. Feel free to move stuff around and stick things inside feature-gated `{}` blocks if that helps.

src/clients/mod.rs
JakeStanger: Right. I forgot the sway code is sort of split out at the moment. Happy with this for now, we'll re-review for the bindmode changes.

JakeStanger:
Cheers, I've responded to your comments. I'll fully review when I can.

I don't mind whether you separate commits or squash - I'll re-review the whole thing prior to merge, and squash through GH if necessary.

src/clients/compositor/hyprland.rs
Rodrigodd: > stick things inside feature-gated {} blocks if that helps

That is what I would have done, but there were already many blocks for scope-only reasons, and I was trying to avoid to increase the indentation further.

But now I renamed the function to `listen_events` and extract part of it to `listen_workspace_events` and `listen_keyboard_events`. Worked very well, much clearer.

JakeStanger (approved):
Cool, LGTM. Thanks for changing those bits. I'll get working on the CI changes

PR #949, opened 2025-04-27
feat!: support glob patterns for keyboard layout icons

postsolar:
The keyboard module now uses glob pattern matching instead of exact string matching when looking up the layout icon from the user-provided map. This allows users to define more flexible rules for displaying icons based on layout names, for example, using `*` to match variations of a layout (`en_*`).

This was implemented using the `wildmatch` crate.

My usecase is that
```nix
    icons.layout_map = {
      "English (Colemak-DH ISO)" = "🇺🇸";
      "Estonian"                 = "🇪🇪";
      "Russian (GOST 14289-88)"  = "🇷🇺";
    };

```
this looks kinda ugly and hard to manage. I'd prefer to just have `"English *"` instead.


JakeStanger:
Thanks for this. I think this is a good addition, but I think bringing in an extra dep for full globbing is a bit overkill.

 I reckon we could get away with a simple `starts_with` check (either in place of an exact match, or I guess ideally checking for an `*` at the end of the string).

postsolar:
I touch Rust maybe once a year so if there's something I got wrong I'd welcome it pointed out 🙃

src/modules/keyboard.rs
```diff
@@ -329,7 +329,20 @@ impl Module<gtk::Box> for KeyboardModule {
...
             }
             KeyboardUpdate::Layout(KeyboardLayoutUpdate(language)) => {
-                let text = icons.layout_map.get(&language).unwrap_or(&language);
+                let text = icons
+                    .layout_map
+                    .iter()
+                    .find_map(|(pattern, display_text)| {
+                        let is_match = if pattern.ends_with("*") {
+                            let pattern_stripped = pattern.strip_suffix("*").unwrap();
+                            language.starts_with(pattern_stripped)
```
JakeStanger: You can use string slices here to avoid the extra allocation and the unwrap:

```suggestion
                            language.starts_with(pattern[..pattern.len()-1])
```

(I've not tested that so you'll wanna make sure it's right)

JakeStanger:
> Update: now that I think about it, does icons.layout_map preserve the order from the config, and does .iter() preserve that order too? Obviously not something really important in practice, but I guess we would want Enghlish (Colemak* to match earlier than English* if the former comes first in the config.

`HashMap` does not preserve order. An `IndexMap` will though. It's an external dep that's included, but its `serde` feature would need enabling.

postsolar:
> don't be afraid to stick in another comment so I'm notified :)

Noted :)

I tested it with `English (Colemak*` and `English*` (in this order in the config) and it was matching things correctly.

docs/modules/Keyboard.md
```diff
@@ -25,7 +25,7 @@ Displays the toggle state of the capslock, num lock and scroll lock keys, and th
 | `icons.num_off`    | `string` or [image](images)    | `''`    | Icon to show for disabled num lock indicator.                                                                             |
 | `icons.scroll_on`  | `string` or [image](images)    | ``     | Icon to show for enabled scroll lock indicator.                                                                           |
 | `icons.scroll_off` | `string` or [image](images)    | `''`    | Icon to show for disabled scroll lock indicator.                                                                          |
-| `icons.layout_map` | `Map<string, string or image>` | `{}`    | Map of icons or labels to show for a particular keyboard layout. Layouts use their actual name if not present in the map. |
+| `icons.layout_map` | `Map<string, string or image>` | `{}`    | Map of icons or labels to show for a particular keyboard layout. Layouts use their actual name if not present in the map. Layouts are matched in the order they appear in the map. If a pattern to match ends with a `*`, it acts as a wildcard, matching any layout name that begins with the part before the `*` |
```
postsolar, line 28: To be honest I wasn't sure of the right formatting here (the last `|`), and erred on the side of smaller diffs. Does this fit?

JakeStanger: I think that's fine. My editor might auto-format at some point in the future but I'm not too fussed.

src/modules/keyboard.rs
JakeStanger: I think this comment possibly got lost somewhere, but if you're able to amend I'll get this merged

postsolar:
@JakeStanger is there something else to be done before this could be merged?

JakeStanger:
Yeah there's still an outstanding comment that's still relevant, if you can address that please:

image

I assume GitHub just ate the thread because it's "outdated"?

postsolar:
I'm not seeing this thread at all, even an outdated one. Given it says "Pending" on your screenshot, maybe you forgot to hit "Review" button? This happened to me too recently :grin:

JakeStanger:
> Given it says "Pending" on your screenshot

:facepalm: :facepalm: :facepalm:

postsolar:
Yep, as awesome as GitHub is, these "Review" and "Pending" labels are not quite obvious :grin: 

I addressed that comment now :+1:

Issue #950, opened 2025-04-29
[Feature Request] NixOS home-manager module - automatically reload ironbar, once config changes


74k1:
**Is your feature request related to a problem? Please describe.**
The current Nix home-manager Module doesn't allow ironbar to reload, once the configuration has changed.
The Ironbar CSS file from the home-manager Module also doesn't get loaded at all initially.

This requires the user, even if `programs.ironbar.systemd = enable;` is set, to have to manually restart ironbar with `ironbar restart` and load the css once / at start with `ironbar load-css ~/.config/ironbar/ironbar.css`.

**Describe the solution you'd like**
I've already implemented a solution to this problem here https://github.com/74k1/ironbar/tree/update-flake. As per [CONTRIBUTING.md](https://github.com/JakeStanger/ironbar/blob/master/CONTRIBUTING.md) I'd first like to check if @JakeStanger is open for a PR that would solve this minor issue. :)

My changes include:
- set `onChange` (simple `ironbar reload` command) for `xdg.ConfigFile."ironbar/config.json"`
- use lib.mkEnableOptions for systemd option as well to make it disabled by default
- slight modifications to the systemd options
  - add Documentation URL
  - use `config.wayland.systemd.target` instead of `"graphical-session.target"` to `PartOf` and `Install.WantedBy`
  - add "tray.target" to `PartOf` and `Install.WantedBy`
  - add `ExecReload` command
  - add optional `ExecPost` if `cfg.style != ""` for loading css on service start
  - matched the format more to waybar's home-manager systemd options

**Describe alternatives you've considered**
We could consider submitting this to [home-manager](https://github.com/nix-community/home-manager) upstream later, once we've refined the module further.

**Additional context** 
After initial approval, I'd start updating the documentation for nix as per contributing guidelines.
And additionally, if it's okay, I'd like to reorganize the Nix flake structure to improve modularity and visibility, rather than keeping everything in flake.nix and nix/default.nix.

Thanks!

JakeStanger:
Yes this all sounds good to me, thanks for checking :)

The nix code does need some love but it's outside of my wheelhouse a bit so glad it's getting the attention it deserves. Happy to take on PRs for everything mentioned.

PR #953, opened 2025-05-01
feat(nix)!: standardize home-manager options & restructure flake

74k1:
This is a draft PR, since I haven't gotten around restructuring the flake into modules. - Once that's done, I'll edit this message.

For now, changes include:

- set `onChange` (simple `ironbar reload` command) for `xdg.ConfigFile."ironbar/config.json"`
- use lib.mkEnableOptions for systemd option as well to make it disabled by default
- slight modifications to the systemd options
  - add Documentation URL
  - use `config.wayland.systemd.target` instead of `"graphical-session.target"` to `PartOf` and `Install.WantedBy`
  - add "tray.target" to `PartOf` and `Install.WantedBy`
  - add `ExecReload` command
  - add optional `ExecPost` if `cfg.style != ""` for loading css on service start
  - matched the format more to waybar's home-manager systemd options (**breaking**, see below)

---

BREAKING CHANGE: `programs.ironbar.systemd` is no longer enabled by default

(opinionated) reasons for this breaking change being: 
- `programs` should not have a service enabled by default. For those cases, home-manager has `service` options too.
- one could seperately add `${lib.getExe pkgs.ironbar}` to their own Window Manager (for niri: `spawn-at-startup`, for hyprland: `exec-once`) to get ironbar running. - and therefore would just want/need the `programs.ironbar` option to configure ironbar.

I'm mainly basing this off of `waybar`'s home-manager options.

---

closes: #950


74k1:
Just pushed a new commit. Many thanks to @reo101 for sanity checking.

During our checks, we found that any change outside of the main build source files would result in different derivatives. We are working on making it deterministic :+1:

Afterwards, this PR would be ready for a review. :)

nix/module.nix
```diff
@@ -0,0 +1,92 @@
...
+
+        Service =
+          {
+            ExecReload = "${lib.getExe cfg.package} reload";
+            ExecStart = "${lib.getExe cfg.package}";
+            KillMode = "mixed";
+            Restart = "on-failure";
+          }
+          // lib.optionalAttrs (cfg.style != "") {
+            ExecPost = "${lib.getExe cfg.package} load-css ${config.xdg.configHome}/ironbar/style.css";
```
JakeStanger, line 80: What is the purpose of this line?

nix/module.nix
```diff
@@ -0,0 +1,92 @@
...
+          };
+
+        Install.WantedBy = [
```
JakeStanger, line 83: Is there a `niri` target? If so would be good to get that added here

nix/module.nix
74k1: I noticed that the css wouldn't load on start of the ironbar service. So I added these two lines to add the loading of the css.

ExecPost essentially waits until the service started, then executes `ironbar load-css <location of the css file>` if `cfg.style` is not empty.

I use `config.xdg.configHome` because we use `xdg.configFile` to create `ironbar/style.css` and the user could have `config.xdg.configHome` set anywhere else than `~/.config`.

I hope that answered your question. :)

nix/module.nix
74k1: `niri` is not yet upstreamed in home-manager. There is a small option in nixpkgs, but configuring niri is more widely used on nixos-systems using https://github.com/sodiboo/niri-flake.
I'll see if there's a target for both options (nixpkgs nixosmodule & niri-flake) and add both with the necessary mkIf checks, if they're not the same.

74k1: Even with the usage of https://github.com/sodiboo/niri-flake, we don't get a new attribute under `config.wayland.windowManager`. The flake also doesn't provide any `niri.target` (yet?). I believe, if and once niri-flake gets upstreamed (https://github.com/sodiboo/niri-flake/issues/480), there will be an option.
But for now, this doesn't seem to be the case.

As additional context, I am currently using that flake. It does work without the target thanks to `config.wayland.systemd.target`. :)

nix/module.nix
JakeStanger: Oh strange. It should be loaded as part of initializing the bar, which isn't part of any event or anything, so not sure why it wouldn't.

nix/module.nix
JakeStanger: Cool, I'll keep an eye on that and look at adding it myself in the future. Cheers for the info.

JakeStanger (approved):
LGTM, big thanks for these changes

JakeStanger:
Are you just able to squash any commits that are affecting previous commits in the PR down please? There's a few individual items here I want to preserve so don't want to squash it all the way down. 

Once that's done, I'll merge. Thanks again

74k1:
Happy to contribute! :tada:

JakeStanger:
Perfect thank you

74k1:
Hi @atagen 

Happy to chime in and answer your questions with what little I know.

> why does this build with naersk, crane and rustPlatform?
I did not change the package definition. Merely renamed it. I could imagine it was used for tests. Jake seems to have confirmed my suspicions.

> ... but all 3 doesn't really make sense. with a binary cache serving it to the majority of consumer ...

The cache only gets built with naersk. (see https://github.com/JakeStanger/ironbar/blob/89073dc0cfa61a079ecf6a0e723811611a6ed122/flake.nix#L97)

I could also think of other maintainers, trying other builders than naersk. So in _my_ opinion, we could keep it in the flake.

> what does flake-parts add here besides the platform abstraction?

Again, I'm no expert in Nix. Perhaps @reo101 could help me out with this question.
I've been seeing flake-parts almost everywhere, and thought it could be a good idea to implement it as well, rather than the trivial replication of the "systems stuff" that you've mentioned.

> I agree a systemd service activated under a programs option is surprising in nixos/hm status quo, but shouldn't the option simply be moved to services.ironbar.enable in the first place? programs.ironbar.systemd can become a (gradually deprecated) alias for this if needed, for backwards compatibility.

I was basing this off of `programs.waybar` - which is also a _bar_.
I also think, that people might just want to configure the settings and run it themselves using execStart in their WM, instead of running it as a systemd service (as I mentioned before already)

I'd rather keep it as `programs.ironbar` than `services.ironbar`.
I'm unsure, what the majority of people would think. - therefore this was **opinionated**, as mentioned in OP.

reo101:
Hey @74k1 && @atagen

Happy to chime in as well.

> what does flake-parts add here besides the platform abstraction? the systems stuff can be trivially replicated without lockfile pollution for consumers, and the only fp import I see is the devshell, which can be broken out the same way in vanilla nix.

I'm also on-board on reducing unneeded abstraction where it's not needed. Here, there is an argument against using libraries like `flake-utils` and `flake-parts`, yes. This [DetSys article](https://determinate.systems/posts/best-practices-for-nix-at-work/#avoid-flake-helper-libraries-if-possible) talks about the same thing and I agree with them too.

About the lockfile pollution, `flake-parts` is *just* one flake, with only one dependency - which is a [*fork* of `nixpkgs`](https://github.com/nix-community/nixpkgs.lib), which is merely `nixpkgs`, but purely the `lib` part of it (simply `nix`, no `pkgs`, kek). Having noted that, I don't really think this is a big burden on consumers' flakes.
So, at this point, I'd say it's mostly personal preference.

I myself have gone through the path of `flake-utils` (because I saw some people were using it) -> hand-rolled `forEachSystem`-style helpers -> `flake-parts` (hand-rolling everything suddenly became unbearably complex with me recreating the module system a bit, by accident).
I'd say that this project (`ironbar`) has sufficiently many *things* (`package`, `devShell`, `overlay` and `nixosModule`) it does through nix that it justifies the usage of modules to separate concerns, and hence, `flake-parts` is a good solution (again, not purely because of the `perSystem` helper (which is still nice), but more because of the easier, streamlined modularization we can do with it) instead of hand-rolling stuff with a lot of `import`-ing.

If you beg to differ (would love to hear some solid arguments against it), I'd love to have a discussion on the topic! :smile:

Cheers

JakeStanger:
> would the best one be naersk, given that's what the CI uses?

That was at least the case two years ago. Since moving to Cachix for Nix builds they've been fast enough and cached well enough that I doubt it makes much difference regardless.

> also, could I ask about the use of rust-overlay?

Looks like that's been there since the initial flake code back in Nov 2022. Ironbar targets the latest stable Rust version always, so sounds like that can go.

---

I know very little about Nix (just about enough to get by as a user, and not really enough to write working derivations) so I cannot review incoming PRs with as much rigour as I'd like. There's not really any reason for anything being the way it is other than that's what got written, so if you do see something that looks weird or out of place or done badly, chances are you're right. Improvements always welcome :)

PR #959, opened 2025-05-06
Add support to change the volume scrolling on the volume icon

tglman:
Hi,

This add the ability to turn up and down the volume simply scrolling on the volume icon.

it works now but it's still a bit of a work in progress.

I do not have hardware(for another month) to test scroll with a mouse wheel, I think in that case I need to add a `-` for some events, that's why there are two different event signals up and down, otherwise would be enough one change (maybe still can be simplified to only one event).

Also I saw the WIP migration of GTK4, maybe is easier to way that the migration is done and then rework this on top of that.

Just opening this to make this feature public, even if not yet mainstream.

### DIFF
--- a/src/clients/volume/mod.rs
+++ b/src/clients/volume/mod.rs
@@ -29,6 +29,8 @@ pub enum Event {
     AddInput(SinkInput),
     UpdateInput(SinkInput),
     RemoveInput(u32),
+    VolumeUp(f64),
+    VolumeDown(f64),
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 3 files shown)


JakeStanger:
Rebased onto latest master.

Currently there's a bug using the mouse wheel, where if you scroll too fast it can crash the client, and take down Ironbar with it:

```
Assertion '*_head == _item' failed at ../pulseaudio/src/pulse/operation.c:85, function operation_unlink(). Aborting.

Process finished with exit code 134 (interrupted by signal 6:SIGABRT)
```

I'm going to convert this to a draft PR for now

PR #978, opened 2025-05-12
fix(mpris client): use set_position instead of seek

### LINKED ISSUE #970: popup-music: seeking position doesn't use precise position but instead changes it in steps of 5 seconds
**Describe the bug**

When seeking position, it doesn't set the position to its absolute value but instead increments or decrements it in intervals of 5 seconds. So, if position at, say, 03:00, and I try to set it to 04:00, it would instead be set to 03:05. And vice versa, if it's at 03:00 and I try to set it to 02:00, it would be set to 02:55 instead.

**To Reproduce**

1. Add module `music`
2. Click popup
3. Try to seek position

**Expected behavior**

Position seeking works

**System information:**
- Distro: NixOS
- Compositor: Hyprland
- Ironbar version: 0.16.1

**Configuration**

```json
{
  "start": [
  ],
  "center": [
    {
      "truncate": {
        "max_length": 60,
        "mode": "start"
      },
      "type": "music"
    }
  ],
  "end": [
  ],
  "height": 30,
  "name": "hyprbar",
  "position": "top"
}
```

</details>

**Additional context**

playerctl v2.4.1, launched in daemon mode (`playerctld`)

postsolar:
Closes #970

Tested it in Firefox, mpv and Telegram, seems to work well

### DIFF
--- a/src/clients/music/mpris.rs
+++ b/src/clients/music/mpris.rs
@@ -290,14 +290,22 @@ impl MusicClient for Client {
 
     fn seek(&self, duration: Duration) -> Result<()> {
         if let Some(player) = Self::get_player(self) {
-            let pos = player.get_position().unwrap_or_default();
+            // if possible, use `set_position` instead of `seek` because some players have issues with seeking
+            // see https://github.com/JakeStanger/ironbar/issues/970
+            if let Ok(metadata) = player.get_metadata() {
+                if let Some(track_id) = metadata.track_id() {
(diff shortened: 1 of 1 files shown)


JakeStanger:
Thanks for this.

I reckon just to be on the safe side, if there is no track id it should fall back to seek instead of being a no op. Should cover all bases that way.

JakeStanger (approved):
LGTM, thanks:)

PR #983, opened 2025-05-16
fix(bindmode): hide bindmode label if it's empty

### LINKED ISSUE #982: feature request: don't show the bindmode when it's the default mode

If the bar has a particular styling (the module for bindmode is visually isolated), there would be an empty block for the "default" or "global" bindmode (`submap = reset` in Hyprland lingo). This is visually unappealing and provides no useful information.

Non-default mode on:
Image

Default mode on:
Image

I'd like to just not see the widget when the bindmode is the default one.

There are many ways to address it. Off the top of my head I could think of:
1. Don't display the widget if the default bindmode is active, add it as a config option
2. Add a CSS selector for the default mode
3. Add a mode filtering list, e.g.
```json
{
  "type": "bindmode",
  "except": [ "", "some_other_mode" ]
}
```

1. Implementable with Ironvars dynamic booleans and the `show_if` field for `label` widgets. Ok but extra code.
2. Implementable via `load-css` command. Quite hacky but works too.

Both could be done with either:

- No Hyprland IPC, just `bind = mod, key, exec, hyprctl dispatch submap foo && ironbar var set submap foo && ironbar var set submapVisible true`, which adds quite a bit of coupling between Hyprland config and Ironbar
- Hyprland IPC, which keeps the config cleaner, but less accessible to a lot of users

**Additional context**

postsolar:
Closes #982

This works but I'm not sure how to also make it hidden upon initialization (needed because there would be no `BindModeUpdate`'s at this point yet). Tried to insert `label.hide()` right after building the widget but it doesn't work.

### DIFF
--- a/src/modules/bindmode.rs
+++ b/src/modules/bindmode.rs
@@ -79,6 +79,12 @@ impl Module<Label> for Bindmode {
                 trace!("mode: {:?}", mode);
                 label.set_use_markup(mode.pango_markup);
                 label.set_label_escaped(&mode.name);
+
+                if mode.name.is_empty() {
+                    label.hide();
+                } else {
+                    label.show();
+                }
             };
 
             context.subscribe().recv_glib(on_mode);
--- a/src/modules/bindmode.rs
+++ b/src/modules/bindmode.rs
@@ -2,7 +2,7 @@ use crate::channels::{AsyncSenderExt, BroadcastReceiverExt};
 use crate::clients::compositor::BindModeUpdate;
 use crate::config::{CommonConfig, LayoutConfig, TruncateMode};
 use crate::gtk_helpers::IronbarLabelExt;
-use crate::modules::{Module, ModuleInfo, ModuleParts, WidgetContext};
+use crate::modules::{Module, ModuleInfo, ModuleParts, ModuleUpdateEvent, WidgetContext};
 use crate::{module_impl, spawn};
 use color_eyre::Result;
 use gtk::Label;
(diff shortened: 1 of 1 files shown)


JakeStanger:
This is something I hit with the keyboard module too. I may at some point need to add some `start_hidden` or similar method to the `Module` trait to handle this, but for now the only real way is to send a dummy event from the controller on init.

JakeStanger:
I'd say this isn't really a breaking change by the way, and I'd probably consider it a fix rather than a feature. 

The module showing when empty was unexpected/undesired, I wouldn't expect this to break any reasonable setup, and it doesn't introduce any config/api changes.

src/modules/bindmode.rs
```diff
@@ -71,13 +71,31 @@ impl Module<Label> for Bindmode {
...
         }
 
+        // Send a dummy event on init so that the widget starts hidden
+        {
+            let tx = context.tx.clone();
+            try_send!(
+                tx,
+                ModuleUpdateEvent::Update(BindModeUpdate {
+                    name: String::new(),
+                    pango_markup: true
```
postsolar, line 81: I didn't understand what is the purpose of markup here so just used the value that's used during label building

postsolar:
I cannot reproduce this CI failure. Just for a sanity check, it IS supposed to use the lockfiles and all that from my branch, right?

JakeStanger:
It should do. Did you rebase onto/merge master? I've broken the build in my last commit :/

postsolar:
No, I didn't rebase. I've seen some recent commits but I don't understand how/why they would impact the CI on my branch. Or are you saying this, too?

JakeStanger:
Yeah I'm not sure why it's affecting your build either. I wonder if the logic is:

- If diffs for workflows present in PR, use the PR version
- Otherwise, use latest master version

I guess that would allow you to make changes to workflows that instantly affect all outstanding PRs, so it does make sense.

I'm going to get the build sorted now.

postsolar:
Ah, so the CI tests not the PR branch but master branch with PR changes applied. Yeah this makes sense, actually more than testing the PR branch 😁. I didn't know this, neat.

JakeStanger (approved):
LGTM, cheers!

PR #988, opened 2025-05-18
feat: add halign option to box module

### LINKED ISSUE #985: Feature Request: To set horizontal alignment of boxes inside their parent container

Currently, child widgets are packed to the left inside their parent boxes without any option present to alter this behaviour. This creates a potential headache for users who want to define a Popup with many vertically aligned boxes containing potentially multiple widgets centered horizontally.

So for example:

Image

This shows a popup containing several different nested widgets with the following structure:

button
[box[button, slider]]
[box[button]]
[box[button, button, button]]

Ideally, the button in the third row could be center aligned, which could be achieved with css margins assuming the popup width is static, but the slider in row two is only shown after the button before it is clicked, changing the popup width with it.

Basically, I want the ability to set the halign property of a GtkBox to GTK_ALIGN_CENTER via some box module option. While this wouldn't actually center align the widgets inside the box, it would give users a working solution to the problem above with minimal changes to the codebase.

Tried finding some combination of CSS properties and nested buttons and boxes to accomplish the look I was going for but it doesn't seem possible currently.

BBFifield:
Closes #985

This implementation makes a working solution possible for the issue described in #985. I'm a little surprised the issue regarding box alignment wasn't raised by someone in the past unless I'm missing something.

Documentation has also been written for the added feature.

Thanks.

### DIFF
--- a/src/modules/custom/box.rs
+++ b/src/modules/custom/box.rs
@@ -6,6 +6,31 @@ use gtk::prelude::*;
 use serde::Deserialize;
 
 #[derive(Debug, Deserialize, Clone)]
+#[serde(rename_all = "lowercase")]
+#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
+pub enum ModuleAlignment {
+    /// Align widget to the start (left for horizontal, top for vertical).
+    Start,
+    /// Align widget to the center.
+    Center,
+    /// Align widget to the end (right for horizontal, bottom for vertical).
+    End,
+    /// Stretch widget to fill available space.
+    Fill,
+}
+
+impl From<ModuleAlignment> for gtk::Align {
+    fn from(align: ModuleAlignment) -> Self {
+        match align {
+            ModuleAlignment::Start => gtk::Align::Start,
+            ModuleAlignment::Center => gtk::Align::Center,
+            ModuleAlignment::End => gtk::Align::End,
+            ModuleAlignment::Fill => gtk::Align::Fill,
+        }
+    }
+}
+
+#[derive(Debug, Deserialize, Clone)]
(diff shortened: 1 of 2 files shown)


src/modules/custom/box.rs
```diff
@@ -21,10 +46,15 @@ pub struct BoxWidget {
...
     /// **Valid options**: `horizontal`, `vertical`, `h`, `v`
-    /// <br />
     /// **Default**: `horizontal`
     orientation: Option<ModuleOrientation>,
 
+    /// Horizontal alignment for the box.
+    ///
+    /// **Valid options**: `start`, `center`, `end`, `fill`
+    /// **Default**: `fill`
+    halign: Option<ModuleAlignment>,
```
JakeStanger, line 56: My preference here, if not adding a pairing `valign` property, would be to rename this to just be called `align` and set its axis depending on the box orientation.

src/modules/custom/box.rs
```diff
@@ -47,6 +77,10 @@ impl CustomWidget for BoxWidget {
             }
         }
 
+        // Use provided halign, defaulting to Center if not specified.
```
JakeStanger, line 80: This comment is unnecessary and incorrect - please remove

JakeStanger (changes requested):

JakeStanger:
Thanks for the PR & for updating docs too. A couple of changes if you can please.

Don't worry about the build, I've got to fix that.

src/modules/custom/box.rs
BBFifield: Will do. Thanks for the quick review response!

JakeStanger:
LGTM, thanks!

PR #1003, opened 2025-05-25
Overhaul `.desktop` and image resolver code

JakeStanger:
This overhauls the `.desktop` parser and image resolver code to greatly improve their performance. 

This is achieved by caching all of the following:

- `.desktop` file locations
- `.desktop` file contents
- Image locations
- Image `Pixmap`s

The resolvers are fully async and only load data as required, including loading images without blocking the UI.

With the changes come a couple of new features:

- The system icon theme (as set through dconf/gsettings) will be used if no icon theme is configured. Previously this just fell back to Adwaita.
- The `icon_overrides` option now applies to ALL images across all modules, supporting all formats, making it much more powerful. For example you could set `icon_overrides.steam = "/path/to/steam.jpg"` or `icon_overrides.'music/Pink Floyd/Dark Side of the Moon/cover.jpg' = "https://example.com/dsotm.jpg"`.

**BREAKING CHANGE**: The PR moves the `icon_theme` option from the bar-level to the top-level. If setting a custom icon theme, this will need to be updated.


src/image/provider.rs
```diff
@@ -1,289 +1,350 @@
...
+    icon_theme: RefCell<Option<IconTheme>>,
+    overrides: HashMap<String, String>,
+    cache: Arc<Mutex<Cache>>,
+}
 
-        Some(Self { location, size })
+impl Provider {
+    pub fn new(desktop_files: DesktopFiles, overrides: &mut HashMap<String, String>) -> Self {
+        let mut overrides_map = HashMap::with_capacity(overrides.len());
+        overrides_map.extend(overrides.drain());
```
JakeStanger, line 68: might as well clone here

PR #1004, opened 2025-05-25
feat(keyboard): implement labels with icons

tobias47n9e:
I would like to build a purely text-based layout in the bar and was unable to configure the keyboard indicators to just use a `format` string similar to some of the other widgets that I have tried. I put a patch together, but it makes the code and config a bit repetitive (see caps-lock config below).

Would a patch like this be acceptable? I could put in some more time to clean it up. Or would it be better to split the keyboard indicators into 3 widgets (caps-lock, num-lock and layout)?

sway_screenshot

sway_screenshot_2

```toml
[[end]]
type = "keyboard"
show_caps = false
show_caps_icon = true
show_caps_format = true
show_layout = false
caps_format = "<span color='#0091e5'>CPS LCK\n[</span>{caps_icon}<span color='#0091e5'>]</span>"
```

### DIFF
--- a/src/modules/keyboard.rs
+++ b/src/modules/keyboard.rs
@@ -2,6 +2,7 @@ use std::collections::HashMap;
 
 use color_eyre::Result;
 use color_eyre::eyre::Report;
+use gtk::Label;
 use gtk::prelude::*;
 use serde::Deserialize;
 use tokio::sync::mpsc;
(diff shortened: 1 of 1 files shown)


JakeStanger:
Thanks for the PR. I really like the bar too, looks great!

This should have already been possible, but I've found a small bug in the icon code (resolved by #1005) preventing it.

I've just merged that, so if you rebuild on the latest version you should be able to do something like:

```toml
[[end]]
type = "keyboard"

[end.icons]
caps_on = "<span color='#0091e5'>CPS LCK\n[</span>^<span color='#0091e5'>]</span>"
caps_off = "<span color='#0091e5'>CPS LCK\n[ ]</span>"
```

Let me know if that suits, or if I'm missing something additional this brings?

tobias47n9e:
Thanks a lot! That solves my issues with caps lock. I still can't get the html to escape for the `icons.layout_map`. I can look into that in a separate merge request.

PR #1013, opened 2025-05-27
chore(nix): update to gtk4

postsolar:


JakeStanger:
@postsolar other than the fact that git seems to be throwing a tantrum about my six billion rebases on the target branch, is this good to go? 

Wasn't sure whether this is marked as draft because it's unfinished, or just to match the GTK4 branch

postsolar:
Both, actually. I think I missed a variable there, and obviously commented out fields would need to be removed, then an update would need to be re-run.

It's not particularly involved, but I'm quite short on time currently, so if anybody wants to finish it and submit a new PR, the steps to go about this (starting from scratch) are roughly:
1. Replace gtk3 packages with gtk4 packages in all files (use https://search.nixos.org/packages?channel=unstable to verify)
2. Run `nix flake update`

JakeStanger:
Cool, in that case I'll pick this up and see what I can do, as it's the last piece of the puzzle in the port (other than bugfixing...)

JakeStanger:
I've copied your changes into the main PR rather than merging this, because the Nix files were restructured at some point and rebasing onto that was looking like a nightmare. I've credited you in the commit still though :)

PR #1018, opened 2025-05-28
feat(keyboard): allow layout map to use markup

tobias47n9e:
Still working on my text only ironbar and was not able to escape the keyboard layout.

This commit allows to escape the keyboard layout:

layout_escaped

Using this config:

```toml
[end.icons.layout_map]
German = "<span color='#0091e5'>LAYOUT</span>\n<span color='#0091e5'>[</span>de<span color='#0091e5'>]</span>"
```

### DIFF
--- a/src/image/gtk.rs
+++ b/src/image/gtk.rs
@@ -69,6 +69,10 @@ impl IconButton {
     pub fn label(&self) -> &Label {
         &self.label
     }
+
+    pub fn set_label(&self, input: &str) {
+        self.label.set_label_escaped(input);
+    }
 }
 
 #[cfg(any(


JakeStanger:
This should already work without this patch. The existing `IconLabel` implementation will call `set_label_escaped` if the input is not detected to be an image:

https://github.com/JakeStanger/ironbar/blob/098df1ee26a42deacd31ed9aaedcfd40d7c12cb4/src/image/gtk.rs#L160-L170

Are you not seeing it work currently?

JakeStanger:
Oh no, I see the issue. It's an `IconButton` not `IconLabel` so it's using the button's in-built `set_label` method. 

The fix here is to extend `IconButton` to include a `set_label` method similar to that of `IconLabel`, rather than re-implementing in the module as you have here.

tobias47n9e:
@JakeStanger I updated the PR.

BTW do you get an event if you change the keyboard layout? In sway i use `swaymsg input "type:keyboard" xkb_layout en`, but that does not trigger an event in the keyboard module. So I was not able to test if changing keyboard layout works while ironbar is running (If i restart it, it works fine).

JakeStanger:
> BTW do you get an event if you change the keyboard layout?

It's definitely *supposed* to work. I don't have a working Sway or Hyprland setup at the moment so I can't easily test.

PR #1019, opened 2025-05-29
chore(nix): simplify packaging and outputs

atagen:
a lighter and more straightforward flake.
tested apps + packages + devshell. module is untouched aside from some lib inherits to clean up the code.

changed:
- removed overlay
- removed flake-parts, crane
- removed rust-overlay (nixpkgs-unstable should have latest stable rustc)
- brought devshell definition back inline
- refactored system-dependent outputs to use forAllSystems pattern
- added nix-systems to allow systems outputs to be overridden by consumers
- removed crane and default rustPlatform builds from flake+package
- hoisted some definitions in package to avoid requiring recursion
- added a default.nix for flake-compat
- formatted all files with alejandra


flake.nix
```diff
@@ -3,135 +3,86 @@
...
 
-  outputs = inputs @ {
+  outputs = {
     self,
     nixpkgs,
-    rust-overlay,
-    crane,
     naersk,
-    flake-parts,
+    nix-systems,
```
postsolar, line 15: Is an extra input really needed for 2 hardcoded strings?

JakeStanger:
Thanks for this. I merged a flake lock update which is conflicting, if you're able to resolve that please.

JakeStanger:
This LGTM based on my limited Nix knowledge. Thanks again

PR #1029, opened 2025-06-05
refactor: ported upower,volume,networkmanager modules to gtk4

tglman:
Hi, 

This migrate the volume and upower modules to gtk4

### DIFF
--- a/src/modules/battery.rs
+++ b/src/modules/battery.rs
@@ -1,24 +1,22 @@
-use crate::channels::{AsyncSenderExt, BroadcastReceiverExt};
-use crate::clients::upower;
-use crate::clients::upower::BatteryState;
-use crate::config::{CommonConfig, LayoutConfig};
-use crate::gtk_helpers::{IronbarGtkExt, IronbarLabelExt};
-use crate::image::IconLabel;
-use crate::modules::PopupButton;
-use crate::modules::{
-    Module, ModuleInfo, ModuleParts, ModulePopup, ModuleUpdateEvent, WidgetContext,
-};
-use crate::{module_impl, spawn};
 use color_eyre::Result;
 use futures_lite::stream::StreamExt;
 use gtk::{Button, prelude::*};
 use gtk::{Label, Orientation};
 use serde::Deserialize;
-use std::cmp::Ordering;
-use std::collections::HashMap;
-use std::fmt::{Display, Formatter, Write};
-use tokio::sync::mpsc;
-use zbus::zvariant::OwnedValue;
+use tokio::sync::{broadcast, mpsc};
+use zbus;
+use zbus::fdo::PropertiesProxy;
+
+use crate::channels::{AsyncSenderExt, BroadcastReceiverExt};
+use crate::clients::upower;
+use crate::clients::upower::BatteryState;
+use crate::config::{CommonConfig, LayoutConfig};
+use crate::gtk_helpers::{IronbarGtkExt, IronbarLabelExt};
+use crate::modules::PopupButton;
+use crate::modules::{
+    Module, ModuleInfo, ModuleParts, ModulePopup, ModuleUpdateEvent, WidgetContext,
+};
+use crate::{module_impl, spawn};
 
 const DAY: i64 = 24 * 60 * 60;
 const HOUR: i64 = 60 * 60;
@@ -45,34 +43,6 @@ pub struct BatteryModule {
     #[serde(default, flatten)]
     layout: LayoutConfig,
 
-    /// A map of threshold names to apply as classes,
-    /// against the battery percentage at which to apply them.
-    ///
-    /// Thresholds work by applying the nearest value
-    /// above the current percentage, if present.
-    ///
-    /// For example, using the below config:
-    /// ```corn
-    /// {
-    ///   end = [
-    ///     {
-    ///       type = "battery"
-    ///       format = "{percentage}%"
-    ///       thresholds.warning = 20
-    ///       thresholds.critical = 5
-    ///     }
-    ///   ]
-    /// }
-    /// ```
-    /// At battery levels below 20%,
-    /// the `.warning` class will be applied to the top-level widget.
-    /// Below 5%, `.critical` will be applied instead.
-    /// Above 20%, no class applies.
-    ///
-    /// **Default**: `{}`
-    #[serde(default)]
-    thresholds: HashMap<Box<str>, f64>,
-
     /// See [common options](module-level-options#common-options).
     #[serde(flatten)]
     pub common: Option<CommonConfig>,
@@ -86,7 +56,7 @@ const fn default_icon_size() -> i32 {
     24
 }
 
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug)]
 pub struct UpowerProperties {
     percentage: f64,
     icon_name: String,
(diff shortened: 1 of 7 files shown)


src/modules/upower.rs
```diff
@@ -274,7 +275,7 @@ impl Module<Button> for UpowerModule {
             label.set_label_escaped(&format);
         });
 
-        container.show_all();
+        container.show();
```
JakeStanger: You should be able to remove this as widgets are shown by default in gtk4

src/modules/upower.rs
```diff
@@ -177,7 +177,8 @@ impl Module<Button> for UpowerModule {
         let label = Label::builder()
             .label(&self.format)
             .use_markup(true)
-            .angle(self.layout.angle(info))
+            // TODO: find replacement
```
JakeStanger: I've not tested but angle should be settable using CSS in gtk4

src/modules/volume.rs
```diff
@@ -319,26 +320,24 @@ impl Module<Button> for VolumeModule {
...
         slider.set_value(50.0);
-        sink_container.add(&slider);
+        sink_container.append(&slider);
 
         {
             let tx = tx.clone();
             let selector = sink_selector.clone();
 
-            slider.connect_button_release_event(move |scale, _| {
+            slider.connect_value_changed(move |scale| {
```
JakeStanger: Changing the event type here could cause issues - unless debounced, this can overload the pulseaudio client and cause it to crash

src/modules/volume.rs
```diff
@@ -433,12 +431,12 @@ impl Module<Button> for VolumeModule {
...
                         }
 
-                        item_container.add(&label);
-                        item_container.add(&slider);
-                        item_container.add(&btn_mute);
-                        item_container.show_all();
+                        item_container.append(&label);
+                        item_container.append(&slider);
+                        item_container.append(&btn_mute);
+                        item_container.show();
```
JakeStanger: Can also be removed

JakeStanger (changes requested):
Thanks for this. A few notes if you can have a look please. The slider events might be a bit fiddly so give me a shout if you have any trouble with that

tglman:
added also the network manager

src/modules/volume.rs
tglman: Oh yep, just tried and crashed .... I could not find the other event in gtk4, will look deeper, any suggestion are welcome

src/modules/upower.rs
tglman: Hi, yep I was searching and it should, will remove it.

src/modules/volume.rs
JakeStanger: The event is still there, it's just been moved to the `GtkGestureClick` controller struct you have to create and attach. You'll find examples of that elsewhere in the port.

tglman: Hi, 

I tried and could not make it work, I though than to try another approach that was to rate limits the events, and seems to work ok, is a bit longer code but on the other side you can direct feedback on the volume level while you scroll it here the impl detail: https://github.com/JakeStanger/ironbar/pull/1029/commits/d632d4baf6fae234a09ca6b4d6ab7b4554b91106#diff-8def1800cfd70c672062713f20a02cd1af7d852bc2058887066768d171ba6698R195  the key part is the `sleep(30)` which is not really perceptible from the users but seems to save from a crash (at least in my tests)

JakeStanger: What issues did you run into that meant you couldn't get it to work? Something GTK4 specific?

This approach *may* work but it runs the risk that 30ms isn't enough on some machines, or that if events aren't processed fast enough (due to the delay) you'll let go of the slider and watch it partially "replay" the drag.

tglman: The issue I had is that I created the GestureClick connected the release event and add it as controller to the scale, but no event was fired when I switched the volume, will try to redo it and post it here.

Yes I thought the same 30ms may not be enough in some slow machines it could be increased probably up to 100ms without user feeling the gap(or made it configurable). anyway this patch actually send one event every 30ms and drop all the event that happen in the middle, so it won't replay the drag it will jump between states.

tglman: Actually do you know if is this is a reported bug in pulse audio, it may be worth to report it upstream.

JakeStanger: > but no event was fired when I switched the volume, will try to redo it and post it here.

Strange, deffo share when you can. As far as I'm aware in GTK4 everything gets all events by default.

> this patch actually send one event every 30ms and drop all the event that happen in the middle, so it won't replay the drag it will jump between states.

Ah cool, that might work okay then. If option A is a dead end, I'll give it a test when I can.

> do you know if is this is a reported bug in pulse audio

No idea. I think I did look before, but couldn't find anything. Pulse documentation is basically non-existent, so it could be that I'm doing something wrong and just don't know.

tglman: added  [this](https://github.com/JakeStanger/ironbar/pull/1029/commits/13336e0d59cc27e55044a3859c780ddc2f42ed46) commit that comment the delay implementation and use the connect_release event, but it does not work, so you can double check it.

tglman:
Hi,

I also added a commit that start the migration of the tray module, as now there are no compilation error, and the tray show with a new menu based on PopoverMenu, anyway I did not yet connect the actions, and also the UI seems a bit messy in my tests.

I have to discover yet how to connect the actions (I saw example using [SimpleActionGroup](https://gtk-rs.org/gtk-rs-core/stable/0.20/docs/gio/struct.SimpleActionGroup.html)) and how to compute the required parameters of [Client.activate](https://docs.rs/system-tray/0.7.0/system_tray/client/struct.Client.html#method.activate)

JakeStanger:
Really appreciate the undertaking there. Could you move the tray port into a separate PR please because that's gonna be a big one. It'll make it easier to maintain and hopefully means I can get the other ports merged first. 

Once that's picked across and removed here I'll give this another review and see where we're at.

tglman:
Hi,

Yes it does make sense to move the tray to a different PR, done it, with this commit: https://github.com/JakeStanger/ironbar/pull/1029/commits/13336e0d59cc27e55044a3859c780ddc2f42ed46 this changes try to do change the volume with the mouse released event, but it does not work, removing it, the changes will work with the value change event and the "rate limitated" event propagation.

I'm actually feeling a bit sorry for stealing all the fun of the migration  :)

JakeStanger:
Perfect cheers, I'll try and review this evening if I can.

> I'm actually feeling a bit sorry for stealing all the fun of the migration :)

I've certainly had more fun :P 

I'm very glad to have somebody else helping out, so seriously a huge thanks!

src/modules/battery.rs
```diff
@@ -45,34 +43,6 @@ pub struct BatteryModule {
...
-    /// ```
-    /// At battery levels below 20%,
-    /// the `.warning` class will be applied to the top-level widget.
-    /// Below 5%, `.critical` will be applied instead.
-    /// Above 20%, no class applies.
-    ///
-    /// **Default**: `{}`
-    #[serde(default)]
-    thresholds: HashMap<Box<str>, f64>,
-
```
JakeStanger: I accidentally removed this chunk & related feature code when rebasing - need to re-add

src/modules/battery.rs
```diff
@@ -247,101 +243,85 @@ impl Module<Button> for BatteryModule {
...
-    }
-}
-
-fn seconds_to_string(seconds: i64) -> Result<String> {
+fn seconds_to_string(seconds: i64) -> String {
     let mut time_string = String::new();
     let days = seconds / (DAY);
     if days > 0 {
-        write!(time_string, "{days}d")?;
+        time_string += &format!("{days}d");
```
JakeStanger: Was there a reason this was switched from `write!`? Using `+= format!` will alloc a new string every time instead of appending to the existing one.

JakeStanger:
I've resolved the underlying issues with the volume code in #1153 which means it now supports adjusting the volume as dragged, which resolves that problem. 

I've also rebased this onto the latest GTK 4 branch, and updated a lot of the new issues that have come out of that. Ironbar is now targeting a minimum of GTK 4.10, which brought in loads of deprecations so a lot of the work was replacing that. 

My hope is that in the next couple of days I should be able to get this completed and merged.

JakeStanger (approved):
Actually lgtm, let's do this. Dunno what's going on with CI but I know it works.

PR #1043, opened 2025-06-11
Migrate the tray to the gtk4 implementation

tglman:
Hi,

This include only the commit for the tray, together with https://github.com/JakeStanger/ironbar/pull/1029 should allow all the module to be successfully compiled to gtk4.

The thing to complete in this pr are:

- [ ] Connect events on the button.
- [ ] Send back events on the menu click to the app.
- [ ] Connect secondary details to menu, like shourtcuts
- [ ] Ui adjustaments

### DIFF
--- a/.github/scripts/ubuntu_setup.sh
+++ b/.github/scripts/ubuntu_setup.sh
@@ -1,4 +1,6 @@
-#!/bin/sh
+#!/usr/bin/env bash
+
+GTK4_LAYER_SHELL_VERSION="1.1.1"
 
 # sudo needed for github runner, not available by default for cross images
 if command -v sudo >/dev/null 2>&1; then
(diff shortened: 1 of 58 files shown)


tglman:
Hi,

I've done almost everything, so this would complete the migration to gtk4, there are a few thing I'm not actually sure they work because I didn't have systray menu that had them, which are:

- menu with icons
- menu with radio buttons
- shortcuts

all the rest seems to work pretty ok!

JakeStanger:
Awesome work, thanks so much! The first I can test, the latter two I can't think of anything that has those off the top of my head.

Give me a ping when you're ready for review.

JakeStanger:
Just tested and this is really nice! Way better than the menus, I really like it. Thanks for your hard work on this. 

There are definitely some bugs, but I think they're long-standing issues with the `system-tray` client implementation rather than with the UI. VLC for example doesn't seem to play very nice.

PR #1046, opened 2025-06-14
fix(nix): default HM module `config` option to `null`

postsolar:
This makes it more semantically approachable IMO.

I would expect this behavior:
- Default value is `null`
- Config is created if the value is not `null`
- Option unset (i.e. the default) => no config created
- This matches a lot of other HM modules

Currently it's this behavior (makes little sense to me):
- Default value is `{}`
- Config is created if the values is not `""`
- Need to explicitly set the option to `""` in order to not create a config

This would be a breaking change for:
- People who were setting the option to `""` to achieve not creating a config
- (Contrived) People who needed some sort of a nonsense config like `{}`

### DIFF
--- a/nix/module.nix
+++ b/nix/module.nix
@@ -36,7 +36,7 @@ in {
 
     config = mkOption {
       type = jsonFormat.type;
-      default = {};
+      default = null;
       description = "The config to pass to ironbar.";
     };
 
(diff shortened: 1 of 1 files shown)


postsolar:
I need no config to be created in `$xdg_config_home/ironbar/config.json`, because I create 2 configs separately for Hyprland and Niri and pass them via systemd unit environment.

JakeStanger (approved):
Agreed this makes more sense, thanks!

Issue #1011, opened 2025-05-26
Truncate (or rotate) player metadata in volume and music popups


postsolar:
**Is your feature request related to a problem? Please describe.**
> A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]

It often gets ridiculously long. This one is more than half my monitor width.

![Image](https://github.com/user-attachments/assets/42d0005c-a95f-4be3-a649-342e981a4a82)

I'd like it to have a consistent width.

**Describe the solution you'd like**
> A clear and concise description of what you want to happen.
> The more info here about what you are trying to achieve, the better - there's likely more than one way to go about implementing a solution.

Ideally:
- Truncate at selected direction, for selected length, with a selected string
- Optionally, also rotate

**Describe alternatives you've considered**
> A clear and concise description of any alternative solutions or features you've considered.

**Additional context**
> Add any other context or screenshots about the feature request here.

JakeStanger:
Yeah makes sense. Reckon this can be done by copying the existing `truncate` and `layout` options for the header. Not sure what I'd call the options off the top of my head...

postsolar:
To be clear by rotation I mean continuously showing a shifted slice of a string, e.g.
```
Frame 1: some long s…
Frame 2: …ome long s…
Frame 3: …me long st…
Frame 4: …e long str…
Frame 5: … long stri…
```
And so on.

postsolar:
It did sound easy until I started thinking about how exactly to move between frames. I mean it shouldn't just redraw the string every, say, 0.3 seconds, it should be smooth. Is there something like pixel-perfect placement of text that would also be safe from user's CSS intervention? 😄

It seems that GTK2 [has exactly this](https://metacpan.org/pod/Gtk2::Ex::TickerView), but for now I'm struggling to find a GTK3 equivalent.

Another idea I have is a horizontal scrollbox that would be sent scroll events in a loop.

JakeStanger:
We *could* render to a Cairo surface and translate the image but that might get nasty.

postsolar:
So I tested it with Gtk4 but it has a number of issues:
- No way to prevent the animation if the text doesn't need to be scrolled (if it's short enough)
- Some performance issue, 1 of my cores spikes to 100% usage, animation itself and also GTK debugger become jittery
- Gtk4 CSS doesn't allow specifying percentage for `transform: translateX(-100%)`

Probably not the way to go.

PR #1047, opened 2025-06-14
feat(volume): Add `truncate` option for `volume` popup

postsolar:
Part of #1011

As noted in the issue, I'm still down for marquee/rotation effect, but truncation would be an easy start.


JakeStanger:
Cheers

PR #1060, opened 2025-06-21
Add a launch_command setting

### LINKED ISSUE #1059: uwsm integration for menu/launcher modules
Not really a big problem but I'm wondering if it'd be possible to have integration with [uwsm](https://wiki.hypr.land/Useful-Utilities/Systemd-start/#launching-applications-inside-session) for the menu/launcher modules
It allows apps to be launched in separate systemd scopes via `uwsm app -- command` (or a .desktop file instead of a command)

Ironbar is great btw, it's been waaaay easier to configure than waybar/eww 👌

Username404-59:
Resolves #1059 

(feel free to roast the code, this is the first rust I've written)


Username404-59:
image
seems to have worked for my uwsm use case

Username404-59:
I think the CI should build now unless there's other errors that were just warnings locally

src/modules/launcher/mod.rs
```diff
@@ -158,6 +164,25 @@ fn default_icon_page_forward() -> String {
     String::from("󰅂")
 }
 
+pub(crate) fn default_launch_command() -> String { String::from("gtk-launch") }
```
JakeStanger: This should be across multiple lines with a gap after it. If you run `cargo fmt` it'll format all the source code correctly for you

src/modules/menu/config.rs
```diff
@@ -1,5 +1,6 @@
 use crate::config::{CommonConfig, TruncateMode};
 use crate::modules::menu::{MenuEntry, XdgSection};
+use crate::modules::launcher::default_launch_command;
```
JakeStanger: The build is failing here because you're importing from the `launcher` module into the `menu` module. Each module needs to be separate, because they're toggle-able by separate feature flags.

I reckon the best thing to do for now is to add it to the end of `config/mod.rs` (after `default_true` and `default_false`) and import it from there.

src/modules/launcher/mod.rs
```diff
@@ -158,6 +164,25 @@ fn default_icon_page_forward() -> String {
...
+pub fn launch_command(file_name: &String, str: &String) {
+    let launch_command_parts: Vec<&str> = str.split_whitespace().collect();
+    if let Err(err) = Command::new(&launch_command_parts[0])
+        .args(&launch_command_parts[1..])
+        .arg(file_name)
```
JakeStanger: What would give us even more flexibility here is the ability to put the filename in an arbitrary location in the command, not just at the end.

If we define a placeholder token for that (eg `{app_name}`) you can then do a replace for the real thing before `split_whitespace`, and pass all the args in one go.

src/modules/menu/ui.rs
```diff
@@ -17,6 +16,7 @@ pub fn make_entry<R>(
     tx: mpsc::Sender<ModuleUpdateEvent<R>>,
     image_provider: &image::Provider,
     truncate_mode: TruncateMode,
+    launch_command_str: &String,
```
JakeStanger: Passing a `&String` is considered bad practise, because under the hood a `String` type is already behind a pointer, so this requires a double lookup.

Instead you should use `&str` and the compiler will automatically coerce the types for you.

src/modules/launcher/mod.rs
Username404-59: Done I think

src/config/mod.rs
```diff
@@ -435,3 +437,25 @@ pub const fn default_false() -> bool {
 pub const fn default_true() -> bool {
     true
 }
+
+pub fn default_launch_command() -> String {
+    String::from("gtk-launch {app_name}")
+}
+
+pub fn launch_command(file_name: &str, str: &str) {
```
JakeStanger: I don't think this belongs in the config module, and I'm not sure on the name. 

There's not really a good home for it at the moment, which probably suggests I need to do some more refactoring, but for now I reckon it can go somewhere in `desktop_file.rs`. I'd also maybe rename it to something like `open_program` which better describes its purpose.

JakeStanger (changes requested):
Code lgtm, just one small thing from an organisation standpoint then I'm happy to merge. Cheers

src/desktop_file.rs
```diff
@@ -323,6 +324,25 @@ fn files(dir: &Path) -> Vec<PathBuf> {
         .collect()
 }
 
+/// Starts a `.application` file with the provided formatted command
```
JakeStanger:
```suggestion
/// Starts a `.desktop` file with the provided formatted command
```

Username404-59: oops

src/modules/launcher/mod.rs
```diff
@@ -2,14 +2,18 @@ mod item;
 mod open_state;
 mod pagination;
 
+#[path = "../../desktop_file.rs"]
+mod desktop_file;
```
JakeStanger: This shouldn't be needed. `mod` brings files into the module tree, almost always at the same level. Once they're there, they don't need to be brought in again and you can just use `use` to import from them.

Username404-59: I couldn't compile without it

JakeStanger: I can't see any reason why that'd be. If you remove it and still get an error and you're not sure feel free to share and I'll have a look.

Username404-59: > I can't see any reason why that'd be. If you remove it and still get an error and you're not sure feel free to share and I'll have a look.

Yep didn't get an error this time
I had added this because, somehow, at some point, rust gave me errors about not finding crate::desktop_file which was really weird, maybe a compiler bug or something idk.
Anyways now it seems to work without

Username404-59:
not a big deal but I just realized idiot me forgot to update https://github.com/JakeStanger/ironbar/blob/bb7e9273092de44846d0b702980ce25dc2ae2e93/src/modules/launcher/mod.rs#L115 and https://github.com/JakeStanger/ironbar/blob/bb7e9273092de44846d0b702980ce25dc2ae2e93/src/modules/menu/config.rs#L130

JakeStanger:
Oh whoops I should have spotted that. That's fine, that can get amended whenever

PR #1066, opened 2025-06-24
feat(volume): add marquee effect to popup

postsolar:
Part of #1011 

- [x] add minimal working scrolling
- [x] implement easing
- [x] implement pause on hover
- [x] implement configuration

### DIFF
--- a/src/gtk_helpers.rs
+++ b/src/gtk_helpers.rs
@@ -149,6 +149,10 @@ pub fn create_marquee_widget(
         .vscrollbar_policy(gtk::PolicyType::Never)
         .build();
 
+    if let Some(hbar) = scrolled.hscrollbar() {
+        hbar.hide();
+    }
+
     // Set `min-width` to the pixel width of the text, but not wider than `max_length` (as calculated)
     if let Some(max_length) = max_length {
         let sample_string = text.chars().take(max_length as usize).collect::<String>();


postsolar:
Draft but tips/guidance is quite welcome if there is any

postsolar:
Functionality-wise it's basically ready and works really well for me, I'll just try to polish the code a bit more before the review.

Edit:
Or not really ready. Hiding the scrollbar via traditional methods\* makes the animation slow down on mouse events for some reason. Still stuck on this.

\* "traditional methods" being:
<details><summary>PolicyType::External</summary>

```rust
    let scrolled = ScrolledWindow::builder()
        .vscrollbar_policy(gtk::PolicyType::Never)
        .hscrollbar_policy(gtk::PolicyType::External)
        .build();
```

</details>

<details><summary>.hide()</summary>

```rust
    if let Some(hbar) = scrolled.hscrollbar() {
        hbar.hide(); // or hbar.set_visible(false)
    }
```

This in particular also generally breaks scrolling when GTK phocus theme is used.

</details>

JakeStanger:
Just a thought if you're still struggling with this. A radically different approach, but it would probably be relatively trivial to use Cairo and Pango for this. You could draw text onto a "label" canvas and just translate the text coords.

postsolar:
Thank you for the suggestion. I'll try this approach next, though definitely not anytime soon as I'll be busy for a while, so if anyone wants to pick this up that'd be welcome.

Alternatively, we could just wait for the GTK4 port to be completed, and maybe GTK4 would have a more sensible behavior wrt to that one last issue of hiding the scrollbar in a theming-independent way and without causing slowdown on mouse movement.

postsolar:
Superseded by #1250.

GTK4 actually fixes that scrollbars bug, so pretty much no changes aside from updating to GTK4 and expanding what's configurable.

PR #1067, opened 2025-06-24
fix(ipc): message size limited to 1024 bytes

### LINKED ISSUE #1065: `ironbar var set`: accept bigger values

1. Use this config:
```json
{
  "start": [
    {
      "type": "label",
      "label": "#foo"
    }
  ]
}
```
2. Run `ironbar var set foo "$(printf '%965s' a)"`, see `ok`
3. Run `ironbar var set foo "$(printf '%966s' a)"`, see
```
2025-06-24T01:52:17.432655Z ERROR ironbar: 107: Connection reset by peer (os error 104)
```

I understand this to mean I can't pass an argument this long for setting an Ironvar. I'm not well-versed in lower-level Linux details, from some googling it seems like it should be fine but for some reason it stops working after the 966th byte. I don't know if it's actually related to argument size, so don't mind the feature request labeling too much.

967 bytes isn't that much, I want a label (within a popup) with more than that.

If it is in fact related to argument size, then it would be nice to have any of:
- `command-with-long-output | ironbar var set foo`
- `command-with-long-output | ironbar var set foo -`
- `command-with-long-output | ironbar var set foo --from=/dev/stdin`

Using `{{ script }}` (necessarily in `poll` mode, as `watch` would only show the last line), but I'd prefer a different architecture for my use case.

I have a long-running service that does a lot of things related to Hyprland, among them setting some Ironvars. It's simple and centralized to this service, feels nicer and easier to manage this way.

**Additional context**

JakeStanger:
Fixes #1065

### DIFF
--- a/docs/Controlling Ironbar.md
+++ b/docs/Controlling Ironbar.md
@@ -35,12 +35,10 @@ All error responses will cause the CLI to exit code 3.
 The server listens on a Unix socket. 
 The path is printed on startup, and can usually be found at `/run/user/$UID/ironbar-ipc.sock`.
 
-Commands and responses are sent as JSON objects.
+Commands and responses are sent as JSON objects. 
+The JSON should be minified and must NOT contain any `\n` characters.
 
 Commands will have a `command` key, and a `subcommand` key when part of a sub-command.
-
-The message buffer is currently limited to `1024` bytes. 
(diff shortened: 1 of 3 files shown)


JakeStanger:
There's something up with this where it blocks after a minute or so. Holding off until I can rework

Issue #487, opened 2024-03-04
[Launcher] Better mapping between favourites, executables and window classes


JakeStanger:
**Is your feature request related to a problem? Please describe.**

The launcher `favorites` option accepts any value that `xdg-open` takes, but for grouping to work this value has to match that of the window app_id / class.

**Describe the solution you'd like**

Either the `favorites` or grouping logic (or both) should be smarter to make it easier to configure. Some well-knowns should be handled without requiring user configuration.

**Describe alternatives you've considered**

None yet.

**Additional context**

Discussed in #465.

Properties in the `.desktop` file such as `StartupWMClass` may potentially lend a hand.

slowsage:
Two thoughts:

1. How compatible is gtk-launch with wayland? Can we use that as a guide? `gtk-launch code` works vs `code-url-handler` in ironbar favorites.
2. Apps using older version of electron do not set class properly (eg: Spotify). Maybe this sort of mapping can be set in the config versus requiring users to modify desktop files with StartupWMClass. Corn config allows mixed types. So maybe something like:
```
    $launcher = {
        type = "launcher"
        favorites = ["kitty" "code" {name: "spotify", title: "Spotify Premium", class: ""} ]
    }
```

When an entry is not a string:

- name corresponds the part before .desktop
- title - will match and track the window with that title exposed through wayland.
- class - same as title

`title` and `class` are optional. When only one is specified, the launcher entry is matched with that value for `title` or `class`. When both are specified, it should match both (AND).

One thing I am unclear on is `initialTitle` and `initialClass` and how that can be mapped/tracked.

JakeStanger:
Ironbar currently uses `gtk-launch` to start favourites. This is probably why it's relatively # smart in its current state.

Spotify is CEF rather than Electron, not that it matters too much. I'll have a look at that one in particular because there should be a way of identifying it properly without needing to parse a desktop file.

I'd consider reading the desktop files a last resort. That said, there is already code for doing exactly that to pull out icons. Some tweaks there if necessary would give us a service + cache for any additional properties we would need. 

I generally stay clear of mixed types where I can, but I think you're right to suggest it here. It may be that this is a place to introduce some pattern matching.

slowsage:
Re: Spotify see [this](https://community.spotify.com/t5/Desktop-Linux/No-window-icon-in-Wayland-because-app-id-is-not-correctly-set/td-p/5626427).
+1 on the pattern matching - regexes to track windows would be great.

[...]

PR #1093, opened 2025-07-19
fix(launcher): improve app ID handling: fallback for desktop entries.…

slowsage:
… Fixes: #487

This approach has decent enough defaults and we will not need to change the config parsing as I initially suggested (#487). A mapping can be forced by setting StartupWMClass in the .desktop file.

The app_id found in window/desktop environment will first be matched against the string specified in the favorites list (`items`, initially). Failing that, it will try to reverse lookup the app_id as a startup_wm_class. the reverse lookup dict `wm_app_ids` is a map from startup_wm_class to app_id that the app was initialized with.

As long as the same app is launched with the same wm_class, duplicates are handled for both favorites (secondary lookup to find key as it was specified in config file) and non-favorites (app_id/wm_class wont be in favorites, so the app_id passed in will be used as key).

Want to let it run for a bit but figured I'd send what I have for feedback.

P.S. To avoid a secondary IndexMap, I looked at using [multi_map](https://docs.rs/multi-map/latest/multi_map/) but that uses Hashmap underneath - which will not preserve order.
[This](https://github.com/lun3x/multi_index_map/issues/60) is the datastructure we need eventually IMO.

### DIFF
--- a/src/modules/launcher/mod.rs
+++ b/src/modules/launcher/mod.rs
@@ -244,21 +244,43 @@ impl Module<gtk::Box> for LauncherModule {
         let tx2 = context.tx.clone();
 
         let wl = context.client::<wayland::Client>();
+        let desktop_files = context.ironbar.desktop_files();
         spawn(async move {
             let items = items2;
             let tx = tx2;
+
+            // Build app_id mapping once at startup
+            let mut app_id_map = IndexMap::<String, String>::new();
+            {
+                let favorites: Vec<_> = lock!(items).keys().cloned().collect();
+                for fav in favorites {
+                    if let Ok(Some(file)) = desktop_files.find(&fav).await {
+                        if let Some(wm_class) = file.startup_wm_class {
+                            app_id_map.insert(wm_class, fav);
+                        }
+                    }
+                }
+            }
+
+            let resolve_app_id = |app_id: &str| {
+                app_id_map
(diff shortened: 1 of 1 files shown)


slowsage:
As mentioned in my earlier comment, [this](https://github.com/lun3x/multi_index_map/issues/60) could greatly simplify the code - we would just add the starupwmclass as key when reading favorites from config file and that would be all thats needed.

Till then, we just need to do a check if window's app_id is in the wm_class -> launcher app_id map. If a match is found, it uses the app_id from launcher instead. If no match, it uses the window's app_id.

src/modules/launcher/mod.rs
```diff
@@ -244,21 +244,43 @@ impl Module<gtk::Box> for LauncherModule {
...
         let wl = context.client::<wayland::Client>();
+        let desktop_files = context.ironbar.desktop_files();
         spawn(async move {
             let items = items2;
             let tx = tx2;
 
+            // Build app_id mapping once at startup
+            let mut app_id_map = IndexMap::<String, String>::new();
+            {
+                let favorites: Vec<_> = lock!(items).keys().cloned().collect();
```
JakeStanger, line 255: ~~Is it necessary to clone here to get the lock? Ideally it'd only clone when inserting.~~ Scrap that, it's behind an Arc so all good

PR #1160, opened 2025-09-16
fix(config): use serde_json instead of __private.

slowsage:
This fixes cargo install --git and the build issue in #1156 (building without lock file).


JakeStanger:
Brilliant, thanks for this. It looks like there's an issue currently with some old code that needs cleaning up. If you're able to take a look at that I'll get this merged

slowsage:
First time: Used claude to help with this one (learned about visitor pattern) - else would have to add serde_json/ serde_value dependency. Not sure what your ai policy is but figured I'd mention.

PR #1180, opened 2025-09-29
fix(flake): correct apps.default to point to ironbar

GavBog:
Fixes a broken reference in the apps output of the flake.
Previously, `default = self.apps.ironbar;` caused evaluation errors when running `nix run .` because self.apps.ironbar didn’t exist yet inside its own definition.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -72,12 +72,14 @@
     });
 
     # Apps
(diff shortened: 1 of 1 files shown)


JakeStanger (approved):
Thanks!

PR #1190, opened 2025-10-07
feat(inhibit): Add inhibit module

### LINKED ISSUE #929: Idle Inhibitor module
https://wayland.app/protocols/idle-inhibit-unstable-v1

> More detail to be added

### Discussed in https://github.com/JakeStanger/ironbar/discussions/928

<div type='discussions-op-text'>

<sup>Originally posted by **ildar** April  8, 2025</sup>
Couldn't find any info if there is the Idle Inhibitor in the ironbar. Could it be implemented?

ref: https://github.com/Alexays/Waybar/wiki/Module:-Idle-Inhibitor</div>

slowsage:
Resolves #929.
Sample config:
```corn
{
      type = "inhibit"
      backend = "systemd"
      durations = ["30m" "1h" "1h30m" "*2h" "inf"]
      on_click_left = "toggle"
      on_click_right = "cycle"
      format_on = "☕ {duration}"
      format_off = "💤 {duration}"
 }
```
Notes:
*2h => * implies 2h is picked as default.
Can pick systemd or wayland backend.
systemd persists across ironbar restarts (launches a transient service that sleeps for desired time).
wayland surface dies with ironbar process.
Inhibit.md included.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ default = [
     "custom",
     "focused",
     "http",
+    "inhibit",
     "ipc",
     "keyboard+all",
     "launcher",
@@ -76,6 +77,8 @@ clock = ["chrono"]
 custom = []
 
 focused = []
+
+inhibit = ["chrono"]
 
 keyboard = ["dep:colpetto", "dep:evdev-rs", "dep:rustix", "futures-lite"]
 "keyboard+all" = ["keyboard", "keyboard+sway", "keyboard+hyprland"]
@@ -165,9 +168,6 @@ reqwest = { version = "0.12.24", default-features = false, features = ["default-
 lua-src = { version = "548.1.2", optional = true }
 mlua = { version = "0.11.4", optional = true, features = ["luajit", "send"] }
 
-# clock
-chrono = { version = "0.4.42", optional = true, default-features = false, features = ["clock", "unstable-locales"] }
-
 # keyboard
 colpetto = { version = "0.7.0", features = ["tokio", "tracing"], optional = true }
 evdev-rs = { version = "0.6.3", optional = true }
@@ -192,6 +192,7 @@ libpulse-binding = { version = "2.30.1", optional = true }
 futures-lite = { version = "2.6.1", optional = true } # battery, network_manager, workspaces, keyboard
 zbus = { version = "5.12.0", default-features = false, features = ["tokio"], optional = true } # battery, network_manager, notifications
 swayipc-async = { version = "3.0.0", optional = true } # workspaces, keyboard
+chrono = { version = "0.4.42", optional = true, default-features = false, features = ["clock", "unstable-locales"] } # clock, inhibit
 hyprland = { version = "0.4.0-beta.3", optional = true } # workspaces, keyboard
 rustix = { version = "1.1.2", default-features = false, features = ["std", "fs", "pipe", "event"], optional = true } # clipboard, input
 serde_json = { version = "1.0.145", optional = true } # ipc, niri
--- a/docs/modules/Inhibit.md
+++ b/docs/modules/Inhibit.md
@@ -1,0 +1,116 @@
+Prevents the compositor from auto-suspending the system or auto-locking the screen. Click to toggle inhibit on/off or cycle through preset durations.
+
+> [!NOTE]
+> Inhibit is tracked globally (system-wide).
+> Multiple inhibit modules share the same inhibit state.
+
+## Configuration
+
+> Type: `inhibit`
+
+| Name               | Type                    | Default         | Description                                                     |
+| ------------------ | ----------------------- | --------------- | --------------------------------------------------------------- |
+| `durations`        | `string[]`              | See below       | List of durations to cycle through. See duration format.        |
+| `default_duration` | `string`                | `00:30:00`      | Starting duration. Defaults to first in `durations`.            |
+| `on_click_left`    | `'toggle'` or `'cycle'` | `'toggle'`      | Action on left click.                                           |
+| `on_click_right`   | `'toggle'` or `'cycle'` | `'cycle'`       | Action on right click.                                          |
+| `on_click_middle`  | `'toggle'` or `'cycle'` | `null`          | Action on middle click.                                         |
+| `format_on`        | `string`                | `☕ {duration}` | Format string when inhibit is active. Pango markup supported.   |
+| `format_off`       | `string`                | `💤 {duration}` | Format string when inhibit is inactive. Pango markup supported. |
+
+> **Default durations:** `["00:30:00", "01:00:00", "01:30:00", "02:00:00", "inf"]`
+>
+> **Duration format:** Time format `HH:MM:SS` (e.g., `01:30:00` for 1 hour 30 minutes). Use `0`, `00:00:00`, or `inf` for infinite duration.
+
+### Click Actions
+
+- **`toggle`**: Toggles inhibit on/off using the selected duration
+- **`cycle`**: Cycles to the next duration in the list (and applies it if already active)
+
+<details>
+<summary>JSON</summary>
+
+```json
+{
+  "end": [
+    {
+      "type": "inhibit",
+      "durations": ["00:30:00", "01:00:00", "01:30:00", "02:00:00", "inf"],
+      "default_duration": "00:30:00",
+      "on_click_left": "toggle",
+      "on_click_right": "cycle",
+      "format_on": "☕ {duration}",
+      "format_off": "💤 {duration}"
+    }
+  ]
+}
+```
+
+</details>
+
+<details>
+<summary>TOML</summary>
+
+```toml
+[[end]]
+type = "inhibit"
+durations = ["00:30:00", "01:00:00", "01:30:00", "02:00:00", "inf"]
+default_duration = "00:30:00"
+on_click_left = "toggle"
+on_click_right = "cycle"
+format_on = "☕ {duration}"
+format_off = "💤 {duration}"
+```
+
+</details>
+
+<details>
+<summary>YAML</summary>
+
+```yaml
+end:
+  - type: "inhibit"
+    durations: ["00:30:00", "01:00:00", "01:30:00", "02:00:00", "inf"]
+    default_duration: "00:30:00"
+    on_click_left: "toggle"
+    on_click_right: "cycle"
+    format_on: "☕ {duration}"
+    format_off: "💤 {duration}"
+```
+
+</details>
+
+<details>
+<summary>Corn</summary>
+
+```corn
+{
+  end = [
+    {
+      type = "inhibit"
+      durations = ["00:30:00" "01:00:00" "01:30:00" "02:00:00" "inf"]
+      default_duration = "00:30:00"
+      on_click_left = "toggle"
+      on_click_right = "cycle"
+      format_on = "☕ {duration}"
+      format_off = "💤 {duration}"
+    }
+  ]
+}
+```
+
+</details>
+
+### Formatting Tokens
+
+| Token        | Description       |
+| ------------ | ----------------- |
+| `{duration}` | Current duration. |
+
+## Styling
+
+| Selector   | Description           |
+| ---------- | --------------------- |
+| `.inhibit` | Inhibit widget button |
+
+For more information on styling, please see the [styling guide](styling-guide).
--- a/src/clients/inhibit.rs
+++ b/src/clients/inhibit.rs
@@ -1,0 +1,68 @@
+use crate::{lock, register_client};
+use gtk::ApplicationInhibitFlags;
+use gtk::prelude::*;
+use std::sync::{Arc, Mutex, Weak};
+use tracing::{error, trace};
+
+fn get_app() -> gtk::Application {
+    gtk::gio::Application::default()
+        .and_downcast()
+        .expect("GTK application not initialized")
+}
+
+/// Cookie holder that uninhibits on drop
+pub struct InhibitCookie(pub u32);
+
+impl Drop for InhibitCookie {
+    fn drop(&mut self) {
+        get_app().uninhibit(self.0);
+    }
+}
+
+/// Client for managing GTK application inhibit state.
+///
+/// Uses Weak reference to delegate ownership to modules - when all modules
+/// drop their `InhibitCookie`, the system uninhibit is called automatically.
+#[derive(Debug)]
+pub struct Client {
+    cookie: Mutex<Option<Weak<InhibitCookie>>>,
+}
+
+impl Client {
+    pub(crate) fn new() -> Self {
+        trace!("Initializing inhibit client");
+        Self {
+            cookie: Mutex::new(None),
+        }
+    }
+
+    /// Acquires an inhibit cookie.
+    ///
+    /// Uses global inhibition (no window-specific behavior).
+    /// Reuses existing cookie when multiple modules request inhibit.
+    /// Returns None if platform doesn't support inhibit.
+    pub fn acquire(&self) -> Option<Arc<InhibitCookie>> {
+        let mut cookie_opt = lock!(self.cookie);
+
+        // Upgrade weak cookie ref if it exists else create a new one
+        cookie_opt.as_ref().and_then(Weak::upgrade).or_else(|| {
+            let cookie = get_app().inhibit(
+                None::<&gtk::Window>,
+                ApplicationInhibitFlags::IDLE,
+                Some("Ironbar inhibit"),
+            );
+
+            if cookie == 0 {
+                error!("GTK inhibit failed - platform may not support it");
+                return None;
+            }
+
+            trace!("Created inhibit cookie: {}", cookie);
+            let rc = Arc::new(InhibitCookie(cookie));
+            *cookie_opt = Some(Arc::downgrade(&rc));
+            Some(rc)
+        })
+    }
+}
+
+register_client!(Client, inhibit);
--- a/src/modules/inhibit/config.rs
+++ b/src/modules/inhibit/config.rs
@@ -1,0 +1,160 @@
+use crate::config::{CommonConfig, LayoutConfig};
+use chrono::Timelike;
+use serde::de::Error;
+use serde::{Deserialize, Deserializer};
+use std::time::Duration;
+
+/// Command to control inhibit state.
+///
+/// **Valid options**: `toggle`, `cycle`
+#[derive(Debug, Clone, Copy, Deserialize)]
+#[cfg_attr(feature = "extras", derive(schemars::JsonSchema))]
+#[serde(rename_all = "lowercase")]
+pub enum InhibitCommand {
+    /// Toggle inhibit on/off
+    Toggle,
+    /// Cycle to next duration
+    Cycle,
+}
+
+#[derive(Debug, Deserialize, Clone)]
+#[cfg_attr(feature = "extras", derive(schemars::JsonSchema))]
+#[serde(default)]
+pub struct InhibitModule {
+    /// Duration configuration: list of durations to cycle through and default.
+    /// Use `0` for infinite inhibit. Format: `HH:MM:SS` (e.g., `01:30:00`)
+    ///
+    /// If `default_duration` is not specified, the first item in `durations` is used.
+    ///
+    /// **Default**: `durations = ["00:30:00", "01:00:00", "01:30:00", "02:00:00", "0"]`
+    /// **Default**: `default_duration = "00:30:00"`
+    #[serde(flatten)]
+    pub(super) duration_spec: DurationSpec,
+
+    /// Command to execute on left click.
+    ///
+    /// **Valid options**: `toggle`, `cycle`
+    /// <br>
+    /// **Default**: `toggle`
+    pub(super) on_click_left: Option<InhibitCommand>,
+
+    /// Command to execute on right click.
+    ///
+    /// **Valid options**: `toggle`, `cycle`
+    /// <br>
+    /// **Default**: `cycle`
+    pub(super) on_click_right: Option<InhibitCommand>,
+
+    /// Command to execute on middle click.
+    ///
+    /// **Valid options**: `toggle`, `cycle`
+    /// <br>
+    /// **Default**: `null`
+    pub(super) on_click_middle: Option<InhibitCommand>,
+
+    /// Format string when inhibit is active.
+    /// `{duration}` token shows remaining/selected time.
+    ///
+    /// **Default**: `"☕ {duration}"`
+    pub(super) format_on: String,
+
+    /// Format string when inhibit is inactive.
+    /// `{duration}` token shows selected duration.
+    ///
+    /// **Default**: `"💤 {duration}"`
+    pub(super) format_off: String,
+
+    /// See [layout options](module-level-options#layout).
+    #[serde(flatten)]
+    pub(super) layout: LayoutConfig,
+
+    /// See [common options](module-level-options#common-options).
+    #[serde(flatten)]
+    pub common: Option<CommonConfig>,
+}
+
+impl Default for InhibitModule {
+    fn default() -> Self {
+        Self {
+            duration_spec: DurationSpec::default(),
+            on_click_left: Some(InhibitCommand::Toggle),
+            on_click_right: Some(InhibitCommand::Cycle),
+            on_click_middle: None,
+            format_on: "☕ {duration}".to_string(),
+            format_off: "💤 {duration}".to_string(),
+            layout: LayoutConfig::default(),
+            common: Some(CommonConfig::default()),
+        }
+    }
+}
+
+fn parse_duration(s: &str) -> Result<Duration, String> {
+    if matches!(s, "0" | "00:00:00" | "inf") {
+        return Ok(Duration::MAX);
+    }
+    chrono::NaiveTime::parse_from_str(s, "%H:%M:%S")
+        .map_err(|_| format!("invalid duration: {s}"))
+        .map(|time| Duration::from_secs(time.num_seconds_from_midnight() as u64))
+}
+
+#[derive(Debug, Clone)]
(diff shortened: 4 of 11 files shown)


src/modules/inhibit/wayland.rs
JakeStanger, line 1: This partially duplicates the existing Wayland client in `clients/wayland`. It needs to be integrated with it.

src/modules/inhibit/wayland.rs
```diff
@@ -0,0 +1,88 @@
...
+
+impl WaylandBackend {
+    pub async fn new() -> Result<Self> {
+        let conn = Connection::connect_to_env()?;
+        let (globals, mut event_queue) = wayland_client::globals::registry_queue_init(&conn)?;
+        let qh = event_queue.handle();
+        let manager = globals.bind(&qh, 1..=1, ())?;
+        let compositor: wayland_client::protocol::wl_compositor::WlCompositor =
+            globals.bind(&qh, 1..=1, ())?;
+        let surface = compositor.create_surface(&qh, ());
```
JakeStanger: With GTK4 it should now be possible to get the actual window surface, rather than creating a dummy one.

https://docs.rs/gdk4-wayland/latest/gdk4_wayland/prelude/trait.WaylandSurfaceExtManual.html

src/modules/inhibit/systemd.rs
JakeStanger, line 1: This should be moved to the clients module - something like `clients/systemd_idle_inhibit.rs`

src/modules/inhibit/config.rs
```diff
@@ -0,0 +1,94 @@
...
+    pub(super) on_click_middle: Option<InhibitCommand>,
+    pub(super) format_on: String,
+    pub(super) format_off: String,
+    #[serde(flatten)]
+    pub(super) layout: LayoutConfig,
+    #[serde(flatten)]
+    pub common: Option<CommonConfig>,
+}
+
+impl Default for InhibitModule {
```
JakeStanger, line 76: This default impl isn't used anywhere. Values from here won't be merged into optional fields - you'll need to use `#[serde(default)]` for that

src/modules/inhibit/config.rs
```diff
@@ -0,0 +1,94 @@
...
+    fn default() -> Self {
+        Self {
+            backend: Some(BackendType::Systemd),
```
JakeStanger: Is there a reason for defaulting to Systemd? I would have thought Wayland would be the preferable option

src/modules/inhibit/config.rs
```diff
@@ -0,0 +1,94 @@
...
+pub enum InhibitCommand {
+    Toggle,
+    Cycle,
+}
+
+#[derive(Debug, Deserialize, Clone)]
+#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
+#[serde(default)]
+pub struct InhibitModule {
+    pub(super) backend: Option<BackendType>,
```
JakeStanger: Fields that are required by the module to work should not be an `Option` type. Generally they should have a default:

```rs
pub struct InhibitModule {
    #[serde(default)]
    pub(super) backend: Option<BackendType>,
    // ...
}

#[derive(Debug, Deserialize, Clone, Copy, PartialEq, Eq, Default)]
#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
#[serde(rename_all = "lowercase")]
pub enum BackendType {
    Systemd,
    #[default]
    Wayland,
}
```

There's a few very rare circumstances where a field should be mandatory to specify in config.

src/modules/inhibit/config.rs
```diff
@@ -0,0 +1,94 @@
...
+#[serde(rename_all = "lowercase")]
+pub enum InhibitCommand {
+    Toggle,
+    Cycle,
+}
+
+#[derive(Debug, Deserialize, Clone)]
+#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
+#[serde(default)]
+pub struct InhibitModule {
```
JakeStanger, line 23: Each config option should have doc comments detailing its usage, default value, and (for enums) valid options.

docs/modules/Inhibit.md
```diff
@@ -0,0 +1,116 @@
...
+| `on_click_right`  | `'toggle'` or `'cycle'`    | `'cycle'`       | Action on right click.                                                                      |
+| `on_click_middle` | `'toggle'` or `'cycle'`    | `null`          | Action on middle click.                                                                     |
+| `format_on`       | `string`                   | `☕ {duration}` | Format string when inhibit is active. Pango markup supported.                               |
+| `format_off`      | `string`                   | `💤 {duration}` | Format string when inhibit is inactive. Pango markup supported.                             |
+
+> **Note:** The systemd backend persists across ironbar restarts; wayland does not.
+
+**Default durations:** `["30m", "1h", "1h30m", "*2h", "inf"]`
+
+**Duration format:** Accepts human-readable durations like `30m`, `1h30m`, `2h`. Use `inf`, `infinity`, `∞`, or empty string for infinite duration. Prefix with `*` to set as default.
```
JakeStanger: I'm not sure I love this as the way of representing it in the config. I don't know what I'd do differently though - open to other ideas.

src/modules/inhibit/config.rs
```diff
@@ -0,0 +1,94 @@
...
+    }
+}
+
+fn parse_duration(s: &str) -> color_eyre::Result<(Duration, bool)> {
+    // "*" prefix marks which duration is the default selection
+    let is_default = s.trim().starts_with('*');
+    let s = s.trim().trim_start_matches('*').trim();
+    let duration = match s.to_lowercase().as_str() {
+        "inf" | "infinity" | "infinite" | "∞" | "" => Duration::MAX,
+        _ => humantime::parse_duration(s)?,
```
JakeStanger: I don't know how much value `humantime` adds vs just a `01:30:00` timestamp. I'd personally rather keep it simple here and drop the additional dep

src/modules/inhibit/mod.rs
```diff
@@ -0,0 +1,206 @@
...
+        }
+    }
+
+    async fn stop(&mut self) -> Result<()> {
+        match self {
+            Backend::Systemd(b) => b.stop().await,
+            Backend::Wayland(b) => b.stop().await,
+        }
+    }
+}
```
JakeStanger: This should be moved into the `clients` module too - you should be able to call eg `let client = clients.inhibit(Backend::Wayland)`

src/modules/inhibit/mod.rs
```diff
@@ -0,0 +1,206 @@
...
+        d => Utc::now().checked_add_signed(chrono::Duration::from_std(d).ok()?),
+    }
+}
+
+async fn create_backend(ty: BackendType) -> Result<Backend> {
+    match ty {
+        BackendType::Systemd => Ok(Backend::Systemd(systemd::SystemdBackend::new().await?)),
+        BackendType::Wayland => Ok(Backend::Wayland(wayland::WaylandBackend::new().await?)),
+    }
+}
```
JakeStanger: Should be `Backend::new`

src/modules/inhibit/mod.rs
```diff
@@ -0,0 +1,206 @@
...
+
+fn get_state(backend: &Backend, selected_duration: Duration) -> State {
+    match backend.expiry() {
+        None => State::Inactive { selected_duration },
+        Some(dt) if dt == chrono::DateTime::<Utc>::MAX_UTC => State::Active {
```
JakeStanger: What does `dt` mean here?

Cargo.toml
```diff
@@ -77,6 +78,8 @@ custom = []
 
 focused = []
 
+inhibit = ["zbus", "futures-lite", "chrono", "humantime"]
```
JakeStanger: `chrono` should be moved from the 'clock' section to 'shared'

src/modules/inhibit/mod.rs
```diff
@@ -0,0 +1,206 @@
...
+        let format_on = self.format_on;
+        let format_off = self.format_off;
+        rx.recv_glib(&label, move |label, state| {
+            let (format, duration) = match state {
+                State::Active { remaining } => (&format_on, remaining),
+                State::Inactive { selected_duration } => (&format_off, selected_duration),
+            };
+            let duration_str = match duration {
+                Duration::MAX => format!("{:>7}", "∞"),
+                d => format!("{:>7}", humantime::format_duration(d)),
```
JakeStanger: It should be reasonably straightforward to implement this without `humantime`

src/modules/inhibit/systemd.rs
```diff
@@ -0,0 +1,153 @@
...
+        let expiry = async {
+            let path = unit_path.as_ref()?;
+            let props = PropertiesProxy::builder(&conn)
+                .path(path)
+                .ok()?
+                .build()
+                .await
+                .ok()?;
+            let env_value = props
+                .get("org.freedesktop.systemd1.Service", "Environment")
```
JakeStanger: Can these strings be lifted into `const`s please

src/modules/inhibit/systemd.rs
```diff
@@ -0,0 +1,153 @@
...
+pub struct SystemdBackend {
+    conn: Connection,
+    unit_path: Option<OwnedObjectPath>,
+    pub(super) expiry: Option<DateTime<Utc>>,
+}
+
+impl SystemdBackend {
+    pub async fn new() -> Result<Self> {
+        let conn = Connection::session().await?;
+        let proxy = SystemdManagerProxy::new(&conn).await?;
```
JakeStanger: This proxy is created here, and every time `start` and `stop` are called - can we attach it to the struct?

slowsage:
These are solid points Will get back with an updated PR.
I much prefer systemd since I restart hyprland which restarts ironbar. For this reason, I wanted something persistent as default - fine with either.

JakeStanger:
My personal opinion on that:

- The Wayland protocol is already very well supported, and as Wayland is a hard requirement, it's more likely to be available (ie vs Void/Artix/Gentoo/etc users where systemd may not be present).
- As it is a protocol, it means there could be compositor-specific implementation details, which talking directly to systemd would bypass these. I've no idea what extra server-side logic you *could* tie to idle inhibit, but best to make it available by default.
- It sounds to me like your use-case is fairly niche. Most people are going to keep their compositor and bar running for as long as the machine is up.

I'd therefore say Wayland by default, Systemd available as fallback.

src/modules/inhibit/config.rs
slowsage: I'll move to field level `#[serde(default)]` as you suggest but just to clarify, Struct level `#[serde(default)]` should pull in the defaults for the missing fields and use the fields specified per [this](https://serde.rs/container-attrs.html). I was able to run it with just type: inhibit and {type: inhibit backend: wayland}so to me it seems like it is merging - unless you mean something else.

JakeStanger: Oh interesting, I had no idea about that. I also completely missed the struct-level attr. I'd say in that case your approach (implementing `Default`) is the preferred one. That also opens up an opportunity to clean up some of the config code in other modules.

slowsage:
Ended up needing a new client and removing systemd (for v1) and figured you'd have some thoughts.

After going through client code, I agree - systemd should be postponed till a proper systemd client is created - out of scope for now. Would be wasteful to use the niche systemd use-case to drive the client. systemd should be its own flag etc - want to wait to add that.
Why new gtk-wayland client?:

1. The wayland client cannot be reused - it gets the connection before gtk is loaded (by design - its meant to be used before ui - from comments). Will get super confusing to have one client for two connections.
2. I could have changed bar.rs as well to cache the surface but that seemed not ideal - did not want to change bar to accommodate inhibit.
3. The client could be called inhibit/inhibitor/etc. - open to suggestions if design looks ok. I figured each client wraps a resource so each client name maps to a resource. module names map to functionality. This client wraps the gtk enabled wayland connection [where surface id from wayland can be found], so I called it gtk_wayland.

Either way, it looks like we need a wayland conn object post gtk initialization -> implies new client. Please let me know if you have any thoughts. Not sure if its possible without perf hit to make the wayland client connection work with gtk surface id. Seems pretty core - did not want to change it.

Please let me know what you think - dont like a dedicated client for this but dont see an obvious, consistent way out.
Thank you!

Edit: High level feedback ok for now.

JakeStanger:
I'm not sure if point 1 is fully correct. The wayland client gets *initialised* at startup (before the UI), but it persists for the application lifetime, and there are modules that both send/receive events from it consistently (clipboard, focused, launcher). The architecture is a little complicated because there's effectively two layers of channels that requests/responses need to travel through (module <--> wayland client <--> wayland environment) but it should be possible.

You'd have to get the surface at request-time, rather than init-time, but that should be virtually free. 

I'd rather not add to the `Bar` if not necessary, but I'm not too strongly against point 2 if you deem it the best path forward.

As for option 3, agreed it shoul dbe named after the service/resource name rather than the module name. For Wayland this should ideally match the protocol name.

slowsage:
> I'm not sure if point 1 is fully correct. The wayland client gets _initialised_ at startup (before the UI), but it persists for the application lifetime, and there are modules that both send/receive events from it consistently (clipboard, focused, launcher). The architecture is a little complicated because there's effectively two layers of channels that requests/responses need to travel through (module <--> wayland client <--> wayland environment) but it should be possible.

I had an earlier draft that tried this but ended up with one client and two wayland connections. The issue is gtk/gk4_wayland has its own wayland connection (it should really take one as an arg but oh well).  Trying the window `surface id` from gtk/gdk4_wayland post gtk init on the old connection through existing wayland client (for inhibition) did not work. Debugging also became less clear with two connections. We could keep it as one client, one connection by having it support a phase transition from pre-gtk -> gtk connection as a backup.

> You'd have to get the surface at request-time, rather than init-time, but that should be virtually free.

I tried getting the widget (from button) at click and passing that. It still did not work with the non gtk wayland connection object.

Why I feel this is appropriate for now:
The main issue is having a separate post-gtk wayland connection in addition to existing connection is confusing. For the sake of scope creep, I think its better to have a separate diff to consolidate later.

Happy to be assigned an issue to investigate: I want to run some tests to see if gtk connection is actually slower and see what tasks exactly need running before ui - perhaps they can be queued or something else. There are a few possibilities on which direction to take based on what will be found. It could have its own set of tradeoffs, testing etc  - need to dig into timings, slowdowns etc between two clients.

Worst case, we may decide to swap out the old connection on first inhibition as I mentioned above. Did not want to change such a core part of the code when adding a module. Keeps testing much more manageable.

Lmk what you think.

slowsage:
Might be much simpler: [`gtk4::Application::inhibit()`](https://aaronerhardt.github.io/docs/relm4/gtk4/struct.Inhibit.html).
Are you familiar with this api?

JakeStanger:
Good find.

https://docs.gtk.org/gtk4/method.Application.inhibit.html

I've not used it before, but it certainly sounds like it does what we want? If so, that'd be a much cleaner implementation.

If not, am I right in thinking the need for the separate wl client stems from the request to use the bar's existing surface? If it's just that adding the complexity then using a dummy surface as before might be preferable after all.

JakeStanger:
This will still need a central client-based implemenentation btw, as otherwise it will create a seperate inhibitor per module instance

slowsage:
Its ready for review.
Monitor name is now the key to support multi monitors to hopefully inhibit screens independently based on which monitor the button is on. Code now uses smart pointer for multiple inhibits on same monitor.

Picked gtk over ModuleInfo for monitor name - to avoid passing around an extra arg.

src/modules/inhibit/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+            (MouseButton::Middle, self.on_click_middle),
+        ]
+        .into_iter()
+        .filter_map(|(btn, cmd)| cmd.map(|c| (btn, c)))
+        .for_each(|(btn, action)| {
+            let tx = tx.clone();
+            button.connect_pressed(btn, move || {
+                let tx = tx.clone();
+                spawn(async move {
+                    let _ = tx.send(action).await;
```
JakeStanger:
```suggestion
                    tx.send_expect(action).await;
```

src/modules/inhibit/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+                });
+            });
+        });
+
+        let client = ctx.client::<inhibit::Client>();
```
JakeStanger, line 116: Ideally the client should never be used from the widget code, and should only be used in the controller. This makes no functional difference but helps separate concerns and keep things cleaner, and makes it easier to refactor in future.

src/clients/inhibit.rs
```diff
@@ -0,0 +1,75 @@
...
+impl Drop for InhibitCookie {
+    fn drop(&mut self) {
+        if self.0 != 0 {
+            trace!("Uninhibiting cookie: {}", self.0);
+            get_app().uninhibit(self.0);
+        }
+    }
+}
+
+/// Client for managing GTK application inhibit state per monitor.
```
JakeStanger: Is there a particular reason or use-case why we want this to be per monitor, if the inhibit state is system-wide? To me this would seem confusing vs sharing this state across the whole application.

src/modules/inhibit/config.rs
```diff
@@ -0,0 +1,126 @@
...
+}
+
+impl Default for InhibitModule {
+    fn default() -> Self {
+        Self {
+            durations: ["00:30:00", "01:00:00", "01:30:00", "02:00:00", "0"]
+                .into_iter()
+                .map(|s| parse_duration(s).expect("default durations entry invalid"))
+                .collect(),
+            default_duration: parse_duration("02:00:00").expect("default duration invalid"),
```
JakeStanger: I think the default durations are fine, but a shorter default *default* might be preferable


```suggestion
            default_duration: parse_duration("00:30:00").expect("default duration should be valid"),
```

JakeStanger (changes requested):
Cool, getting there. I appreciate you have put a lot of work into this and there has been a lot of big changes along the way, so a big thanks for keeping at it and being so patient.

There's a few more concerns I have, most are pretty minor at this point. Once those are cleared up I reckon we'll be about ready to go.

src/clients/inhibit.rs
slowsage: Yes - this is intentional. I do not want to assume much about client environments. Given the state of apis and inconsistencies, seemed better to expose what its doing while leaving in deduplication - can make it a single instance or let each module manage later. The key issue is other system components that need to respect the state set by gtk/wayland. Since the wayland api requires surface and sleep can be at the monitor level, I aggregate at monitor level.

I think this is the safest medium between no client and singleton inhibit.

A systemd/dbus client will be the other option.

src/modules/inhibit/mod.rs
slowsage: This is intentional. It works fine outside widget code - had it outside recv_glib before but this seemed safer long term by gtk convention (app.inhibit is not guaranteed to run outside main thread - will have to do more work).

JakeStanger: Ah right of course, forgot about that. In that case could you just add a comment to make it clear that's why it's there please?

src/clients/inhibit.rs
JakeStanger: Sorry I'm not sure I'm following the logic here. The inhibit is created/managed at the GTK application level, which means it lives for as long as Ironbar runs and we don't need to think about the other potential APIs. Why not use a singleton?

slowsage:
Thanks for your patience and persistence in pushing back - had a couple of false starts before realizing:
1. gtk4's `inhibit()` is the way to go. it delegates to dbus and wayland (wayland - Not applicable for ironbar - see below).
2. wayland is a fallback after dbus fails. dbus is system wide. wayland may choose to inhibit (and docs says this is what it does) per screen.
3. wayland - `wl_surface` based inhibition is not available for gtk-layer-shell apps - `inhibit()` silently fails for non top-level apps: 
    ```rust
    // gtkapplication-wayland.c:180-192:
    
      if (flags & GTK_APPLICATION_INHIBIT_IDLE && window && impl->application == gtk_window_get_application (window))
      {
          surface = gtk_native_get_surface (GTK_NATIVE (window));
          if (GDK_IS_WAYLAND_TOPLEVEL (surface))
          {
              success = gdk_wayland_toplevel_inhibit_idle (GDK_TOPLEVEL (surface));
              if (success)
              {
                  flags &= ~GTK_APPLICATION_INHIBIT_IDLE;
                  inhibitor->surface = surface;
              }
          }
      }
    ```
    The only way to do wayland protocol based inhibition is to reuse the surface (`gdk4_wayland` crate) /create a dummy `wl_surface`. This adds a lot of code with `gdk4_wayland` or adds a invisible surface (not the cleanest solution) as you saw in my previous attempt.
6. This is the dbus flow:
gtk::Application::inhibit(None, IDLE, reason)
org.freedesktop.portal.Desktop (xdg-desktop-portal-gtk)
org.freedesktop.ScreenSaver (hypridle/swayidle/gnome-session)
7. gtk4 has two flags: IDLE and SUSPEND.
dbus IDLE is inhibit picked up by xdg-desktop-portal-gtk which then sends to the ScreenSaver dbus interface. This should inhibit suspend on hypridle, swayidle and gnome-session. However gnome-session also allows for manual inhibition (laptop lid close, etc.) when SUSPEND is passed. This is out of scope.

Main changes:
1. Cookie is now a singleton, not per-monitor.
2. Strict validation/deserialization around durations and default_duration consistent with other modules. Now, it fails when default_duration not in list.

docs/modules/Inhibit.md
```diff
@@ -0,0 +1,114 @@
...
+| `durations`        | `string[]`              | See below       | List of durations to cycle through. See duration format.        |
+| `default_duration` | `string`                | `00:30:00`      | Starting duration. Defaults to first in `durations`.            |
+| `on_click_left`    | `'toggle'` or `'cycle'` | `'toggle'`      | Action on left click.                                           |
+| `on_click_right`   | `'toggle'` or `'cycle'` | `'cycle'`       | Action on right click.                                          |
+| `on_click_middle`  | `'toggle'` or `'cycle'` | `null`          | Action on middle click.                                         |
+| `format_on`        | `string`                | `☕ {duration}` | Format string when inhibit is active. Pango markup supported.   |
+| `format_off`       | `string`                | `💤 {duration}` | Format string when inhibit is inactive. Pango markup supported. |
+
+> **Note:** Inhibit is tracked globally (system-wide). Multiple inhibit modules share the same inhibit state via reference counting.
```
JakeStanger: Can be formatted as a properly rendered note block:

```suggestion
> [!NOTE]
> Inhibit is tracked globally (system-wide). 
> Multiple inhibit modules share the same inhibit state.
```

I'd also drop the detail on reference counting - that's interesting to us but not to end users :P

Also - can this be moved to just before the configuration section please? I think it's better at the top.

docs/modules/Inhibit.md
```diff
@@ -0,0 +1,114 @@
...
+}
+```
+
+</details>
+
+### Formatting Tokens
+
+| Token        | Description                                          |
+| ------------ | ---------------------------------------------------- |
+| `{duration}` | Current duration (formatted as human-readable time). |
```
JakeStanger:
```suggestion
| `{duration}` | Current duration. |
```

src/modules/inhibit/config.rs
```diff
@@ -0,0 +1,160 @@
...
+            format_on: "☕ {duration}".to_string(),
+            format_off: "💤 {duration}".to_string(),
+            layout: LayoutConfig::default(),
+            common: Some(CommonConfig::default()),
+        }
+    }
+}
+
+fn parse_duration(s: &str) -> Result<Duration, String> {
+    if s == "0" {
```
JakeStanger: May be a good idea to support the full timestamp here too:

```suggestion
    if s == "0" || s == "00:00:00" {
```

src/modules/inhibit/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+fn format_duration(d: Duration) -> String {
+    if d == Duration::MAX {
+        return "∞".to_string();
+    }
+    let s = d.as_secs();
+    let (h, m, s) = (s / 3600, s % 3600 / 60, s % 60);
+    match (h, m) {
+        (h, m) if h > 0 => format!("{h:02}:{m:02}:{s:02}"),
+        (_, m) if m > 0 => format!("{m:02}:{s:02}"),
+        _ => format!("00:{s:02}"),
```
JakeStanger: I think these two branches equate to the same thing

JakeStanger (changes requested):
Thanks again for your hard work, implementation looks solid now. One last round of tiny changes and then I think we're good to merge.

I will be holding off merging for a couple of weeks btw, as I want to ensure the GTK4 update is solid and ship that before introducing any new major features.

src/modules/inhibit/mod.rs
```diff
@@ -0,0 +1,149 @@
...
+        ]
+        .into_iter()
+        .filter_map(|(btn, cmd)| cmd.map(|c| (btn, c)))
+        .for_each(|(btn, action)| {
+            let tx = tx.clone();
+            button.connect_pressed(btn, move || {
+                let tx = tx.clone();
+                spawn(async move {
+                    tx.send_expect(action).await;
+                });
```
JakeStanger: Sorry, missed this last time - you can use `tx.send_spawn(action)` as a shorthand here. Same below too.

JakeStanger (approved):
LGTM! This'll be first in for 0.19.

slowsage:
Since you are not going to merge for a while, will undraft after running it for a while to catch any issues - a lot of interacting pieces with this one.

slowsage:
Feel its ready. Merge and assign issues as appropriate. See [this](https://github.com/hyprwm/hypridle/pull/175) for possible hypridle bugs (that this module helped uncover).

Brisingr05:
Works properly on niri compositor with Swayidle or Hypridle.

Issue #1196, opened 2025-10-15
Duplicate Bar Appears After Disconnecting Secondary Monitor


benoitlx:
**Describe the bug**

A duplicate bar appears on my screen after I disconnect my secondary monitor.

**To reproduce**

1. Connect a secondary monitor.
2. Launch a graphical session that spawns Ironbar.
3. Disconnect the secondary monitor.
4. Observe a second bar appearing beneath the first one on the primary screen.

**Expected behavior**

Disconnecting a monitor should remove the associated bar.

**System information:**
- Distro: Fedora Silverblue
- Compositor: Niri
- Ironbar version: 0.18.0-pre (I believe the bug was introduced by the GTK4 merge, but I can pinpoint the exact commit if needed.)

**Configuration**

<details><summary>Config</summary>

```
let {
  $tray = {
    type = "tray"
  }
  $clock = {
    type = "clock"
    format = "%B %d %H:%M:%S"
    format_popup = ""
  }
  $battery = {
    type = "battery"
    format = "{percentage}%"
    icon_size = 5
  }
  $workspaces = {
    type = "workspaces"
    name_map.1 = "main"
  }
  $notifications = {
    type = "notifications"
    show_count = true
  }

  $sys_info = {
    type = "sys_info"
    
    interval.memory = 30
    interval.cpu = 1
    interval.disks = 300
    interval.temps = 3

    format = [
      "{cpu_percent}%"
      "{memory_percent}%"
      "{temp_c:CPUTIN}°C"
      "{disk_used@/var}GB"
    ]
  }

  $music = {
    type = "music"
    player_type = "mpd"
  }

  $volume = {
    type = "volume"
    format = "{icon} {percentage}%"
    max_volume = 100
  }

  $cava = {
    transition_type = "slide_end"
    transition_duration = 750
    name = "cava"
    type = "script"
    mode = "watch"
    cmd = "~/.config/ironbar/scripts/cava"
  }

  $network = {
    type = "network_manager"
  }
  $ble = {
    type = "bluetooth"
    icon_size = 32
    format.not_found = ""
    format.disabled = " Off"
    format.enabled = " On"
    format.connected = " {device_alias}"
    format.connected_battery = " {device_alias} • {device_battery_percent}%"
    popup.scrollable = true
    popup.header = " Enable Bluetooth"
    popup.disabled = "{adapter_status}"
    popup.device.header = "{device_alias}"
    popup.device.header_battery = "{device_alias}"
[...]

PR #1197, opened 2025-10-16
fix(workspaces): send workspace move for hyprland

azertyfun:
As discussed in #1196.

I am not 100 % clear on the details of why moving `rx` after `Workspaces::get()` fixes the race condition I observed – but it works and Jake mentions the whole subsystem will eventually be migrated to `ext-workspace-v1` so I'm not too worried about it.

### DIFF
--- a/src/modules/workspaces/mod.rs
+++ b/src/modules/workspaces/mod.rs
@@ -356,7 +356,7 @@ impl Module<gtk::Box> for WorkspacesModule {
 
                         has_initialized = true;
                     }
-                    WorkspaceUpdate::Add(workspace) => {
+                    WorkspaceUpdate::Add(workspace) if has_initialized => {
                         if !self.hidden.contains(&workspace.name)
                             && (self.all_monitors || workspace.monitor == output_name)
                         {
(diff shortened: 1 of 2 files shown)


JakeStanger:
Thanks for this.

The reason is because it's a broadcast channel, it's receiving all events from the Hyprland workspace client. As soon as you open it, it starts receiving, even if you're not listening.

As there was a gap where several IPC requests were performed before the init event was sent and receiver was returned to the subscribing module, this left room where other events could be sent beforehand.

I think the proposal here could still suffer from the issue - really the rx needs to be created *right* before the init event is sent.

azertyfun:
Hypothetically we could run into a concurrency issue even then (even if it's much less likely). Instead I've elected to outright ignore all events until `has_initialized` is set. This removes the possibility of a race condition altogether.

However in my test I did get a `duplicate workspace init event`. I suppose since the output went away then came back, the container was never destroyed. Either way it seems to work.

src/modules/workspaces/mod.rs
```diff
@@ -337,11 +337,8 @@ impl Module<gtk::Box> for WorkspacesModule {
             context
                 .subscribe()
                 .recv_glib((), move |(), event| match event {
-                    WorkspaceUpdate::Init(workspaces) => {
-                        if has_initialized {
-                            return;
-                        }
-
+                    WorkspaceUpdate::Init(_) if has_initialized => warn!("duplicate workspace init event"),
```
JakeStanger: Every module that joins will send a new `Init` event, which is expected. as the event is broadcast across the entire channel. This means if you have multiple bars, each with their own `workspaces` module you'll see this warning. You can remove this.

PR #1198, opened 2025-10-18
Fix race condition on monitor connection preventing bar creation

azertyfun:
As discussed in #1196

I don't have a ton of experience in Rust, so if I've implemented some anti-patterns I'd be happy to fix them if you point them out.

Cheers!

### DIFF
--- a/src/clients/outputs.rs
+++ b/src/clients/outputs.rs
@@ -4,64 +4,88 @@ use std::sync::{Arc, Mutex};
 
 use crate::channels::SyncSenderExt;
 use crate::clients::wayland;
-use crate::{Ironbar, arc_mut, debug, get_display};
+use crate::{Ironbar, arc_mut, debug, get_display, info, lock, register_client};
 use gtk::gdk::Monitor;
 use gtk::gdk::prelude::*;
 use smithay_client_toolkit::output::OutputInfo;
-use tokio::sync::broadcast;
+use tokio::sync::broadcast::{self, Sender};
 
-/// There are two ways in which we are notified of output events:
-/// 1. Through a wayland client event
-/// 2. Through a GDK event
-///
-/// We need to collate both events before we can call `load_output_bars`.
 #[derive(Debug, Clone)]
 pub enum MonitorState {
     Disconnected,
-    WaylandConnected(OutputInfo),
-    GdkConnected(glib::SendWeakRef<Monitor>),
-    BothConnected(OutputInfo, glib::SendWeakRef<Monitor>),
+    Connected(OutputInfo, glib::SendWeakRef<Monitor>),
 }
 
 #[derive(Debug, Clone)]
 pub struct MonitorEvent {
     pub connector: String,
     pub state: MonitorState,
-    tx: broadcast::Sender<MonitorEvent>,
 }
 
-impl MonitorEvent {
-    fn send(&self) {
-        debug!("Monitor event: {self:?}");
-        self.tx.send_expect(self.clone());
+/// There are two ways in which we are notified of output events:
+/// 1. Through a wayland client event
+/// 2. Through a GDK event
+///
+/// We need to collate both events before we can call `load_output_bars`.
+enum InternalMonitorState {
+    Disconnected,
+    WaylandConnected(OutputInfo),
+    GdkConnected(glib::SendWeakRef<Monitor>),
+    BothConnected(OutputInfo, glib::SendWeakRef<Monitor>),
+}
+
+struct MonitorProxy {
+    pub connector: String,
+    state: InternalMonitorState,
+}
+
+impl MonitorProxy {
+    fn disconnect(&mut self) -> &mut Self {
+        self.state = InternalMonitorState::Disconnected;
+
+        self
     }
 
-    fn disconnect(&mut self) {
-        self.state = MonitorState::Disconnected;
-
-        self.send();
-    }
-
-    fn connect_wayland(&mut self, wl_monitor: &OutputInfo) {
+    fn connect_wayland(&mut self, wl_monitor: &OutputInfo) -> &mut Self {
         self.state = match &self.state {
-            MonitorState::GdkConnected(gdk_monitor) => {
-                MonitorState::BothConnected(wl_monitor.clone(), gdk_monitor.clone())
+            InternalMonitorState::GdkConnected(gdk_monitor) => {
+                InternalMonitorState::BothConnected(wl_monitor.clone(), gdk_monitor.clone())
             }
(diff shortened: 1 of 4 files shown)


JakeStanger:
Thanks for this. From a high-level review this looks good, so no worries about the approach. Can you fix the build failures please, and I'll give a full review once they're sorted?

azertyfun:
Sorry about that, I hadn't realized that I also had to run `cargo clippy --no-default-features`.

I'd update the contributing guidelines to mention that, but what is the official stance on `cargo clippy --no-default-features` currently spewing a bunch of warnings for unused imports/code?

JakeStanger:
Yeah you're right, I'll add something to make that clearer. One of the principals of Ironbar is that all modules are feature-gated, as well as several core features. This means it needs to be build with no features, and CI will additionally check flag individually.

> but what is the official stance on cargo clippy --no-default-features currently spewing a bunch of warnings for unused imports/code?

The more important part is that it *builds*, so long as there's no warnings for the full feature set. It's nice to try and cut down on the warnings where possible, but it can often have a negative impact on code readability. The CI job allows some warnings for the no-default-features builds.

```
-A unused-imports -A unused-variables -A unused-mut -A dead-code
```

src/main.rs
```diff
@@ -464,6 +448,47 @@ fn load_output_bars(
...
+pub fn load_output_bars(ironbar: &Rc<Ironbar>, app: &Application) {
+    let wl = ironbar.clients.borrow_mut().wayland();
+    let outputs = wl.output_info_all();
+
+    let display = crate::get_display();
+    let monitors = display.monitors();
+
+    for output in outputs {
+        let Some(monitor_name) = &output.name else {
+            warn!("Output missing monitor name");
```
JakeStanger: This should be a (critical) error rather than a warning, as it's a state that we should never get into. The function should return a `Result` and bail out (return Err) if it's missing

src/outputs.rs
```diff
@@ -0,0 +1,168 @@
...
+                             * At this point, we have a `gdk::Monitor`. However, the object initially
+                             * has all its fields `None`, including `connector`.
+                             *
+                             * We have to listen for the notify event for the `connector` being set.
+                             */
+                            monitor.connect_notify(Some("connector"), move |m, _| {
+                                if let Some(connector) = m.connector() {
+                                    monitors
+                                        .lock()
+                                        .unwrap()
```
JakeStanger: `unwrap` should be avoided in production code. You can instead use the `lock!()` macro (`crate::lock`):


```suggestion
                                    lock!(monitors)
```

Under the hood this will still panic if the lock fails (this should never happen), but it uses `expect` which presents a clearer message.

src/outputs.rs
```diff
@@ -0,0 +1,168 @@
...
+    WaylandConnected(OutputInfo),
+    GdkConnected(glib::SendWeakRef<Monitor>),
+    BothConnected(OutputInfo, glib::SendWeakRef<Monitor>),
+}
+
+#[derive(Debug, Clone)]
+pub struct MonitorEvent {
+    pub connector: String,
+    pub state: MonitorState,
+    tx: broadcast::Sender<MonitorEvent>,
```
JakeStanger: This `tx` field should not be part of the event struct. Even though it's not public, it is still being unnecessarily cloned and broadcast to every consumer for every event. The `send` method should additionally be moved into the `Service` struct.

src/outputs.rs
```diff
@@ -0,0 +1,168 @@
...
+    output_channel: BroadcastChannel<MonitorEvent>,
+}
+
+impl Service {
+    pub(crate) fn new(ironbar: &Rc<Ironbar>) -> Self {
+        let output_channel = broadcast::channel(32);
+
+        let mut rx_wl_outputs = ironbar.clients.borrow_mut().wayland().subscribe_outputs();
+
+        let monitors = Arc::new(Mutex::new(HashMap::new()));
```
JakeStanger: You can use the `arc_mut!` macro here just to reduce boilerplate a tad


```suggestion
        let monitors = arc_mut!(HashMap::new());
```

src/outputs.rs
```diff
@@ -0,0 +1,168 @@
...
+
+#[derive(Debug)]
+pub struct Service {
+    // Associates a connector ID (e.g. HDMI-1) to a state
+    output_channel: BroadcastChannel<MonitorEvent>,
+}
+
+impl Service {
+    pub(crate) fn new(ironbar: &Rc<Ironbar>) -> Self {
+        let output_channel = broadcast::channel(32);
```
JakeStanger: The channel buffer here can get away with being smaller (`8` is probs sufficient), as we'd never expect to need to send an excess of connected/disconnected events at once.

src/outputs.rs
```diff
@@ -0,0 +1,168 @@
...
+    Disconnected,
+    WaylandConnected(OutputInfo),
+    GdkConnected(glib::SendWeakRef<Monitor>),
+    BothConnected(OutputInfo, glib::SendWeakRef<Monitor>),
+}
+
+#[derive(Debug, Clone)]
+pub struct MonitorEvent {
+    pub connector: String,
+    pub state: MonitorState,
```
JakeStanger: There should not ever be a case outside of this service code where we need to worry about when only one of the WL or GDK monitors exist. Rather than sending the internal state, this should ideally send its own enum with one of `Connected(OutputInfo, glib::SendWeakRef<Monitor>` or `Disconnected`.

Additionally, the events only need to be broadcast when the internal state changes to `Disconnected` or `BothConnected`.

This makes it clearer for consumers, as currently `BothConnected` is nonsensical outside of this context. It also means that a `match` covers all states, avoiding the need for a `_ => {}` branch.

src/clients/outputs.rs
JakeStanger, line 1: I think for the sake of consistency, this should be moved into `src/clients` and the `Service` class renamed to `Client`. 

A method can be placed on the `Clients` struct to grab it (as with the other clients), and the `register_client!()` macro should be used to make it easily accessible from modules.

This isn't hugely useful at the moment, but I do have some plans for the client code so good to keep things as standard as possible.

JakeStanger (changes requested):
Generally looking good. There's a few things to refactor if you can please, shouldn't be anything major. Any questions, let me know.

src/outputs.rs
azertyfun: This one is a bit tricky because the event handlers require their variables to be `move`d. I therefore can't do `self.send()` from there since `self` isn't guaranteed to live for `'static`.

I've instead added `fn maybe_send(&self, tx: &Sender<MonitorEvent>)` to `struct MonitorProxy`.

PR #1218, opened 2025-10-25
chore(nix): add nixpkgs override for naersk

TaskForceTorture:
Makes sure we are using our nixpkgs version to avoid nixpkgs duplication


JakeStanger:
Thanks!

Issue #1224, opened 2025-10-30
Nix: failed to build


TaskForceTorture:
**Describe the bug**

An error occurs when you build the flake
```error: 'wrapGAppsHook' has been renamed to/replaced by 'wrapGAppsHook3'```

**To reproduce**

1. Add the flake as an input
2. Build the flake
3. Observe build error

**Expected behavior**

Should build

**System information:**
- Distro: NixOS (nixos-unstable)
- Compositor: Niri
- Ironbar version: Trunk

TaskForceTorture:
This is due to https://github.com/NixOS/nixpkgs/commit/853d9f31ea6206e86b72db1ba221a12cb503510f

PR #1225, opened 2025-10-30
fix(nix): replace deprecated wrapGAppsHook

TaskForceTorture:
Might fix https://github.com/JakeStanger/ironbar/issues/1224


JakeStanger:
Thanks!

PR #1233, opened 2025-11-04
feat: rework networkmanager module

Rodrigodd:
- Allow a dynamic number of devices.
- Show WiFi connection strength and update in real time.
- Allow configuring the icons.
- Show network info as tool-tip.
- Allow whitelist/blacklisting device names/types.

I only have a WiFi and Wireguard interfaces, so I was  unable to test other device types, but I tried to preserve the existing functionality as much as I could.

### Screenshot

Screenshot from 2025-11-04 18-29-04

#148

### DIFF
--- a/src/modules/networkmanager.rs
+++ b/src/modules/networkmanager.rs
@@ -1,7 +1,8 @@
 mod config;
 
 use crate::channels::{AsyncSenderExt, BroadcastReceiverExt};
-use crate::clients::networkmanager::{Client, NetworkManagerUpdate};
+use crate::clients::networkmanager::state::DeviceTypeData;
+use crate::clients::networkmanager::{Client, DeviceType, NetworkManagerUpdate};
 use crate::gtk_helpers::IronbarGtkExt;
 use crate::image::Provider;
 use crate::modules::{Module, ModuleInfo, ModuleParts, WidgetContext};
@@ -14,6 +15,92 @@ use gtk::{Box as GtkBox, ContentFit, Picture};
 use tokio::sync::mpsc::Receiver;
 
 pub use config::NetworkManagerModule;
+
+use self::config::{ConnectionState, ProfileState, WifiConnectionState};
+
+impl NetworkManagerModule {
+    fn get_tooltip(&self, device: &crate::clients::networkmanager::state::Device) -> String {
+        let mut tooltip = device.interface.clone();
+        if let Some(ip) = &device.ip4_config {
+            for x in &ip.address_data {
+                tooltip.push('\n');
+                tooltip.push_str(&x.address);
+                tooltip.push('/');
+                tooltip.push_str(&x.prefix.to_string());
+            }
+        }
+        if let DeviceTypeData::Wireless(wireless) = &device.device_type_data
+            && let Some(connection) = &wireless.active_access_point
+        {
+            tooltip.push('\n');
+            tooltip.push_str(&String::from_utf8_lossy(&connection.ssid));
+        };
+
+        tooltip
+    }
+
+    fn get_profile_state(
+        &self,
+        device: &crate::clients::networkmanager::state::Device,
+    ) -> Option<ProfileState> {
+        fn whitelisted<T: PartialEq>(list: &[T], x: &T) -> bool {
+            list.is_empty() || list.contains(x)
+        }
+
+        let type_whitelisted = whitelisted(&self.types_whitelist, &device.device_type);
+        let interface_whitelisted = whitelisted(&self.interface_whitelist, &device.interface);
+        let type_blacklisted = self.types_blacklist.contains(&device.device_type);
+        let interface_blacklisted = self.interface_blacklist.contains(&device.interface);
+
+        if !type_whitelisted || !interface_whitelisted || type_blacklisted || interface_blacklisted
+        {
+            return None;
+        }
+
+        let state = ConnectionState::from(device.state);
+
+        let state = match device.device_type {
+            DeviceType::Wifi => match state {
+                ConnectionState::Acquiring => ProfileState::Wifi(WifiConnectionState::Acquiring),
+                ConnectionState::Disconnected => {
+                    ProfileState::Wifi(WifiConnectionState::Disconnected)
+                }
+                ConnectionState::Connected => match &device.device_type_data {
+                    DeviceTypeData::Wireless(wireless) => match &wireless.active_access_point {
+                        Some(connection) => ProfileState::Wifi(WifiConnectionState::Connected {
+                            signal_strength: connection.strength,
+                        }),
+                        None => ProfileState::Wifi(WifiConnectionState::Disconnected),
+                    },
+                    _ => ProfileState::Unknown,
+                },
+            },
+            DeviceType::Modem | DeviceType::Wimax => match state {
+                ConnectionState::Acquiring => ProfileState::Cellular(ConnectionState::Acquiring),
+                ConnectionState::Disconnected => {
+                    ProfileState::Cellular(ConnectionState::Disconnected)
+                }
+                ConnectionState::Connected => ProfileState::Cellular(ConnectionState::Connected),
+            },
+            DeviceType::Wireguard
+            | DeviceType::Tun
+            | DeviceType::IpTunnel
+            | DeviceType::Vxlan
+            | DeviceType::Macsec => match state {
+                ConnectionState::Acquiring => ProfileState::Vpn(ConnectionState::Acquiring),
+                ConnectionState::Disconnected => ProfileState::Vpn(ConnectionState::Disconnected),
+                ConnectionState::Connected => ProfileState::Vpn(ConnectionState::Connected),
+            },
+            _ => match state {
+                ConnectionState::Acquiring => ProfileState::Wired(ConnectionState::Acquiring),
+                ConnectionState::Disconnected => ProfileState::Wired(ConnectionState::Disconnected),
+                ConnectionState::Connected => ProfileState::Wired(ConnectionState::Connected),
+            },
+        };
+
+        Some(state)
+    }
+}
 
 impl Module<GtkBox> for NetworkManagerModule {
     type SendMessage = NetworkManagerUpdate;
--- a/src/modules/networkmanager/config.rs
+++ b/src/modules/networkmanager/config.rs
@@ -1,6 +1,5 @@
 use crate::clients::networkmanager::DeviceState;
 use crate::clients::networkmanager::DeviceType;
-use crate::clients::networkmanager::state::DeviceTypeData;
 use crate::config::{CommonConfig, Profiles, State, default};
 use crate::profiles;
 use serde::Deserialize;
@@ -36,92 +35,6 @@ pub struct NetworkManagerModule {
 
     #[serde(flatten)]
     pub common: Option<CommonConfig>,
-}
-impl NetworkManagerModule {
-    pub(super) fn get_tooltip(
-        &self,
-        device: &crate::clients::networkmanager::state::Device,
-    ) -> String {
-        let mut tooltip = device.interface.clone();
-        if let Some(ip) = &device.ip4_config {
-            for x in &ip.address_data {
-                tooltip.push('\n');
-                tooltip.push_str(&x.address);
-                tooltip.push('/');
-                tooltip.push_str(&x.prefix.to_string());
-            }
-        }
-        if let DeviceTypeData::Wireless(wireless) = &device.device_type_data
-            && let Some(connection) = &wireless.active_access_point
-        {
-            tooltip.push('\n');
-            tooltip.push_str(&String::from_utf8_lossy(&connection.ssid));
-        };
-
-        tooltip
-    }
-
-    pub(super) fn get_profile_state(
-        &self,
-        device: &crate::clients::networkmanager::state::Device,
-    ) -> Option<ProfileState> {
-        fn whitelisted<T: PartialEq>(list: &[T], x: &T) -> bool {
-            list.is_empty() || list.contains(x)
-        }
-
-        let type_whitelisted = whitelisted(&self.types_whitelist, &device.device_type);
-        let interface_whitelisted = whitelisted(&self.interface_whitelist, &device.interface);
-        let type_blacklisted = self.types_blacklist.contains(&device.device_type);
-        let interface_blacklisted = self.interface_blacklist.contains(&device.interface);
-
-        if !type_whitelisted || !interface_whitelisted || type_blacklisted || interface_blacklisted
-        {
-            return None;
-        }
-
-        let state = ConnectionState::from(device.state);
-
-        let state = match device.device_type {
-            DeviceType::Wifi => match state {
-                ConnectionState::Acquiring => ProfileState::Wifi(WifiConnectionState::Acquiring),
-                ConnectionState::Disconnected => {
-                    ProfileState::Wifi(WifiConnectionState::Disconnected)
-                }
-                ConnectionState::Connected => match &device.device_type_data {
-                    DeviceTypeData::Wireless(wireless) => match &wireless.active_access_point {
-                        Some(connection) => ProfileState::Wifi(WifiConnectionState::Connected {
-                            signal_strength: connection.strength,
-                        }),
-                        None => ProfileState::Wifi(WifiConnectionState::Disconnected),
-                    },
-                    _ => ProfileState::Unknown,
-                },
-            },
-            DeviceType::Modem | DeviceType::Wimax => match state {
-                ConnectionState::Acquiring => ProfileState::Cellular(ConnectionState::Acquiring),
-                ConnectionState::Disconnected => {
-                    ProfileState::Cellular(ConnectionState::Disconnected)
-                }
-                ConnectionState::Connected => ProfileState::Cellular(ConnectionState::Connected),
-            },
-            DeviceType::Wireguard
-            | DeviceType::Tun
-            | DeviceType::IpTunnel
-            | DeviceType::Vxlan
-            | DeviceType::Macsec => match state {
-                ConnectionState::Acquiring => ProfileState::Vpn(ConnectionState::Acquiring),
-                ConnectionState::Disconnected => ProfileState::Vpn(ConnectionState::Disconnected),
-                ConnectionState::Connected => ProfileState::Vpn(ConnectionState::Connected),
-            },
-            _ => match state {
-                ConnectionState::Acquiring => ProfileState::Wired(ConnectionState::Acquiring),
-                ConnectionState::Disconnected => ProfileState::Wired(ConnectionState::Disconnected),
-                ConnectionState::Connected => ProfileState::Wired(ConnectionState::Connected),
-            },
-        };
-
-        Some(state)
-    }
 }
 
 impl Default for NetworkManagerModule {


JakeStanger:
Does this superseed #690?

docs/modules/Networkmanager.md
JakeStanger, line 1: It looks like this doc file is new without replacing/removing the old one. The docs are also outdated, reflecting the old configuration/behavior - can this be updated please?

docs/modules/Networkmanager.md
```diff
@@ -0,0 +1,91 @@
...
+| `icon_size`                   | `integer`  | `24`                                                  | Size to render icon at.                                                                                            |
+| `icons.wired.connected`       | `string`   | `icon:network-wired-symbolic`                         | Icon for connected wired device                                                                                    |
+| `icons.wired.acquiring`       | `string`   | `icon:network-wired-acquiring-symbolic`               | Icon for acquiring wired device                                                                                    |
+| `icons.wired.disconnected`    | `string`   | `""`                                                  | Icon for disconnected wired device                                                                                 |
+| `icons.wifi.levels`           | `string[]` | `["icon:network-wireless-signal-none-symbolic", ...]` | Icon for each strengh level of a connected wifi connection, from lowest to highest. The default contains 5 levels. |
```
JakeStanger: Levels as an array may be more complex than necessary. It also adds a third approach to handling levels compared to other modules (battery, volume). A low/medium/high approach may be more accessible.

At the very least the full list of icons does need to be included in the docs. See how it's being handled in the inhibit module as example:

https://github.com/JakeStanger/ironbar/pull/1190/files#diff-20bab0fe763db2e74cc41455d0d53cdeaa00c9707c367e2c4fac3a1bde131bcfR13

docs/modules/Networkmanager.md
```diff
@@ -0,0 +1,91 @@
...
+> A solution to this is currently in the works.
+
+## Configuration
+
+> Type: `networkmanager`
+
+| Name                          | Type       | Default                                               | Description                                                                                                        |
+| ----------------------------- | ---------- | ----------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------ |
+| `icon_size`                   | `integer`  | `24`                                                  | Size to render icon at.                                                                                            |
+| `icons.wired.connected`       | `string`   | `icon:network-wired-symbolic`                         | Icon for connected wired device                                                                                    |
```
JakeStanger: Nit but you ensure all the descriptions end with a full stop `.` please for consistency.

docs/modules/Networkmanager.md
```diff
@@ -0,0 +1,91 @@
...
+
+<details>
+<summary>JSON</summary>
+
+```json
+{
+  "end": [
+    {
+      "type": "networkmanager",
+      "icon_size": 32
```
JakeStanger: For examples I prefer to include the default values.

```suggestion
      "icon_size": 24
```

src/modules/networkmanager/config.rs
```diff
@@ -0,0 +1,137 @@
...
+            fn $name() -> String {
+                ($default).to_string()
+            }
+        )*
+    };
+}
+
+#[derive(Debug, Deserialize, Clone, Default)]
+#[cfg_attr(feature = "extras", derive(schemars::JsonSchema))]
+pub struct IconsConfig {
```
JakeStanger: This can just be called `Icons`. Same with other structs - the `Config` part is redundant.

src/modules/networkmanager/config.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    #[serde(default = "default_unknown")]
+    pub unknown: String,
+}
+
+#[derive(Debug, Deserialize, Clone)]
+#[cfg_attr(feature = "extras", derive(schemars::JsonSchema))]
+pub struct IconsConfigWired {
+    #[serde(default = "default_wired_connected")]
+    pub connected: String,
```
JakeStanger: Recently discovery - you can put a `#[serde(default)]` annotation on the struct itself, and it'll merge in the default implmenetation for missing fields. This removes the need to put a `default` function on each field.

Rodrigodd:
Oh, I guess so. I didn’t notice that @Zedfrigg was already working on a reword of his network manager module implementation.

Compared with #690, my PR appears to have more features. The only area I think my PR is lacking is in how I’m passing events between the Client and the Module. My PR sends the entire list of devices whenever a device changes, and future events refer to devices by index, whereas #690 uses an ID system.

Zedfrigg:
Hi there! This seems pretty in line with the features I implemented or plan to implement in my PR. I'm away from home at the moment and don't have the opportunity to read through your code in detail until a couple of days from now, so instead I'll leave you with some of the considerations I've encountered while implementing this myself:

- Handling initialisation of more module instances (e.g. new monitor plugged in) correctly when already running
- Handling the addition and removal of network devices (e.g. usb ethernet adapters)
- Doing all of this without leaking memory or leaving unused tasks running

docs/modules/Networkmanager.md
```diff
@@ -0,0 +1,91 @@
+Displays the current network connection state of NetworkManager.
+Supports wired ethernet, wifi, cellular data and VPN connections among others.
+
+> [!NOTE]
+> This module uses NetworkManager's so-called primary connection, and therefore inherits its limitation of only being able to display the "top-level" connection.
+> For example, if we have a VPN connection over a wifi connection it will only display the former, until it is disconnected, at which point it will display the latter.
+> A solution to this is currently in the works.
```
JakeStanger: Does this still hold true?

src/clients/networkmanager/mod.rs
```diff
@@ -0,0 +1,459 @@
...
+
+type PathMap<'l, ValueType> = HashMap<ObjectPath<'l>, ValueType>;
+
+#[derive(Clone, Debug)]
+pub enum NetworkManagerUpdate {
+    /// Update all devices
+    Devices(Vec<state::Device>),
+    /// Update a single device.
+    ///
+    /// The `usize` is the index of the device in the list of devices, receive in the previous
```
JakeStanger:
```suggestion
    /// The `usize` is the index of the device in the list of devices 
    /// received in the previous `Devices` update.
```

src/clients/networkmanager/mod.rs
```diff
@@ -0,0 +1,459 @@
...
+        let dbus_connection = Connection::system().await?;
+        let root_object = {
+            let root_object = DbusProxy::new(&dbus_connection).await?;
+            // Workaround for the fact that zbus (unnecessarily) requires a static lifetime here
+            Box::leak(Box::new(root_object))
+        };
+
+        let (tx, rx) = broadcast::channel(32);
+
+        std::mem::forget(rx);
```
JakeStanger, line 69: Clever! I need to start doing this

src/clients/networkmanager/state.rs
JakeStanger, line 1: There looks like there's an issue with the field comments throughout this file where some are incorrectly below fields and so not attached to anything. It might be intentional but I can't make sense of why they'd be like that?

src/clients/networkmanager/dbus.rs
```diff
@@ -0,0 +1,206 @@
...
+)]
+pub trait Ip4ConfigDbus {
+    #[zbus(property)]
+    fn address_data(&self) -> Result<Vec<HashMap<String, OwnedValue>>>;
+}
+
+#[derive(Clone, Copy, Debug, OwnedValue, PartialEq, Deserialize)]
+#[repr(u32)]
+#[cfg_attr(feature = "extras", derive(schemars::JsonSchema))]
+pub enum DeviceType {
```
JakeStanger, line 133:
```suggestion
#[derive(Clone, Copy, Debug, OwnedValue, PartialEq, Deserialize)]
#[repr(u32)]
#[serde(rename_all = "snake_case")]
#[cfg_attr(feature = "extras", derive(schemars::JsonSchema))]
pub enum DeviceType {
```

docs/modules/Networkmanager.md
JakeStanger, line 1: There is a large number of device types. Is there a comprehensive list of these externally documented we can link to? Otherwise they will need to all be included here.

src/modules/networkmanager.rs
```diff
@@ -1,53 +1,192 @@
...
+                    },
+                    _ => self.icons.unknown.as_str(),
+                },
+            },
+            DeviceType::Modem | DeviceType::Wimax => match () {
+                _ if acquiring => self.icons.cellular.acquiring.as_ref(),
+                _ if disconnected => self.icons.cellular.disconnected.as_ref(),
+                _ if connected => self.icons.cellular.connected.as_ref(),
+                _ => self.icons.unknown.as_ref(),
+            },
```
JakeStanger: I can't tell if this is genius or cursed. I assume under the hood this just compiles down to the same thing as a `if elseif else`? Just curious.

Regardless if this is represented as an enum it should disappear. See my comment above.

src/modules/networkmanager.rs
```diff
@@ -1,53 +1,192 @@
...
+use crate::clients::networkmanager::state::DeviceTypeData;
+use crate::clients::networkmanager::{Client, DeviceState, DeviceType, NetworkManagerUpdate};
 use crate::config::{CommonConfig, default};
+use crate::gtk_helpers::IronbarGtkExt;
 use crate::modules::{Module, ModuleInfo, ModuleParts, WidgetContext};
 use crate::{module_impl, spawn};
+
 use color_eyre::Result;
-use futures_lite::StreamExt;
-use futures_signals::signal::SignalExt;
```
JakeStanger, line 8: Is `futures_signals` still in use? If not it can be removed from `cargo.toml`.

src/modules/networkmanager.rs
```diff
@@ -1,53 +1,192 @@
...
+            | DeviceState::Failed
+            | DeviceState::Disconnected => disconnected = true,
+            DeviceState::Prepare
+            | DeviceState::Config
+            | DeviceState::NeedAuth
+            | DeviceState::IpConfig
+            | DeviceState::IpCheck
+            | DeviceState::Secondaries => acquiring = true,
+            DeviceState::Activated => connected = true,
+        }
```
JakeStanger: This would be better as an enum

```rust
enum State {
  Disconnected,
  Acquiring,
  Connected
}

impl From<DeviceState> for State {
   // ...
}

File: src/modules/networkmanager.rs [hunk 13]
```diff
@@ -57,52 +196,110 @@ impl Module<GtkBox> for NetworkManagerModule {
...
+                    }
+                }
             }
         });
 
-        Ok(ModuleParts::new(container, None))
+        Ok(ModuleParts::new(container_clone, None))
     }
+
+    module_impl!("networkmanager");
```
JakeStanger: Can this be moved to the top of the impl block please

JakeStanger (changes requested):
Thanks for the hard work. 

I think we'll go with this over #690, pending a review from @Zedfrigg - I'd want to make sure we don't lose anything from there.

There's some initial things I've spotted in review, if you can have a look at those for me please.

docs/modules/Networkmanager.md
Rodrigodd: An array seemed to me to be the simplest option. GTK icon themes have five levels of wireless signal strength, and I would like to have all of them available. Do you prefer that I make a struct with five fields: none, weak, ok, good and excellent?
There’s also the option of making a map of thresholds like in the battery module, but that would be even more complex, and the current mapping is non-linear.

Please let me know which approach you prefer.

But, for now, I’ve moved the default array out of the table.

docs/modules/Networkmanager.md
Rodrigodd: No, I removed it.

src/clients/networkmanager/state.rs
Rodrigodd: I left all fields available in the DBus object as comments, [like in the docs](https://networkmanager.dev/docs/api/latest/gdbus-org.freedesktop.NetworkManager.IP4Config.html), in case they turn out to be useful in the future. I declared the ones currently used by the module as actual fields with doc-comments.

However, I think the comments above the doc-strings make them harder to read. I tried adding a newline between them, but `rustfmt` removes it. I’ve moved all such comments below the fields so they don’t get confused with the doc-comments.

I can also just remove them if you prefer.

docs/modules/Networkmanager.md
Rodrigodd: I added a link to the documentation and added a missing device type. I also added doc comments to each variant.

But I think I missed what you meant here. You meant linking the list of them in the user docs, right?

Rodrigodd: I added instructions of how to discovered the device types using `nmcli device status` to the docs, and included a list of all possible device types. I could not link to the docs directly, because they where in another format (UPPER_SNAKE_CASE with a prefix), and one of the devices is being written differently (`6LOWPAN` vs `six_lowpan`).

src/modules/networkmanager.rs
Rodrigodd: I remove it.

src/modules/networkmanager.rs
Rodrigodd: Yeah, it is basically a if-else chain, but formats more nicely, specially in nested `match`s. But yes, the enum is a better approach.

docs/modules/Network-Manager.md
```diff
@@ -1,80 +1,113 @@
...
-  ```
+<summary>JSON</summary>
+
+```json
+{
+  "end": [
+    {
+      "type": "networkmanager",
+      "icon_size": 24
+      types_blacklist: ["loopback", "bridge"]
```
JakeStanger:
```suggestion
      "types_blacklist": ["loopback", "bridge"]
```

docs/modules/Network-Manager.md
```diff
@@ -1,80 +1,113 @@
...
-  ```
+<summary>Corn</summary>
+
+```corn
+{
+  end = [
+    {
+      type = "networkmanager"
+      icon_size = 24
+      types_blacklist = ["loopback", "bridge"]
```
JakeStanger:
```suggestion
      types_blacklist = [ "loopback" "bridge" ]
```

docs/modules/Network-Manager.md
```diff
@@ -1,80 +1,113 @@
...
+**Device Types:** The device types used in `types_whitelist` and
+`types_blacklists` are the same as those used by NetworkManager. You can find
+the type of the devices on your system by running `nmcli device status` in a
+terminal. The possible device types are: `unknown`, `ethernet`, `wifi`, `bt`,
+`olpc_mesh`, `wimax`, `modem`, `infiniband`, `bond`, `vlan`, `adsl`, `bridge`,
+`generic`, `team`, `tun`, `ip_tunnel`, `macvlan`, `vxlan`, `veth`, `macsec`,
+`dummy`, `ppp`, `ovs_interface`, `ovs_port`, `ovs_bridge`, `wpan`, `six_lowpan`,
+`wireguard`, `wifi_p2p`, `vrf`, `loopback`, `hsr` and `ipvlan`.
+
+**Default `icons.wifi.levels`:** they contain the 5 GTK symbolic icons for wireless signal strength:
```
JakeStanger:
```suggestion
**Default `icons.wifi.levels`**: Contains the 5 GTK symbolic icons for wireless signal strength:
```

src/clients/networkmanager/state.rs
JakeStanger: Ah I'm with you. Makes sense to keep them. Sticking a blank comment line and a header in there should make it clearer:

```rs
//
// # Unmapped:
//  Udi                   readable   s
// ...
```

Alternatively rustfmt might preserve the newline if you use `/* block comments */` instead?

src/clients/networkmanager/mod.rs
```diff
@@ -0,0 +1,459 @@
...
+
+        Ok(Client {
+            root_object,
+            dbus_connection,
+            tx,
+            device_map: Arc::new(RwLock::new(HashMap::new())),
+            ip4config_map: Arc::new(RwLock::new(HashMap::new())),
+            access_point_map: Arc::new(RwLock::new(HashMap::new())),
```
JakeStanger: You can use `crate::arc_rw!` here as a shorthand

```suggestion
            device_map: arc_rw!(HashMap::new()),
            ip4config_map: arc_rw!(HashMap::new()),
            access_point_map: arc_rw!(HashMap::new()),
```

src/modules/networkmanager.rs
```diff
@@ -1,53 +1,207 @@
...
+                .types_whitelist
+                .iter()
+                .any(|t| t == &device.device_type)
+            || self.types_blacklist.contains(&device.device_type)
+            || !self.interface_whitelist.is_empty()
+                && !self
+                    .interface_whitelist
+                    .iter()
+                    .any(|n| n == &device.interface)
+            || self.interface_blacklist.contains(&device.interface)
```
JakeStanger: This condition is rather unweildly. I think it'd be better if each one was its own temporary binding to split it up.

```rs
let is_type_whitelisted = ... // etc
```

src/modules/networkmanager.rs
```diff
@@ -57,52 +211,115 @@ impl Module<GtkBox> for NetworkManagerModule {
...
-            async move {
-                image_provider
-                    .load_into_picture_silent(icon_name, self.icon_size, false, &icon)
-                    .await;
+                        // update each icon to match the device state
+                        for (device, widget) in devices.iter().zip(container.children()) {
+                            this.update_icon(
+                                &image_provider,
+                                device,
+                                widget.downcast_ref::<Picture>().unwrap(),
```
JakeStanger: CI should be flagging `unwrap`, I'm not sure why it isn't

```suggestion
                                widget.downcast_ref::<Picture>().expect("widget should be picture"),
```

Same below too pls

src/modules/networkmanager.rs
```diff
@@ -57,52 +211,115 @@ impl Module<GtkBox> for NetworkManagerModule {
...
+    //     5..=29 => 1,
+    //     30..=54 => 2,
+    //     55..=79 => 3,
+    //     80.. => 4,
+    // }
+
+    // to make it work with a custom number of levels, we approach the logic above with the logic
+    // below (0 for < 5, and a linear interpolation for 5 to 105).
+    // TODO: if there are more than 20 levels, the last level will be out of scale, and never be
+    // reach.
```
JakeStanger: This needs resolving prior to merge ideally please. I'd suggest just documenting only 20 levels are supported, as that should be plenty enough for any reasonable usecase (I say, naively).

JakeStanger (changes requested):

src/clients/networkmanager/mod.rs
Rodrigodd: Those are `tokio::sync::RwLock`, so I cannot use `arc_rw!` here.

src/modules/networkmanager.rs
Rodrigodd: I fixed the TODO by rewriting the implementation.

src/clients/networkmanager/mod.rs
JakeStanger: Ah, didn't spot that then. Could you just stick a comment somewhere to make it clear it's the Tokio one (or alias the type, or include the namespace, or whatever works) please?

docs/modules/Networkmanager.md
JakeStanger: > You meant linking the list of them in the user docs, right?

I did yes, although both is good :)

docs/modules/Networkmanager.md
JakeStanger: I'll get back to you on this one. #1235 is proposing yet another syntax so this needs standardising and I need to have a think about the best approach.

docs/modules/Network-Manager.md
```diff
@@ -1,80 +1,113 @@
...
-  ```
+<summary>Corn</summary>
+
+```corn
+{
+  end = [
+    {
+      type = "networkmanager"
+      icon_size = 24
+      types_blacklist = [ loopback bridge ]
```
JakeStanger: Strings in Corn need double quotes:

```suggestion
      types_blacklist = [ "loopback" "bridge" ]
```

src/modules/networkmanager.rs
```diff
@@ -1,53 +1,205 @@
...
     }
 }
 
 impl Module<GtkBox> for NetworkManagerModule {
-    type SendMessage = ClientState;
+    type SendMessage = NetworkManagerUpdate;
     type ReceiveMessage = ();
 
-    module_impl!("network_manager");
+    module_impl!("networkmanager");
```
JakeStanger: This changes the module name, which is a breaking change. Is this intentional?

src/clients/networkmanager/mod.rs
Rodrigodd: I aliased the type to `AsyncRwLock`.

docs/modules/Network-Manager.md
```diff
@@ -1,80 +1,113 @@
...
 ## Styling
 
-| Selector                 | Description                      |
-|--------------------------|----------------------------------|
-| `.network_manager`       | NetworkManager widget container. |
-| `.network_manager .icon` | NetworkManager widget icon.      |
+| Selector               | Description                      |
+| ---------------------- | -------------------------------- |
+| `.networkmanager`      | NetworkManager widget container. |
+| `.networkmanger .icon` | NetworkManager widget icons.     |
```
JakeStanger:
```suggestion
| `.network_manager`      | NetworkManager widget container. |
| `.network_manger .icon` | NetworkManager widget icons.     |
```

docs/modules/Network-Manager.md
```diff
@@ -1,80 +1,113 @@
...
 
-> Type: `network_manager`
-
-| Name        | Type      | Default | Description             |
-|-------------|-----------|---------|-------------------------|
-| `icon_size` | `integer` | `24`    | Size to render icon at. |
-
-> [!NOTE]
-> This module does not support module-level [layout options](module-level-options#layout).
+> Type: `networkmanager`
```
JakeStanger: References here need to be reverted to contain the `_` since the module name change was reverted in `module_impl!`.

```suggestion
> Type: `network_manager`
```

src/clients/networkmanager/mod.rs
```diff
@@ -0,0 +1,459 @@
...
+    }
+
+    async fn fetch_device_wireless(&self, path: &ObjectPath<'_>) -> Result<state::DeviceTypeData> {
+        let device_object = DeviceWirelessDbusProxy::new(&self.dbus_connection, path).await?;
+        let active_access_point_path = device_object.active_access_point().await?;
+        let active_access_point = if active_access_point_path.as_ref() != "/" {
+            match self.fetch_access_point(&active_access_point_path).await {
+                Ok(x) => Some(x),
+                Err(e) => {
+                    tracing::error!("fail fetch access point: {e}");
```
JakeStanger:
```suggestion
                    tracing::error!("failed to fetch access point: {e}");
```

src/clients/networkmanager/mod.rs
```diff
@@ -0,0 +1,459 @@
...
+    /// The `usize` is the index of the device in the list of devices received in the previous
+    /// `Devices` update.
+    Device(usize, state::Device),
+}
+
+#[derive(Debug)]
+struct ClientDevice {
+    state: state::Device,
+    index: usize,
+    _state_handle: tokio::task::JoinHandle<()>,
```
JakeStanger: Since these fields are read (to call `abort`) they should not be underscore-prefixed.


```suggestion
    state_handle: tokio::task::JoinHandle<()>,
```

(same with below two)

docs/modules/Networkmanager.md
JakeStanger: I've implemented the profiles system in #1349, which is documented in the following locations:

The volume and battery modules also provide implementations which can be used as reference. 

Can this be refactored to switch over to using that system for the icons please? Give me a shout if you need a hand.

You will need to define a `profiles` key with a custom profile type, which should have an `icons` key. You can move the top-level icons option into there & it'll merge up automatically.

Rodrigodd:
I refactored the code to make the icon configuration use the profile system. Now there is a single icon configuration that uses the profile system to change it based on the device type and its properties (currently only the signal strength for the WiFi connection). @JakeStanger, is this approach okay? An alternative would be to keep only the signal strength (and connection state?) in the profile state and have a field for each device icon.

One problem I have is that I could not figure out how to make the profile configuration work. Anything I try results in `Error loading config: data did not match any variant of untagged enum ProfileEntry`. Do you have any idea what I am doing wrong?

I still need to finish testing it, update the documentation, see if I can avoid blocking the profile update callback on an async call, etc., but I wanted to confirm those things first.

JakeStanger:
Thanks. I've not had a full review yet, but I think the general approach looks good.

> Do you have any idea what I am doing wrong?

The state type is quite complex. I suspect you're coming up against the default Serde enum handling. The current model would probably require something similar to this to work:

```corn
profile.example.when = { Wifi = { Connected = { signal_strength = 50 } } }
```

https://serde.rs/enum-representations.html

I'd probably look to use internal tagging (or untagged) instead.

Rodrigodd:
> Anything I try results in `Error loading config: data did not match any variant of untagged enum ProfileEntry`

The problem was that I thought I could access the data of an existing default profile. But no, the default profile is only merged in after the config is deserialized. So I need to create profiles from scratch, defining the `when` field first.

I added a commit refactoring the state representation to use internal tags, but I still need to address the pending work I mentioned in my last comment. I’ll try to work on it during the week or next weekend.

Rodrigodd:
I fix the issues I mention in my previous comment, I think this is okay for a review now.

How do you think the profiles should be documented? I just included a  list of the default profiles in corn format, and make sure to include a single profile in each example configuration so the user can see how it can be used to modify the icons.

JakeStanger:
Cool thanks will run through shortly.

> How do you think the profiles should be documented?

I have plans to expand this as part of #1320 but for now I think what you've included is more than sufficient.

src/modules/networkmanager/config.rs
```diff
@@ -0,0 +1,129 @@
...
+}
+
+#[derive(Default, Debug, Deserialize, Clone, PartialOrd, PartialEq, Eq)]
+#[cfg_attr(feature = "extras", derive(schemars::JsonSchema))]
+pub enum VpnState {
+    #[default]
+    Disconnected,
+    Acquiring,
+    Connected,
+}
```
JakeStanger: It's unlikely the non-wifi states will differ any time soon, so these can be merged. I also think `ConnectionState` would be a better name.

```suggestion
#[derive(Default, Debug, Deserialize, Clone, PartialOrd, PartialEq, Eq)]
#[cfg_attr(feature = "extras", derive(schemars::JsonSchema))]
pub enum ConnectionState {
    #[default]
    Disconnected,
    Acquiring,
    Connected,
}

#[derive(Default, Debug, Deserialize, Clone, PartialOrd, PartialEq, Eq)]
#[cfg_attr(feature = "extras", derive(schemars::JsonSchema))]
#[serde(tag = "state")]
pub enum WifiConnectionState {
    #[default]
    Disconnected,
    Acquiring,
    Connected {
        /// The signal strength of the wifi connection, from 0 to 100.
        signal_strength: u8,
    },
}
```

src/modules/networkmanager/config.rs
```diff
@@ -0,0 +1,129 @@
...
+    Wifi {
+        #[serde(flatten)]
+        state: WifiState,
+    },
+    Cellular {
+        state: CellularState,
+    },
+    Vpn {
+        state: VpnState,
+    },
```
JakeStanger: Untested, but...if you convert these to newtype variants, and use `#[serde(tag="key")]` on `ConnectionState`, I think you can simplify this:

```suggestion
    Wired(ConnectionState),
    Wifi(WifiState),
    Cellular(ConnectionState),
    Vpn(ConnectionState),
```

src/modules/networkmanager/config.rs
```diff
@@ -0,0 +1,129 @@
...
+#[derive(Default, Debug, Deserialize, Clone, PartialOrd, PartialEq, Eq)]
+#[cfg_attr(feature = "extras", derive(schemars::JsonSchema))]
+#[serde(tag = "state")]
+pub enum WifiState {
+    #[default]
+    Disconnected,
+    Acquiring,
+    Connected {
+        /// The signal strength of the wifi connection, from 0 to 100.
+        signal_strength: u8,
```
JakeStanger, line 94: Ideally `signal_strength` should be optional, otherwise any profile matching `{ type = "wifi" state = "connected" }` also must specify `signal_strength`.

Unfortunately you'll probably find you have to manually implement `PartialOrd` to get this to work consistently. The `battery` module config code should provide a decent example for that though.

src/modules/networkmanager/config.rs
```diff
@@ -0,0 +1,129 @@
...
+pub enum WiredState {
+    #[default]
+    Disconnected,
+    Acquiring,
+    Connected,
+}
+
+#[derive(Default, Debug, Deserialize, Clone, PartialOrd, PartialEq, Eq)]
+#[cfg_attr(feature = "extras", derive(schemars::JsonSchema))]
+#[serde(tag = "state")]
```
JakeStanger: Any enums used in configs should use `snake_case`. This applies here, as well as to the other connection states and `ProfileState`

```suggestion
#[serde(tag = "state", rename_all = "snake_case")]
```

src/modules/networkmanager/config.rs
```diff
@@ -0,0 +1,129 @@
...
+#[derive(Default, Debug, Deserialize, Clone, PartialOrd, PartialEq, Eq)]
+#[cfg_attr(feature = "extras", derive(schemars::JsonSchema))]
+pub enum VpnState {
+    #[default]
+    Disconnected,
+    Acquiring,
+    Connected,
+}
+
+#[derive(Default, Debug, Deserialize, Clone, PartialOrd, PartialEq, Eq)]
```
JakeStanger: Each of these enums can derive `Copy`

```suggestion
#[derive(Default, Debug, Deserialize, Clone, Copy, PartialOrd, PartialEq, Eq)]
```

src/modules/networkmanager.rs
```diff
@@ -57,52 +226,108 @@ impl Module<GtkBox> for NetworkManagerModule {
...
+                        .load_into_picture_silent(
+                            &icon_name,
+                            icon_size,
+                            false,
+                            widget.downcast_ref::<Picture>().expect("should be Picture"),
+                        )
+                        .await;
+                    widget.set_visible(true)
+                });
+            });
```
JakeStanger: The manager should be attached outside the `recv_glib` closure, as otherwise this re-runs for every update received. Moving it should also allow you to remove the `is_static` call.

src/modules/networkmanager.rs
```diff
@@ -57,52 +226,108 @@ impl Module<GtkBox> for NetworkManagerModule {
...
         let image_provider = context.ironbar.image_provider();
 
-        glib::spawn_future_local({
+        let container_clone = container.clone();
+        let icon_size = self.icon_size;
+        context.subscribe().recv_glib_async((), move |(), update| {
+            let container = container.clone();
             let image_provider = image_provider.clone();
-            let icon = icon.clone();
+            let this = self.clone();
```
JakeStanger: Although it's a bit cumbersome to clone and pass just the config values you need, cloning the whole of `self` means we end up duplicating the whole config for every update.

What might be better is to avoid placing `get_profile_state` and `get_tooltip` on `NetworkManagerModule` directly, and instead creating a new struct that you can pass the required config into, place the methods on there, and then hold onto (an `Rc` of) that struct.

src/modules/networkmanager/config.rs
Rodrigodd: Can I just make it default to 255, or you do prefer that I make it a `Option<u8>`?

JakeStanger: Defaulting to max might be fine in this case actually, would simplify the impl a bit

src/modules/networkmanager/config.rs
Rodrigodd: I struggled a lot messing with the types to get the serialize format to be nice. But now I tested your suggestion and it just worked. Not sure what was doing wrong before.

src/modules/networkmanager.rs
Rodrigodd: I was struggling in getting it to move into the async block below, so I was creating a new owned manager every update. I fixed this together with the cloning issue below.

The `is_static` was never necessary, it is just a left over from when I was debugging this.

src/modules/networkmanager.rs
Rodrigodd: I figured out I don't need to clone those values: the asynchronous code only runs after all those values have already been used. I also noticed that I don't need `recv_glib_async`, since I already moved the async code into a `glib::spawn_future_local` call.

src/modules/networkmanager.rs
```diff
@@ -1,53 +1,211 @@
...
+    #[serde(default)]
+    interface_blacklist: Vec<String>,
+
+    /// If not empty, only devices whose interface name is in this list will be shown.
+    #[serde(default)]
+    interface_whitelist: Vec<String>,
+
     #[serde(flatten)]
     pub common: Option<CommonConfig>,
 }
```
JakeStanger: This should be moved over to `config.rs` actually. Fields (other than `common`) will need to be `pub(super)`

src/modules/networkmanager.rs
```diff
@@ -1,53 +1,211 @@
...
+        | DeviceState::Disconnected => ConnectionState::Disconnected,
+        DeviceState::Prepare
+        | DeviceState::Config
+        | DeviceState::NeedAuth
+        | DeviceState::IpConfig
+        | DeviceState::IpCheck
+        | DeviceState::Secondaries => ConnectionState::Acquiring,
+        DeviceState::Activated => ConnectionState::Connected,
+    }
+}
```
JakeStanger: `impl From<DeviceState> for ConnectionState` would be preferable here

JakeStanger (changes requested):

src/modules/networkmanager/config.rs
```diff
@@ -0,0 +1,264 @@
...
+    pub(super) interface_blacklist: Vec<String>,
+
+    /// If not empty, only devices whose interface name is in this list will be shown.
+    #[serde(default)]
+    pub(super) interface_whitelist: Vec<String>,
+
+    #[serde(flatten)]
+    pub common: Option<CommonConfig>,
+}
+impl NetworkManagerModule {
```
JakeStanger: Sorry my bad for being vague - `impl` should have stayed in the main module, with just the definition moved to `config`.

Rodrigodd: To be more clear, the `impl` of `get_tooltip` and `get_profile_state` and `impl Module` should be in `networkmanager.rs`, and the struct itself and its `impl Default` should be in `config.rs`?

JakeStanger (approved):

src/modules/networkmanager/config.rs
JakeStanger: Yeah, I prefer to keep the configuration definitions and logic separated

PR #1235, opened 2025-11-10
implement a backlight module #962

### LINKED ISSUE #962: Brightness module
Would be great if a module for screen brightness could be added.

xMAC94x:
This implements a brightness module.
Note: I am using 2 dbus implementations here, because `KbdBacklight` does not work for screens, but `login1` does not allow to read the data, only write it. (Reading is done via the filesystem here).

This is my first PR on ironbar :)
Resolves #962

Feel free to give Feedback, I am happy to adjust commits or do followup PRs to further improve quality as necessary :)

### DIFF
--- a/src/modules/brightness.rs
+++ b/src/modules/brightness.rs
@@ -174,9 +174,9 @@ impl BrightnessModule {
     async fn get_brightness(
         client: &brightness::Client,
         datasource: &BrightnessDataSource,
-        default_resource_name: &Option<String>,
+        default_resource_name: Option<&str>,
     ) -> Result<BrightnessData> {
-        let (current, max): (i32, i32) = match &datasource {
+        let (current, max): (i32, i32) = match datasource {
             BrightnessDataSource::Keyboard => {
                 let brightness_kbd = client.keyboard().get_brightness().await?;
                 let max_brightness_kbd = client.keyboard().get_max_brightness().await?;
@@ -184,11 +184,11 @@ impl BrightnessModule {
             }
             BrightnessDataSource::Systemd { subsystem, name } => {
                 let name = name
-                    .clone()
-                    .or_else(|| default_resource_name.clone())
+                    .as_deref()
+                    .or(default_resource_name)
                     .ok_or(SystemdError::NoResourceName)?;
-                let brightness_screen = brightness(subsystem, &name)?;
-                let max_brightness_screen = max_brightness(subsystem, &name)?;
+                let brightness_screen = brightness(subsystem, name)?;
+                let max_brightness_screen = max_brightness(subsystem, name)?;
                 (brightness_screen, max_brightness_screen)
             }
         };
@@ -205,21 +205,21 @@ impl BrightnessModule {
     async fn set_brightness(
         client: &brightness::Client,
         datasource: &BrightnessDataSource,
-        default_resource_name: &Option<String>,
+        default_resource_name: Option<&str>,
         brightness: i32,
     ) -> Result<()> {
-        match &datasource {
+        match datasource {
             BrightnessDataSource::Keyboard => {
                 client.keyboard().set_brightness(brightness).await?;
             }
             BrightnessDataSource::Systemd { subsystem, name } => {
                 let name = name
-                    .clone()
-                    .or_else(|| default_resource_name.clone())
+                    .as_deref()
+                    .or(default_resource_name)
                     .ok_or(SystemdError::NoResourceName)?;
                 client
                     .screen_writer()
-                    .set_brightness(subsystem.to_string(), name, brightness as u32)
+                    .set_brightness(subsystem, name, brightness as u32)
                     .await?;
             }
         };
@@ -278,7 +278,13 @@ impl Module<Button> for BrightnessModule {
                     mut percent,
                     current,
                     max,
-                } = match Self::get_brightness(&client, &datasource, &default_resource_name).await {
+                } = match Self::get_brightness(
+                    &client,
+                    &datasource,
+                    default_resource_name.as_deref(),
+                )
+                .await
+                {
                     Ok(d) => d,
                     Err(err) => match err.downcast::<SystemdError>() {
                         Ok(err) => {
@@ -310,7 +316,7 @@ impl Module<Button> for BrightnessModule {
                         if let Err(err) = Self::set_brightness(
                             &client,
                             &datasource,
-                            &default_resource_name,
+                            default_resource_name.as_deref(),
                             new_brightness,
                         )
                         .await
--- a/src/clients/brightness/dbus_keyboard.rs
+++ b/src/clients/brightness/dbus_keyboard.rs
@@ -21,5 +21,5 @@ pub trait KbdBacklight {
 
     /// BrightnessChangedWithSource signal
     #[zbus(signal)]
-    fn brightness_changed_with_source(&self, value: i32, source: String) -> zbus::Result<()>;
+    fn brightness_changed_with_source(&self, value: i32, source: &str) -> zbus::Result<()>;
 }
--- a/src/clients/brightness/dbus_login1.rs
+++ b/src/clients/brightness/dbus_login1.rs
@@ -7,7 +7,7 @@ use zbus::proxy;
 )]
 pub trait Login1Session {
     /// Sets the current brightness level.
-    fn set_brightness(&self, subsystem: String, name: String, value: u32) -> zbus::Result<()>;
+    fn set_brightness(&self, subsystem: &str, name: &str, value: u32) -> zbus::Result<()>;
 }
 
 #[proxy(


JakeStanger:
Thanks for this, I'll properly review as soon as I get the chance.

I think "brightness" is a better name, as not every screen has a backlight (OLED for example).

xMAC94x:
> Thanks for this, I'll properly review as soon as I get the chance.
> 
> I think "brightness" is a better name, as not every screen has a backlight (OLED for example).

thank you very much :) 
I renamed the module to `brightness`.

xMAC94x:
@JakeStanger fixed the `cargo fmt` error. should be ready for the additional workflow tests.

btw, should I add a test that checks the `brightness` feature flag or do you want to have that in a followup PR ?

src/modules/brightness.rs
```diff
@@ -0,0 +1,307 @@
...
+            common: Some(CommonConfig::default()),
+        }
+    }
+}
+
+impl Default for BrightnessDataSource {
+    fn default() -> Self {
+        BrightnessDataSource::Login1Fs {
+            subsystem: "backlight".to_string(),
+            name: "amdgpu_bl1".to_string(),
```
JakeStanger: I asssume this will only work for the AMDGPU driver? It'd be great if we could try and auto-detect NVIDIA (and I guess Nouveau) and Intel too, as well any other popular drivers.

src/clients/brightness/fs_brightness.rs
```diff
@@ -0,0 +1,43 @@
...
+            ReadParseError::Read(err) => write!(f, "I/O error: {}", err),
+            ReadParseError::Parse(err) => write!(f, "Parse error: {}", err),
+        }
+    }
+}
+
+impl std::error::Error for ReadParseError {}
+
+impl FsLogin1Session {
+    pub fn get_brightness(&self, subsystem: &str, name: &str) -> Result<i32, ReadParseError> {
```
JakeStanger: Getters in Rust tend not to include the 'get' verb:

```suggestion
    pub fn brightness(&self, subsystem: &str, name: &str) -> Result<i32, ReadParseError> {
```

src/clients/brightness/fs_brightness.rs
```diff
@@ -0,0 +1,43 @@
...
+        let mut path = PathBuf::from("/sys/class");
+        path.push(subsystem);
+        path.push(name);
+        path.push("brightness");
+        let s = fs::read_to_string(path).map_err(ReadParseError::Read)?;
+        let value = s.trim().parse::<i32>().map_err(ReadParseError::Parse)?;
+        Ok(value)
+    }
+
+    pub fn get_max_brightness(&self, subsystem: &str, name: &str) -> Result<i32, ReadParseError> {
```
JakeStanger: Same as above

```suggestion
    pub fn max_brightness(&self, subsystem: &str, name: &str) -> Result<i32, ReadParseError> {
```

src/clients/brightness/mod.rs
```diff
@@ -0,0 +1,93 @@
...
+        &self.screen_writer
+    }
+
+    pub fn screen_reader(&self) -> &FsLogin1Session {
+        &self.screen_reader
+    }
+}
+
+const KEY_KDB_MAX_BRIGHTNESS: &str = "KbdBacklight.GetMaxBrightness";
+const KEY_KDB_BRIGHTNESS: &str = "KbdBacklight.GetBrightness";
```
JakeStanger: These keys are needlessly verbose, eg:

```
ironbar var get brightness.KbdBacklight.GetMaxBrightness
```

I'd propose instead:

```suggestion
const KEY_KDB_MAX_BRIGHTNESS: &str = "keyboard-max";
const KEY_KDB_BRIGHTNESS: &str = "keyboard";
```

That leads to the much more succinct:

```
ironbar var get brightness.keyboard-max
```

src/clients/brightness/mod.rs
```diff
@@ -0,0 +1,93 @@
...
+
+#[cfg(feature = "ipc")]
+impl crate::ironvar::Namespace for Client {
```
JakeStanger: The Login1 session also be mapped here, unless there's a good reason it currently isn't?

src/modules/brightness.rs
```diff
@@ -0,0 +1,307 @@
...
+    icons: Icons,
+
+    /// Where to get the brightness data from
+    ///
+    /// **Default**:
+    /// ```
+    /// [data_source.Login1Fs]
+    /// subsystem = "backlight"
+    /// name = "amdgpu_bl1"
+    /// ```
```
JakeStanger: Rather than including the default object in full, it's better to just link to its own docs

```suggestion
    /// See [BrightnessDataSource](#brightness-data-source).
```

(I think it kebab-cases the types, I'd double check that)

src/modules/brightness.rs
```diff
@@ -0,0 +1,307 @@
...
+                        tx.send_update(BrightnessProperties {
+                            icon_name: icons.brightness_icon(percent).to_string(),
+                            screen_brightness: percent,
+                        })
+                        .await;
+                    }
+                    Err(e) => tracing::error!(?e, "Could not retrieve brightness levels"),
+                }
+                sleep(tokio::time::Duration::from_millis(interval)).await;
+            }
```
JakeStanger, line 333: Rather than polling here, Dbus should let you subscribe to property changes as a stream ((at least, it usually does). Sending the updates over a channel would be preferable and would avoid the polling delay.

src/modules/brightness.rs
```diff
@@ -0,0 +1,307 @@
...
+                                        new_cur as u32,
+                                    )
+                                    .await
+                                {
+                                    tracing::error!(?e, "Could not change brightness");
+                                }
+                            }
+                        };
+                    }
+                });
```
JakeStanger: Interacting with clients should be kept to a minimum from the UI code. As much of this as feasible should be moved into the controller as a task listening to updates.

src/modules/brightness.rs
```diff
@@ -0,0 +1,307 @@
...
+            });
+
+            button.add_controller(scroll_controller);
+        }
+
+        let rx = context.subscribe();
+
+        rx.recv_glib(
+            (&self.icons, &self.format),
+            move |(_, format), properties| {
```
JakeStanger: Icons is passed in as a dependency here but not used

```suggestion
            &self.format,
            move |format, properties| {
```

src/modules/brightness.rs
```diff
@@ -0,0 +1,307 @@
...
+        let tx = context.tx.clone();
+        let client = context.try_client::<brightness::Client>()?;
+        let icons = self.icons.clone();
+        let datasource = self.datasource.clone();
+
+        {
+            let scroll_controller =
+                EventControllerScroll::new(EventControllerScrollFlags::VERTICAL);
+
+            scroll_controller.connect_scroll(move |_, _, dy| {
```
JakeStanger: Using scroll speed may not work very well on trackpads, where smooth scrolling can send a lot of large deltas quickly. I've never found a solution to that I particularly like, but the module-level scroll events can be controlled by a multiplier.

https://github.com/JakeStanger/ironbar/blob/25fe56844a5f9bea26b47954df0b84cab7feae41/src/config/common.rs#L137-L141

https://github.com/JakeStanger/ironbar/blob/25fe56844a5f9bea26b47954df0b84cab7feae41/src/config/common.rs#L295-L312

src/modules/brightness.rs
```diff
@@ -0,0 +1,307 @@
...
+                let percentage = properties.screen_brightness;
+                let format = format
+                    .replace("{icon}", &properties.icon_name)
+                    .replace("{percentage}", &percentage.round().to_string());
+
+                button_label.set_label(&format);
+            },
+        );
+
+        Ok(ModuleParts::new(button, None))
```
JakeStanger, line 400: A popup with a slider (`Gtk.Scale`) would be fantastic if you're up for the task. Don't worry if you're not - I can add that in later :)

src/modules/brightness.rs
```diff
@@ -0,0 +1,307 @@
...
+}
+
+#[derive(Debug, Clone, Deserialize)]
+#[cfg_attr(feature = "extras", derive(schemars::JsonSchema))]
+#[serde(default)]
+pub struct Icons {
+    /// Icon to show for respective brightness levels. Needs to be sorted.
+    ///
+    /// **Default**: `[(0, ""), (12, ""), (24, ""), (36, ""), (48, ""), (60,""), (72, ""), (84, ""), (100, "")]`
+    brightness: Vec<(u32, String)>,
```
JakeStanger: I'm aware there's a few different syntaxes sneaking into Ironbar that all achieve roughly the same thing. This might be fine but I want to review it vs other solutions prior to merge.

JakeStanger (changes requested):
I've given this an initial review. It's looking fairly solid already, but there's a few areas that I've raised that need some attention if you can have a look please.

JakeStanger:
> btw, should I add a test that checks the brightness feature flag or do you want to have that in a followup PR ?

I'd regard the CI check as succifient enough for this, but if you want to write more tests then go for it.

src/modules/brightness.rs
xMAC94x: I am fairly new to Gtk and at first I tried it with a popup, but I decided I want to keep the initial PR small and simple, because extending it later on is always possible.  Hopefully this allows us for quicker iterations :)

src/modules/brightness.rs
xMAC94x: Documenting here what we discussed: 
Polling should always be last-resort and event driven communication should be preferred!

However using a crate like `notify` that uses `inotify` under the hood might not work on all platforms / drivers reliable, as `/sys` is mapped directly into the kernel -> we stay with polling for now as it's a reliable fallback.

src/clients/brightness/fs_brightness.rs
```diff
@@ -0,0 +1,102 @@
+use std::{fmt, fs, path::PathBuf};
+
+#[derive(Debug, Default)]
+pub struct FsLogin1Session {}
```
JakeStanger: This struct does not need to exist. Both of its methods take `&self` but do not need to, it has no fields and has no trait implementations.

You'd be better removing it, and moving the methods to top-level functions.

src/clients/brightness/fs_brightness.rs
```diff
@@ -0,0 +1,102 @@
...
+        path.push(subsystem);
+        path.push(name);
+        path.push("max_brightness");
+        let s = fs::read_to_string(path).map_err(ReadParseError::Read)?;
+        let value = s.trim().parse::<i32>().map_err(ReadParseError::Parse)?;
+        Ok(value)
+    }
+}
+
+pub fn default_resource_name(subsystem: &str) -> Option<String> {
```
JakeStanger: I'm not a fan of this implementation. It currently:

1. Takes ALL possible files
2. Filters the list to only files that exist
3. Takes the first of these with a match

This is maximising disk I/O for no reason; 2 and 3 effectively need swapping to start with a list of known possibilities first, and then get the first one that exists.

Falling back to the first existing folder in the case of no match does not make sense to me either - this could be anything.

src/clients/brightness/fs_brightness.rs
```diff
@@ -0,0 +1,102 @@
...
+                "nouveau_backlight",
+                "acpi_video0",
+                "acpi_video1",
+            ]
+            .into_iter();
+            to_check
+                .filter_map(|item| possible_files.iter().find(|v| v.as_str() == item))
+                .next()
+                .cloned()
+                .or_else(|| possible_files.first().cloned())
```
JakeStanger: Cloning should not be necessary here (or in the below branches), since the value owned by the function and exists for the lifetime of the function.

src/clients/brightness/ironvar.rs
JakeStanger, line 1: Although they kinda are now, Ironvars as a system are not intended to be ipc-specific. This module would probably be better called `ironvar`.

src/clients/brightness/ipc.rs
```diff
@@ -0,0 +1,150 @@
...
+        .map(|v| v.to_string())
+    }
+
+    fn list(&self) -> Vec<String> {
+        let mut entries = Vec::new();
+
+        let mut subsystem_path = PathBuf::from(fs_brightness::SYS_PATH);
+        subsystem_path.push(&self.subsystem);
+
+        if let Ok(resource_entries) = std::fs::read_dir(&subsystem_path) {
```
JakeStanger: I'm not a fan of the fact this is carrying its own implementation - it should be querying the client for this information. Otherwise, if the client changes this ends up out of sync.

src/modules/brightness.rs
```diff
@@ -0,0 +1,343 @@
...
+use tokio::time::sleep;
+
+#[derive(Debug, Clone, Deserialize)]
+#[cfg_attr(feature = "extras", derive(schemars::JsonSchema))]
+pub enum BrightnessDataSource {
+    /// using the keyboard dbus. Note: this only works for keyboards, not for screen brightness.
+    Keyboard,
+    /// using the login1 dbus and fs for reading. This works for keyboard and screen brightness, but needs the filesystem for reading the data and dbus for adjusting.
+    Login1Fs {
+        /// The subsystem to take the data from, e.g. `backlight` or `leds`.
```
JakeStanger: It's unclear what "subsystem" actually refers to here.

src/modules/brightness.rs
```diff
@@ -0,0 +1,343 @@
...
+    pub common: Option<CommonConfig>,
+}
+
+impl Default for BrightnessModule {
+    fn default() -> Self {
+        Self {
+            format: "{icon} {percentage}%".to_string(),
+            icons: Icons::default(),
+            datasource: BrightnessDataSource::default(),
+            interval: 5000,
```
JakeStanger: 5 seconds is an exceptionally long time when polling just requires reading a known file from the virtual fs. This probably does not need to be configurable, and could likely be as low as 20-50ms.

src/modules/brightness.rs
```diff
@@ -0,0 +1,343 @@
...
+        ///
+        /// **Default**: `None`
+        name: Option<String>,
+    },
+}
+
+#[derive(Debug, Clone, Deserialize)]
+#[cfg_attr(feature = "extras", derive(schemars::JsonSchema))]
+#[serde(default)]
+pub struct BrightnessModule {
```
JakeStanger: The module is missing a docs page. Are you able to write one pls? If not let me know and I'll fill that in prior to merge.

src/modules/brightness.rs
```diff
@@ -0,0 +1,343 @@
...
+    }
+}
+
+impl Icons {
+    fn brightness_icon(&self, percent: f64) -> String {
+        let percent = percent as u32;
+        self.brightness
+            .iter()
+            .rev() // iterate from the end
+            .find(|&&(v, _)| percent >= v) // find the first where value >= num
```
JakeStanger: These comments are redundant

src/modules/brightness.rs
```diff
@@ -0,0 +1,343 @@
...
+            .map(|(_, label)| label.clone())
+            .unwrap_or("".to_string())
```
JakeStanger:
```suggestion
            .unwrap_or_default()
```

src/modules/brightness.rs
```diff
@@ -0,0 +1,343 @@
...
+    screen_brightness: f64,
+    icon_name: String,
+}
+
+impl BrightnessModule {
+    async fn read_percentage(
+        client: &brightness::Client,
+        datasource: &BrightnessDataSource,
+        default_resource_name: &Option<String>,
+    ) -> Result<(f64, i32, i32)> {
```
JakeStanger: This should probably return a struct with named fields, as it is not obvious at all what these numbers are without reading the implementation.

src/modules/brightness.rs
```diff
@@ -0,0 +1,343 @@
...
+    }
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct BrightnessProperties {
+    screen_brightness: f64,
+    icon_name: String,
+}
+
+impl BrightnessModule {
```
JakeStanger, line 173: The method names are inconsistent here - `read_pecentage` and `write_brightness`. Additionally I'd prefer `get`/`set` as the verbs - read/write doesn't always apply.

src/modules/brightness.rs
```diff
@@ -0,0 +1,343 @@
...
+            }
+        };
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+pub enum UiEvent {
+    AdjustBrightnessScroll(f64),
+    /// Does nothing more than force a fresh of the brightness
+    Refresh,
```
JakeStanger: This comment is redundant

src/modules/brightness.rs
```diff
@@ -0,0 +1,343 @@
...
+    module_impl!("brightness");
+
+    fn spawn_controller(
+        &self,
+        _info: &ModuleInfo,
+        context: &WidgetContext<Self::SendMessage, Self::ReceiveMessage>,
+        mut rx: mpsc::Receiver<Self::ReceiveMessage>,
+    ) -> Result<()> {
+        let tx = context.tx.clone();
+        let ctx = context.controller_tx.clone();
```
JakeStanger: This could be mistaken for `context`, better to use `controller_tx` as the full name

src/modules/brightness.rs
```diff
@@ -0,0 +1,343 @@
...
+                            tracing::error!(?e, "Could not retrieve brightness levels");
+                            continue;
+                        }
+                    };
+
+                if let Some(UiEvent::AdjustBrightnessScroll(dy)) = event {
+                    partial_scroll += dy * scroll_speed;
+
+                    if partial_scroll.abs() >= 1.0 {
+                        let no_steps = partial_scroll.floor() as i32;
```
JakeStanger: `no_steps` sounds more like a number related to zero steps

```suggestion
                        let num_steps = partial_scroll.floor() as i32;
```

src/modules/brightness.rs
```diff
@@ -0,0 +1,343 @@
...
+                        partial_scroll -= partial_scroll.floor();
+
+                        let step_len = ((max as f64 / 100.0) as i32).max(1); // at least modify by 1 if max < 100
```
JakeStanger: >  // at least modify by 1 if max < 100

Why? The comment only describes the code but gives no insight as to why it's doing what it's doing.

src/modules/brightness.rs
```diff
@@ -0,0 +1,343 @@
...
+
+        let button = Button::new();
+        button.set_child(Some(&button_label));
+
+        let ctx = context.controller_tx.clone();
+
+        let scroll_controller = EventControllerScroll::new(EventControllerScrollFlags::VERTICAL);
+        scroll_controller.connect_scroll(move |_, _, dy| {
+            let ctx = ctx.clone();
+            spawn(async move { ctx.send(UiEvent::AdjustBrightnessScroll(dy)).await });
```
JakeStanger: There's an extension method to achieve this pattern:

```suggestion
            ctx.send_spawn(UiEvent::AdjustBrightnessScroll(dy));
```

src/modules/brightness.rs
xMAC94x: done, NOTE: i stole that name from notification module ;)

src/modules/brightness.rs
xMAC94x: hope its more clear now

src/modules/brightness.rs
xMAC94x: I reduced it to 1000 ms for now. reason is: the general usage pattern of screen brightness is you are not changing it often. and if you change it, you are probably using the module, and if you change it via your scroll-wheel it sends an event with the newest percentage anyway. so you only see a delay of 1 second in case you decide to have a module but use some other method to change the brightness and then it's very easy to adjust it in the configs.

src/clients/brightness/ipc.rs
xMAC94x: I moved this to a dedicated function within the client

src/modules/brightness.rs
```diff
@@ -0,0 +1,351 @@
...
+
+    /// Brightness state icons.
+    ///
+    /// See [icons](#icons).
+    icons: Icons,
+
+    /// Where to get the brightness data from
+    ///
+    /// See [BrightnessDataSource].
+    datasource: BrightnessDataSource,
```
JakeStanger: `datasource` feels a bit technical for a config option. I'd prefer `mode` for consistency with existing widgets.

```suggestion
    mode: BrightnessDataSource,
```

src/modules/brightness.rs
JakeStanger: This does not appear to have changed.

src/modules/brightness.rs
```diff
@@ -0,0 +1,351 @@
...
+        let default_resource_name =
+            if let BrightnessDataSource::Login1Fs { subsystem, .. } = &datasource {
+                default_resource_name(subsystem)
+            } else {
+                None
+            };
+
+        spawn(async move {
+            // make sure we have a value on startup and not have to wait for 1 interval
+            let _ = ctx.send(UiEvent::Refresh).await;
```
JakeStanger:
```suggestion
            ctx.send_expect(UiEvent::Refresh).await;
```

src/clients/brightness/fs_reader.rs
```diff
@@ -0,0 +1,98 @@
...
+            // almost all leds have a specific postfix
+            let common_postfix = ["::kdb_backlight"];
+            common_postfix
+                .iter()
+                .find(|item| possible_files.iter().any(|v| v.ends_with(*item)))
+                .map(|s| s.to_string())
+        }
+        _ => None,
+    };
+    prefered.or_else(|| possible_files.into_iter().next())
```
JakeStanger: If no match is found, we should return `None`. Falling back here to the first file in the directory is asking for trouble.

src/clients/brightness/fs_reader.rs
```diff
@@ -0,0 +1,98 @@
...
+                .find(|item| possible_files.iter().any(|v| v == **item))
+                .map(|s| s.to_string())
+        }
+        "leds" => {
+            // almost all leds have a specific postfix
+            let common_postfix = ["::kdb_backlight"];
+            common_postfix
+                .iter()
+                .find(|item| possible_files.iter().any(|v| v.ends_with(*item)))
+                .map(|s| s.to_string())
```
JakeStanger: No point using an array of 1. We can refactor in future if post postfixes need adding.

src/clients/brightness/fs_reader.rs
```diff
@@ -0,0 +1,98 @@
...
+    resource_names
+}
+
+pub fn default_resource_name(subsystem: &str) -> Option<String> {
+    let possible_files = available_resource_names(subsystem);
+
+    let prefered = match subsystem {
+        "backlight" => {
+            // harded list of common names
+            let to_check = [
```
JakeStanger: The list can be a `const`. You'll probably need to specify the actual length.

```suggestion
            // hard-coded list of common names
            const TO_CHECK: [&str, _] = [
```

src/modules/brightness.rs
```diff
@@ -0,0 +1,351 @@
...
+use color_eyre::Result;
+use gtk::{Button, Label};
+use gtk::{EventControllerScroll, EventControllerScrollFlags, prelude::*};
+use serde::Deserialize;
+use tokio::sync::mpsc;
+use tokio::time::sleep;
+
+#[derive(Debug, Clone, Deserialize)]
+#[cfg_attr(feature = "extras", derive(schemars::JsonSchema))]
+pub enum BrightnessDataSource {
```
JakeStanger, line 20: I think this should be tagged, and it needs to use snake case (you'll need to double check what that does to the login1 variant):

```suggestion
#[serde(tag = "type", rename_all = "snake_case")]
pub enum BrightnessDataSource {
```

This will change the config syntax:

```json
{ 
  "mode1": { "type": "keyboard" }
  "mode2": { "type": "login1", "subsystem": "backlight" }
}
```

src/modules/brightness.rs
```diff
@@ -0,0 +1,351 @@
...
+    /// using the keyboard dbus. Note: this only works for keyboards, not for screen brightness.
+    Keyboard,
+    /// using the login1 dbus and fs for reading. This works for keyboard and screen brightness, but needs the filesystem for reading the data and dbus for adjusting.
+    Login1Fs {
```
JakeStanger: I don't like `Login1Fs` as the name here. It's too technical to expose as a config option.

xMAC94x:
> Just thinking, how much use does the `keyboard` backend actually serve? The fact that the `login1` backend requires the filesystem isn't exactly an issue...

We discussed and found out: `login1` is only available when systemd is installed. `keyboard` is a fallback in case systemd is not available. So we decided to keep it in code.

src/modules/brightness.rs
JakeStanger: I still think 1000ms is too long - putting it that high serves no benefit. It can be hard-coded at a much lower interval.

docs/modules/Brightness.md
```diff
@@ -0,0 +1,123 @@
...
+      icons.brightness = [ [ 0 "" ] [ 50 " ] [ 100 "" ] ]
+    }
+  ]
+}
+```
+
+</details>
+
+## Styling
+
```
JakeStanger, line 165: The styling table here is missing:

- Your top-level button will be targetable as `.brightness`.
- The label should ideally have a class (`.label` will do).

src/modules/brightness.rs
JakeStanger: I think we should go with `Systemd`. Only the enum variant itself here needs to change all comments/modules can stay the same to correctly reflect the underlying service.

src/modules/brightness.rs
```diff
@@ -0,0 +1,353 @@
...
+
+    /// Where to get the brightness data from
+    ///
+    /// See [BrightnessDataSource].
+    mode: BrightnessDataSource,
+
+    /// The number of milliseconds between refreshing memory data.
+    ///
+    /// **Default**: `1000`
+    interval: u64,
```
JakeStanger: Let's remove this option and hard-code as 50ms as previously raised please.

src/modules/brightness.rs
```diff
@@ -0,0 +1,353 @@
...
+                let max_brightness_kbd = client.keyboard().get_max_brightness().await?;
+                (brightness_kbd, max_brightness_kbd)
+            }
+            BrightnessDataSource::Login1Fs { subsystem, name } => {
+                let name = name
+                    .clone()
+                    .or_else(|| default_resource_name.clone())
+                    .expect(
+                        "Could not get resource name, consider explicit setting datasource.name",
+                    );
```
JakeStanger: We don't want to panic here. Instead, return an Err variant and handle appropriately in the controller

src/modules/brightness.rs
```diff
@@ -0,0 +1,353 @@
...
+                    tracing::error!(?e, "Could not change brightness");
+                }
+            }
+            BrightnessDataSource::Login1Fs { subsystem, name } => {
+                let name = name
+                    .clone()
+                    .or_else(|| default_resource_name.clone())
+                    .expect(
+                        "Could not get resource name, consider explicit setting datasource.name",
+                    );
```
JakeStanger: Same as above, we don't want to panic

docs/modules/Brightness.md
```diff
@@ -0,0 +1,123 @@
...
+| Selector              | Description                           |
+|-----------------------|---------------------------------------|
+| `.brightness`         | Brightness widget button              |
+| `.brightness.label`   | Notifications widget button           |
```
JakeStanger: The label is a child widget, so needs a space:

```suggestion
| `.brightness .label`   | Notifications widget button           |
```

docs/modules/Brightness.md
```diff
@@ -0,0 +1,123 @@
...
+| `mode.subsystem`                | `backlight` or `leds`                | `backlight`                                                                                            | The name of the subsystem use on the filesystem                                 |
+| `mode.name`                     | `string` or `null`                   | `null`                                                                                                 | When set, using the specific directory, within /sys/class/<subsystem> . If null the module will try to find a reasonable default          |
+| `smooth_scroll_speed`           | `float`                              | `1.0`                                                                                                  | Allows to controll how fast the brightness is changed, e.g. in case touchpad scrolling is used. Negative values swap the scroll direction |
```
JakeStanger:
```suggestion
| `smooth_scroll_speed`           | `float`                              | `1.0`                                                                                                  | Controls how fast the brightness is changed, e.g. in case touchpad scrolling is used. Negative values swap the scroll direction |
```

src/modules/brightness.rs
```diff
@@ -0,0 +1,361 @@
...
+    ) -> Result<()> {
+        match &datasource {
+            BrightnessDataSource::Keyboard => {
+                client.keyboard().set_brightness(brightness).await?;
+            }
+            BrightnessDataSource::Systemd { subsystem, name } => {
+                let name = name
+                    .clone()
+                    .or_else(|| default_resource_name.clone())
+                    .or_else(|| default_resource_name.clone())
```
JakeStanger: This is duplicated:

```suggestion
                    .or_else(|| default_resource_name.clone())
```

docs/modules/Brightness.md
xMAC94x: You are right, i even tested it but forgot to adjust it.

src/clients/brightness/fs_reader.rs
```diff
@@ -0,0 +1,95 @@
...
+                "amdgpu_bl0",
+                "amdgpu_bl1",
+                "intel_backlight",
+                "radeon_bl",
+                "nvidia_0",
+                "nvidia_1",
+                "nouveau_backlight",
+                "acpi_video0",
+                "acpi_video1",
+            ];
```
imnotpoz, line 81:
```
$ ls -l /sys/class/backlight/
total 0
lrwxrwxrwx 1 root root 0 Dec 22 17:40 apple-panel-bl -> ../../devices/platform/soc/389c00000.dcp/backlight/apple-panel-bl
```
add `apple-panel-bl` for asahi linux support

src/clients/brightness/fs_reader.rs
```diff
@@ -0,0 +1,95 @@
...
+            TO_CHECK
+                .iter()
+                .find(|item| possible_files.iter().any(|v| v == **item))
+                .map(|s| s.to_string())
+        }
+        "leds" => {
+            // almost all leds have the same postfix
+            possible_files
+                .into_iter()
+                .find(|v| v.ends_with("::kdb_backlight"))
```
imnotpoz:
```
$ ls -l /sys/class/leds
total 0
lrwxrwxrwx 1 root root 0 Dec 22 17:40 input6::capslock -> ../../devices/platform/soc/2a9b14000.fifo/2a9b30000.input/0019:05AC:0352.0003/input/input6/input6::capslock
lrwxrwxrwx 1 root root 0 Dec 22 17:40 input6::compose -> ../../devices/platform/soc/2a9b14000.fifo/2a9b30000.input/0019:05AC:0352.0003/input/input6/input6::compose
lrwxrwxrwx 1 root root 0 Dec 22 17:40 input6::kana -> ../../devices/platform/soc/2a9b14000.fifo/2a9b30000.input/0019:05AC:0352.0003/input/input6/input6::kana
lrwxrwxrwx 1 root root 0 Dec 22 17:40 input6::numlock -> ../../devices/platform/soc/2a9b14000.fifo/2a9b30000.input/0019:05AC:0352.0003/input/input6/input6::numlock
lrwxrwxrwx 1 root root 0 Dec 22 17:40 input6::scrolllock -> ../../devices/platform/soc/2a9b14000.fifo/2a9b30000.input/0019:05AC:0352.0003/input/input6/input6::scrolllock
lrwxrwxrwx 1 root root 0 Dec 22 17:40 kbd_backlight -> ../../devices/platform/led-controller/leds/kbd_backlight
```

this is just `kbd_backlight` on asahi, not sure if the best way would be to check if it ends with `kbd_backlight`, or add `|| v == "kbd_backlight"`

JakeStanger: I can't see why dropping the `::` would cause any issues

JakeStanger:
Haven't forgotten about this, just struggling for time right now. Will get a decision on the thresholds system and get back to you soon

src/modules/brightness.rs
JakeStanger: I've implemented the profiles system in #1349, which is documented in the following locations:

The volume and battery modules also provide implementations which can be used as reference. 

Can this be refactored to switch over to using that system for the brightness icons please? Give me a shout if you need a hand.

You will need to define a `profiles` key with a custom profile type, which should have a `brightness_icon` key.

docs/modules/Brightness.md
```diff
@@ -0,0 +1,172 @@
...
+  ]
+}
+```
+
+</details>
+
+### Icons
+
+The icon is configured per profile using the `icon_label` option. The icon is rendered separately from the `format` label,
+so you can include it in `format` if you prefer, or leave `format` as text-only.
```
JakeStanger: This clause can be removed

src/clients/brightness/fs_reader.rs
```diff
@@ -0,0 +1,96 @@
...
+    resource_names.sort();
+    resource_names
+}
+
+pub fn default_resource_name(subsystem: &str) -> Option<String> {
+    let possible_files = available_resource_names(subsystem);
+
+    match subsystem {
+        "backlight" => {
+            // harded list of common names
```
JakeStanger:
```suggestion
            // list of common names
```

src/modules/brightness.rs
```diff
@@ -0,0 +1,396 @@
...
+    }
+}
+
+impl std::error::Error for SystemdError {}
+
+impl BrightnessModule {
+    async fn get_brightness(
+        client: &brightness::Client,
+        datasource: &BrightnessDataSource,
+        default_resource_name: &Option<String>,
```
JakeStanger: Taking `Option<&str>` would be preferable. You can use `Option::as_deref()` to convert.

Same for `set_brightness`.

src/modules/brightness.rs
```diff
@@ -0,0 +1,396 @@
...
+                    mut percent,
+                    current,
+                    max,
+                } = match Self::get_brightness(&client, &datasource, &default_resource_name).await {
+                    Ok(d) => d,
+                    Err(err) => match err.downcast::<SystemdError>() {
+                        Ok(err) => {
+                            tracing::error!(
+                                ?err,
+                                "Could not retrieve brightness levels. Error is unrecoverable, fix the config! Stopping brightness module"
```
JakeStanger: Not sure what "fix the config!" means here? Module being stopped is implied from the fact the error is unrecoverable

```suggestion
                                "Could not retrieve brightness levels. Error is unrecoverable"
```

xMAC94x:
FYI: there exist another commit here, IDK if I should do another push at this point: https://github.com/JakeStanger/ironbar/commit/75d01c3e1cd0b93ce5f4023b9b45ef8da2f34902

PR #1236, opened 2025-11-12
Fix formatting for battery thresholds section

ksharizard:
This PR fixes toml config and stops ironbar from erroring out when using the battery module.


JakeStanger:
Thanks!

PR #1237, opened 2025-11-12
Fix typos and improve clarity in Scripts.md

ksharizard:
- Fixed some grammatical mistakes here and there
- Fixed the TOML config being written as YAML

> Shorthand can be used in all cases, but there are some cases (such as embedding scripts inside strings) where longhand
cannot be used.

I was not sure whether or how to fix the above line. I am not sure if longhand is a typo or not and if not, then there is no need for a "but".

### DIFF
--- a/docs/Scripts.md
+++ b/docs/Scripts.md
@@ -22,8 +22,8 @@ type.
 ## Writing script configs
 
 There are two available config formats for scripts: shorthand, as a string, or longhand, as an object.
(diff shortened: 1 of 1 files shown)


JakeStanger:
Thanks again

> there is no need for a "but".

I think you're right. It probably reads better as two sentences, and switched around.

> There are some cases (such as embedding scripts inside strings) where longhand cannot be used.
> Shorthand can be used in all cases.

ksharizard:
> I think you're right. It probably reads better as two sentences, and switched around.

Fixed.

Issue #1039, opened 2025-06-09
GTK symbolic icons are not themed


Rodrigodd:
**Describe the bug**

The battery module, for example, displays a symbolic icon (`battery-missing-symbolic` in my case), but the icon is not themed. The same happens with other uses of symbolic icons in Ironbar.

**To reproduce**

1. Add any module that displays a GTK symbolic icon, such as `battery`.
2. The icon appears with a dark gray color:

![Image](https://github.com/user-attachments/assets/efa2b89e-8fc4-434f-b0c1-43fe2d0fe762)

**Expected behavior**

My system is configure with theme "Adwaita:dark", so the icon should be white like so:

![Image](https://github.com/user-attachments/assets/8c5f78ce-625f-4eae-8a6c-6658181146cf)

**System information:**
- Distro: [e.g. Arch Linux, Ubuntu 22.10]
- Compositor: [e.g. Sway]
- Ironbar version: [e.g. 0.16.1]

**Possible fixes**

In my fork, I’m using the following diff:

```diff
diff --git a/src/image/provider.rs b/src/image/provider.rs
index 5004c37..ee22ea7 100644
--- a/src/image/provider.rs
+++ b/src/image/provider.rs
@@ -4,6 +4,7 @@ use crate::{glib_recv_mpsc, send_async, spawn};
 use cfg_if::cfg_if;
 use color_eyre::{Help, Report, Result};
 use gtk::cairo::Surface;
+use gtk::gdk::RGBA;
 use gtk::gdk::ffi::gdk_cairo_surface_create_from_pixbuf;
 use gtk::gdk_pixbuf::Pixbuf;
 use gtk::prelude::*;
@@ -198,6 +199,10 @@ impl<'a> ImageProvider<'a> {
 
         let pixbuf = match &self.location {
             ImageLocation::Icon { name, theme } => self.get_from_icon(name, theme, scale),
+            // ImageLocation::Icon { name, theme } => {
+            //     image.set_icon_name(Some(name));
+            //     return Ok(());
+            // }
             ImageLocation::Local(path) => self.get_from_file(path, scale),
             ImageLocation::Steam(steam_id) => self.get_from_steam_id(steam_id, scale),
             #[cfg(feature = "http")]
@@ -231,7 +236,14 @@ impl<'a> ImageProvider<'a> {
     fn get_from_icon(&self, name: &str, theme: &IconTheme, scale: i32) -> Result<Pixbuf> {
         let pixbuf =
[...]

PR #1241, opened 2025-11-19
fix(image): symbolic icons not rendering in GTK 4.20+

slowsage:
For symbolic icons to work with ironbar's `icon_theme`, we need to adapt to GTK 4.20 [changes](https://docs.gtk.org/gtk4/migrating-3to4.html#adapt-to-gtkicontheme-api-changes).

Required more refactoring than intended to
1. keep the Image vs Picture check localized
3. handle both sync (Image) and async (Picture).

Also removed duplicate `feature = "clipboard",`.

Relevant:
https://docs.gtk.org/gtk4/migrating-3to4.html#adapt-to-gtkicontheme-api-changes
https://docs.gtk.org/gtk4/migrating-3to4.html#dont-use-gtk-icon-theme-in-your-css

Fixes: #1039

Given the scope, prolly makes sense for v0.19.

### DIFF
--- a/src/image/gtk.rs
+++ b/src/image/gtk.rs
@@ -1,13 +1,56 @@
 use crate::gtk_helpers::IronbarLabelExt;
 use crate::image;
 use gtk::prelude::*;
-use gtk::{Button, ContentFit, Label, Orientation, Picture};
+use gtk::{Button, ContentFit, Image, Label, Orientation, Picture};
+use std::cell::RefCell;
 use std::ops::Deref;
+use std::rc::Rc;
+
+const ICON_PREFIX: &str = "icon:";
+const IMAGE_CLASSES: &[&str] = &["icon", "image"];
+
+fn create_icon<F>(input: &str, size: i32, provider: &image::Provider, on_result: F)
+where
+    F: FnOnce(Result<gtk::Widget, ()>) + 'static,
+{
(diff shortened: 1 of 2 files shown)


slowsage:
> What impact this has on GTK versions prior to 4.20? I would assume nothing but I'd like to confirm that prior to merge.

My understanding it it should work on gtk 4 prior to 4.20 - simply avoided `is_symbolic` (carried over from 3.0) and `get_icon_name` (added in 4.0) - both got removed in 4.20.

> Did the GTK4 port cause a regression in the behaviour in any way, or is this fixing the issue as it was in GTK3?
AFAIK,  the issue was there in both versions:

- GTK3 - ironbar should have used `set_from_icon_name` for symbolic icons - do not think it did.
- Even if it did, it would not have respected `icon_theme`.
- The proper `icon_theme` respecting solution is only possible in gtk4 AFAIK.

Edit: In gtk3, there was a css based gtk-icon-theme config so it may have been possible to inject `icon_theme` into that to set theme even on a per-module basis. But per [this](https://docs.gtk.org/gtk4/migrating-3to4.html#dont-use-gtk-icon-theme-in-your-css): `GTK 4 always uses the current icon theme, with no way to change this.`

JakeStanger:
Thanks for clarifying. I'll give it a quick test to make sure, but otherwise sounds like targeting 0.19 is sensible.

Issue #1186, opened 2025-10-06
Bluetooth module popup cramped


jficz:
**Describe the bug**
Bluetooth popup is cramped.

**To reproduce**

1. Add module `bluetooth`
2. Click popup
3. Observe

I need to scroll down to see all of my BT devices (I'v got about 10 of them).

**Expected behavior**
A bigger popup, wide enough to fit a device name on a single line and high enough to fit ideally all of them or up to a user-specified limit (as in device limit, not absolute size).

**System information:**
- Distro: Debian 13 + Home Manager (nixpkgs-unstable)
- Compositor: Hyprland 0.51.1
- Ironbar version: 0.18.0-pre (from upstream flake)

**Configuration**

<details><summary>Config</summary>

```
{
  "center": [
    {
      "type": "focused"
    }
  ],
  "end": [
    ...
    {
      "type": "bluetooth"
    },
    ...
  ],
  "height": 36,
  "position": "top",
  "start": [
    {
      "all_monitors": true,
      "type": "workspaces"
    }
  ]
}
```

</details>

<details><summary>Styles</summary>

```css
N/A (no explicit styles set)
```

</details>

**Additional context**

n/a
**Screenshots**

<img width="331" height="104" alt="Image" src="https://github.com/user-attachments/assets/20e93883-9e7d-4832-94e9-435bd04835db" />

benoitlx:
I know it's a workaround, but have you tried setting `popup.scrollable` to `false` ?
In my case, with this option enabled, all of my devices are listed in a single block and I don't need to scroll in a cramped popup.

<img width="324" height="358" alt="Image" src="https://github.com/user-attachments/assets/a063ea87-e4f0-4f48-8a6f-222b100c84ac" />

jficz:
Thanks! I missed this option. Guess that would be a good enough workaround for me, though not closing since this would still be just a workaround. But I'd say it lowers the prio of this issue considerably @JakeStanger

JakeStanger:
Cool good to know. I'm still regarding this as a regression from the GTK4 port, so do want to get it fixed before next release

JakeStanger:
Height issues should be resolved. Can anybody confirm if they are still experiencing width issues with `scrollable = true` please?

benoitlx:
I will test in the week, can't guaranty when though.

jficz:
looks good with the latest flake build

JakeStanger:
In that case I'll close this as resolved. Any further issues let me know.

Issue #1050, opened 2025-06-15
Clip `clipboard` popup entries


postsolar:
**Is your feature request related to a problem? Please describe.**
> A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]

I don't currently use `clipboard` module because there's CopyQ and its tray icon, but back when I used to use it I would often copy something with either a lot of lines or very long lines. In these cases the popup size would even exceed my monitor size.

**Describe the solution you'd like**
> A clear and concise description of what you want to happen.
> The more info here about what you are trying to achieve, the better - there's likely more than one way to go about implementing a solution.

I've been thinking about #1011 and remembered about this issue, so I thought just wrapping entries (to address entries' width and height) and the entry list itself (to address the number of entries) into ScrolledWindow should be simple and straightforward.

**Describe alternatives you've considered**
> A clear and concise description of any alternative solutions or features you've considered.

I personally just use CopyQ anyways.

**Additional context**
> Add any other context or screenshots about the feature request here.

Here's what it currently looks like:

![Image](https://github.com/user-attachments/assets/3adb36f9-2918-42ad-946a-a340709af499)

A long line was copied and the popup doesn't fit on the screen.

postsolar:
Actually I'm dumb I forgot `truncate` exists for this module as well 🤦.
But anyways, there's no vertical truncation, so still the same issue with the widget exceeding monitor size.

PR #1245, opened 2025-11-21
[bluetooth] grow popup size until a configureable limit based on the number of devices

xMAC94x:
see issue #1186, though this does not fix the width problems

### DIFF
--- a/docs/modules/Bluetooth.md
+++ b/docs/modules/Bluetooth.md
@@ -17,7 +17,8 @@ Clicking on the widget opens a popout displaying list of available devices and c
 | `format.enabled`              | `string`  | ` On`                                                | Format string to use for the widget button when bluetooth adapter is enabled but no devices are connected.                                       |
 | `format.connected`            | `string`  | ` {device_alias}`                                    | Format string to use for the widget button when bluetooth adapter is enabled and a device is connected.                                          |
 | `format.connected_battery`    | `string`  | ` {device_alias} • {device_battery_percent}%`        | Format string to use for the widget button when bluetooth adapter is enabled, a device is connected and `{device_battery_percent}` is available. |
-| `popup.max_height`            | `integer` | `330`                                                 | The maximum number of pixels the window can reach before scrolling. Leave blank to allow the popup to grow indefinitely.                         |
+| `popup.max_height.pixels`     | `integer` | `330`                                                 | The maximum number of pixels the window can reach before scrolling. Leave blank to allow the popup to grow indefinitely.                         |
+| `popup.max_height.devices`    | `integer` | `null`                                                | See `popup.max_height.pixels`, but the limit is set in number of devices. Specify either `pixels` or `devices`                                   |
 | `popup.header`                | `string`  | ` Enable Bluetooth`                                  | Format string to use for the header of popup window.                                                                                             |
 | `popup.disabled`              | `string`  | `{adapter_status}`                                    | Format string to use for the message that is displayed when the adapter is not found or disabled.                                                |
 | `popup.device.header`         | `string`  | `{device_alias}`                                      | Format string to use for the header of device box.                                                                                               |
@@ -52,7 +53,9 @@ Clicking on the widget opens a popout displaying list of available devices and c
         "connected_battery": " {device_alias} • {device_battery_percent}%"
       },
       "popup": {
-        "scrollable": true,
+        "max_height": {
+          "devices": 5
+        },
         "header": " Enable Bluetooth",
         "disabled": "{adapter_status}",
         "device": {
@@ -98,7 +101,7 @@ connected = " {device_alias}"
 connected_battery = " {device_alias} • {device_battery_percent}%"
 
 [end.popup]
-scrollable = true
+max_height.devices = 5
 header = " Enable Bluetooth"
 disabled = "{adapter_status}"
 
@@ -138,7 +141,8 @@ end:
       connected: " {device_alias}"
       connected_battery: " {device_alias} • {device_battery_percent}%"
     popup:
-      scrollable: true
+      max_height:
+        devices: 5
       header: " Enable Bluetooth"
       disabled: "{adapter_status}"
       device:
@@ -175,7 +179,7 @@ end:
       format.enabled = " On"
       format.connected = " {device_alias}"
       format.connected_battery = " {device_alias} • {device_battery_percent}%"
-      popup.scrollable = true
+      popup.max_height.devices = 5
       popup.header = " Enable Bluetooth"
       popup.disabled = "{adapter_status}"
       popup.device.header = "{device_alias}"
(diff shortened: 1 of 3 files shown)


postsolar:
Could you specify what unit it is in the docs? Like, pixels, centimeters, devices, bananas :smile:

xMAC94x:
> Could you specify what unit it is in the docs? Like, pixels, centimeters, devices, bananas 😄

will do

JakeStanger:
Thanks for this. I'm going to hold off merging this until I have time to properly dig into the issue, because I feel like introducing more config options to resolve it is not the right way to go, especially pixel-based.

kuzy000:
I think this should also work (right after creating `devices`):
```rust
devices.set_max_content_height(grow_height_until);
devices.set_propagate_natural_height(true);
```

Shame that `max-content-height` isn't accessible from CSS.

JakeStanger:
Pros/cons of each approach?

If we do go for this option btw, I'm thinking `max_height` would be a better name for the config option.

xMAC94x:
limiting the `max` is kinda undesired, when we have set `scrollable = false`.

Or we only do this in case when scroll-able is enabled 🤷

kuzy000:
> Or we only do this in case when scroll-able is enabled 🤷

I think `scrollable = false`  is the same as setting `max_height = 99999` . So maybe we can replace `scrollable` with something like  `max_height = -1` or `max_height = null`? Seems like `volume` module have something similar with its `truncate.length`: the default value is `null`. 

Maybe `max_height: Option<i32>` instead of `scrollable` is a good idea.

https://github.com/JakeStanger/ironbar/wiki/volume#configuration

JakeStanger:
> Maybe `max_height: Option<i32>` instead of scrollable is a good idea.

Agreed with this

xMAC94x:
> > Maybe `max_height: Option<i32>` instead of scrollable is a good idea.
> 
> Agreed with this

implemented this.

When revieweing you might probably see the `> 0` check in the `match` clause. This is necessary for the `toml` configs which DO NOT allow unsetting a default value. So when we have a default value for `max_height` other than `null` (like we have) -> toml can NEVER unset this max_height back to `None` ...
And there are too many issues, like https://github.com/toml-lang/toml/issues/30 about this :/

src/modules/bluetooth/config.rs
```diff
@@ -112,7 +113,7 @@ pub struct PopupConfig {
 impl Default for PopupConfig {
     fn default() -> Self {
         Self {
-            scrollable: true,
+            max_height: Some(330),
```
JakeStanger: Is 330 based on anything in particular?

src/modules/bluetooth/config.rs
```diff
@@ -90,10 +90,11 @@ impl Default for FormatConfig {
 #[cfg_attr(feature = "extras", derive(schemars::JsonSchema))]
 #[serde(default)]
 pub struct PopupConfig {
-    /// Whether to make the popup scrollable or stretchable to show all of its content.
+    /// Grows the Height of the Popup automatically based on the content up to this limit in pixel.
+    /// When set a scrollbar will be used if the content doesnt fit. If null the popup grows as needed
```
JakeStanger: This isn't worded very clearly currently (and same in docs). 

```suggestion
    /// The maximum number of pixels the window can reach before scrolling. 
    /// Leave blank to allow the popup to grow indefinitely.
```

src/modules/bluetooth/config.rs
xMAC94x: It's a semi-useful number that I thing results in a good visual height.

And it has the effect, when the List is full the last Item will be chopped in half, which is indicating users clearly that the popup is scroll-able. (tested with `bluetoothctl` and `scan on` )

postsolar:
What is the tradeoff between limiting to number of devices and limiting to specific amount of pixels? I think the final choice will be relevant to some other modules too, e.g. #1050 (and as I just discovered it actually stops opening the popup once intended height approaches monitor height).

xMAC94x:
> What is the tradeoff between limiting to number of devices and limiting to specific amount of pixels? I think the final choice will be relevant to some other modules too, e.g. #1050 (and as I just discovered it actually stops opening the popup once intended height approaches monitor height).

I don't know how to translate between items and pixels. You can change the height of the entries via CSS and I maybe its also possible to have individual entries with different height than others. Do you have an algorithm in mind here ?

src/modules/bluetooth/config.rs
JakeStanger: > And it has the effect, when the List is full the last Item will be chopped in half, which is indicating users clearly that the popup is scroll-able.

This fully depends on styles - any font size, padding, margin or min-height changes can offset this. What was this tested against?

JakeStanger:
Calculating the height reliably would be nigh impossible. It's possible to get the actual height after everything's rendered, but I can't imagine that's very useful.

xMAC94x:
agree, @postsolar  would you also be fine with a pixel version ? 
For personal configuration you can prob do the calculation `item.len() -> required pixels` once and then set it via the config.

postsolar:
I have stitched something together, not sure how robust this is, but seems to work.

[bluetooth-max-devices.patch](https://github.com/user-attachments/files/23765888/bluetooth-max-devices.patch) - this applies to [e1cbed0](https://github.com/JakeStanger/ironbar/commit/e1cbed04dce3bde6ee9cabc7d317b93cab5bb651)

I found this to work with both my usual GTK theme (bare), my usual GTK theme + my usual ironbar [userstyles](https://github.com/postsolar/config/blob/c41ba59243cde70ce42e6ad55866f125c38d6c75/home-manager/ironbar/style.scss#L221-L229), Adwaita (bare), and my usual styles + setting huge font size in GTK Inspector. It seems to be reactive and intuitive, in a quick test I could not find any obvious failure mode.

I'm not saying I have an exact vision for final design, just saying limiting by devices seems to work.

src/modules/bluetooth/config.rs
xMAC94x: I messed up, i wanted to test with defaults, but accedently had:
```css
* {
  font-family: "JetBrainsMono NF ExtraBold", "Font Awesome 6 Free",
    "Font Awesome 6 Brands", monospace;
  font-size: 16px;
}
```

unfortuatly I dont have enough bluetooth devices here to verify another number

docs/modules/Bluetooth.md
```diff
@@ -52,7 +53,9 @@ Clicking on the widget opens a popout displaying list of available devices and c
         "connected_battery": " {device_alias} • {device_battery_percent}%"
       },
       "popup": {
-        "scrollable": true,
+        "max_height": {
+          count: 5
```
postsolar:
```suggestion
          "count": 5
```

src/modules/bluetooth/config.rs
```diff
@@ -86,14 +86,24 @@ impl Default for FormatConfig {
     }
 }
 
+#[derive(Debug, Deserialize, Clone)]
+#[cfg_attr(feature = "extras", derive(schemars::JsonSchema))]
+#[serde(rename_all = "lowercase")]
+pub enum SizeLimit {
+    Count(i32),
+    Pixel(i32),
```
JakeStanger: I think `count` is a bit ambiguous (count of what?) - `devices` is a more descriptive name. Pixels should also be plural:

```suggestion
    Devices(i32),
    Pixels(i32),
```

PR #1247, opened 2025-11-21
feat: add configurable tray icon click actions + double click actions…

### LINKED ISSUE #1242: FR: Optional pre-GTK4 port behavior on clicking tray icons

I really liked clicking on tray icons with LMB to open menus, as I use a touchpad so it's always physically easier to tap with one finger (= left click) than with two (= right click), especially when my left hand's fingers are on the keyboard but the thumb reaches the touchpad. Current single-click behavior is not that useful for me.

Configurable behavior of tray module for LMB/RMB (or single vs double click).

**Additional context**

postsolar:
… for all modules

- Add double-click support for all modules via CommonConfig
- Add TrayClickAction enum with built-in and custom command support
- Support placeholders in custom commands: {name}, {title}, {icon}, {address}
- Add reusable connect_pressed_with_double_click() to IronbarGtkExt
- Update documentation for new click action configuration
- Full backward compatibility

Closes #1242

### DIFF
--- a/src/modules/tray/mod.rs
+++ b/src/modules/tray/mod.rs
@@ -17,6 +17,13 @@ use system_tray::client::{ActivateRequest, UpdateEvent};
 use tokio::sync::mpsc;
 use tracing::{debug, error, trace, warn};
 
+/// Icon configuration for tray items
+struct IconConfig {
+    theme: IconTheme,
+    size: u32,
+    prefer_theme: bool,
+}
+
 /// Reserved tray click actions
 #[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
 #[cfg_attr(feature = "extras", derive(schemars::JsonSchema))]
@@ -247,7 +254,11 @@ impl Module<gtk::Box> for TrayModule {
             let activated_channel = context.controller_tx.clone();
 
             let provider = context.ironbar.image_provider();
-            let icon_theme = provider.icon_theme();
+            let icon_config = IconConfig {
+                theme: provider.icon_theme().clone(),
+                size: self.icon_size,
+                prefer_theme: self.prefer_theme_icons,
+            };
 
             // listen for UI updates
             let click_handlers = self.click_handlers.clone();
@@ -257,9 +268,7 @@ impl Module<gtk::Box> for TrayModule {
                     update,
                     &container,
                     &mut menus,
-                    &icon_theme,
-                    self.icon_size,
-                    self.prefer_theme_icons,
+                    &icon_config,
                     &activated_channel,
                     &click_handlers,
                 );
@@ -279,9 +288,7 @@ fn on_update(
     update: Event,
     container: &gtk::Box,
     menus: &mut HashMap<Box<str>, TrayMenu>,
-    icon_theme: &IconTheme,
-    icon_size: u32,
-    prefer_icons: bool,
+    icon_config: &IconConfig,
     activated_channel: &mpsc::Sender<ActivateRequest>,
     click_handlers: &TrayClickHandlers,
 ) {
@@ -295,7 +302,12 @@ fn on_update(
             let x: Option<&gtk::Widget> = None;
             container.insert_child_after(&menu_item.widget, x);
 
-            if let Ok(image) = icon::get_image(&menu_item, icon_size, prefer_icons, icon_theme) {
+            if let Ok(image) = icon::get_image(
+                &menu_item,
+                icon_config.size,
+                icon_config.prefer_theme,
+                &icon_config.theme,
+            ) {
                 menu_item.set_image(&image);
             } else {
                 let label = menu_item.title.clone().unwrap_or(address.clone());
@@ -325,7 +337,12 @@ fn on_update(
 
                     if icon_name.as_ref() != menu_item.icon_name() {
                         menu_item.set_icon_name(icon_name);
-                        match icon::get_image(menu_item, icon_size, prefer_icons, icon_theme) {
+                        match icon::get_image(
+                            menu_item,
+                            icon_config.size,
+                            icon_config.prefer_theme,
+                            &icon_config.theme,
+                        ) {
                             Ok(image) => menu_item.set_image(&image),
                             Err(e) => {
                                 error!("error loading icon: {e}");


postsolar:
For tray I think ideal would be per-item configuration like
```toml
[tray.items.nm-applet]
on_click_left = "do something"

[tray.items.telegram-desktop]
on_click_right = "do something else"
```
but too complicated + YAGNI.

postsolar:
There is one limitation currently, for tray there's some GTK spam about some CSS errors when opening certain items' menus now (e.g. nm-applet, but not copyq) - IDK what this is and what to do about it, but it's harmless, doesn't seem to be impacting anything.

docs/modules/Tray.md
```diff
@@ -6,11 +6,52 @@ Displays a fully interactive icon tray using the KDE `libappindicator` and `com.
...
+| `on_click_left_double`   | `string`                                                   | `'none'`                | Action to perform on double-left-click. See [Click Actions](#click-actions) below.                                                                                  |
+
+### Click Actions
+
+Click actions can be one of the following built-in actions, or any custom shell command:
+
+**Built-in actions:**
+- `open_menu` - Opens the tray icon's popup menu
+- `trigger_default` - Triggers the tray icon's default (primary) action
+- `trigger_secondary` - Triggers the tray icon's secondary action
```
JakeStanger: Opinions on shortening these? They feel a little verbose to me


```suggestion
- `menu` - Opens the tray icon's popup menu
- `default` - Triggers the tray icon's default (primary) action
- `secondary` - Triggers the tray icon's secondary action
```

docs/modules/Tray.md
```diff
@@ -6,11 +6,52 @@ Displays a fully interactive icon tray using the KDE `libappindicator` and `com.
...
+- `none` - Do nothing
+
+**Custom commands:**
+
+Any other string is treated as a custom shell command. Custom commands support the following placeholders:
```
JakeStanger, line 33: > Any other string is treated as a custom shell command.

I'm wondering if I should adopt this in the `custom` module too? Would be a breaking change but currently shell commands need to be prefixed with a `!`, which is the only place that uses that syntax.

src/gtk_helpers.rs
```diff
@@ -61,6 +78,80 @@ impl<W: IsA<Widget>> IronbarGtkExt for W {
...
+                    let pending = pending_single_click.clone();
+
+                    // Cancel any existing pending single-click
+                    if let Some(source_id) = pending.take() {
+                        source_id.remove();
+                    }
+
+                    // Schedule single-click action after double-click timeout (250ms)
+                    let source_id =
+                        glib::timeout_add_local_once(Duration::from_millis(250), move || {
```
JakeStanger: This interval should ideally be configurable for accessibility purposes, assuming that's not a logistical nightmare. It can be a top-level option. 250ms might be a long default time too, if it causes the single-click action to be delayed that long.

src/modules/tray/mod.rs
```diff
@@ -131,6 +215,11 @@ impl Module<gtk::Box> for TrayModule {
             let icon_theme = provider.icon_theme();
 
             // listen for UI updates
+            let on_click_left = self.on_click_left.clone();
+            let on_click_right = self.on_click_right.clone();
+            let on_click_middle = self.on_click_middle.clone();
+            let on_click_left_double = self.on_click_left_double.clone();
```
JakeStanger: Could we group these up into a eg `ClickHandlers` struct pls, rather than passing all 4 everywhere? It can use `#[serde(flatten)]` on the config.

src/modules/tray/mod.rs
```diff
@@ -17,6 +17,44 @@ use system_tray::client::{ActivateRequest, UpdateEvent};
...
+        let s = String::deserialize(deserializer)?;
+        Ok(match s.as_str() {
+            "open_menu" => Self::OpenMenu,
+            "trigger_default" => Self::TriggerDefault,
+            "trigger_secondary" => Self::TriggerSecondary,
+            "none" => Self::None,
+            _ => Self::Custom(s),
+        })
+    }
+}
```
JakeStanger: What is the reason for implementing this manually? Serde should be able to derive this as far as I can tell.

src/modules/tray/interface.rs
JakeStanger, line 1: There's a lot of duplication happening here, especially in the event handlers (to be fair, there already was). Are you able to see if you can move some of the repeated logic into a helper func/macro please?

docs/modules/Tray.md
postsolar: I could see the benefit to explicit and predictable `!`, but if everything else doesn't require it yeah maybe it's better to remove it for greater consistency. Honestly I could see pros and cons with either approach.

src/gtk_helpers.rs
postsolar: We have two options here. Which one do you prefer?

### gtk-double-click-time

There is https://docs.gtk.org/gtk4/property.Settings.gtk-double-click-time.html. However, I'm not sure how it's meant to be configured, I guess it depends on specific setup/distro. Also the default value for this is even longer, 400ms. So I imagine most baseline users would need to fiddle with DBus or gsettings or `~/.config/gtk-4.0/settings.ini` or whatnot. I myself wasn't even able to quickly find how to set it. Let me compile and see if it works via the simple `~/.config/gtk-4.0/settings.ini`, I will update you here.

### Ironbar native implementation

Much simpler and more controllable, but renders somebody's global GTK config useless and requires duplicating their settings for Ironbar too.

src/modules/tray/mod.rs
postsolar: It's not able to derive it _correctly_, it matches everything to Custom. ~~I will update with a comment though.~~ Looking at ScriptInput though I think I could make a nested structure.

src/gtk_helpers.rs
postsolar: I lean:
```jsonc
// top-level config
  // ...
  "double_click_time": 200, // positive int or special value "gtk", default is 200ms
  // ...
```

postsolar:
@JakeStanger all feedback addressed. For configuring `gtk-double-click-time` I wasn't able to do it with `~/.config/gtk-4.0/settings.ini` (maybe it's issues with my setup, IDK), but I was able to confirm it works via GTK Inspector (I set it there and Ironbar behavior was reflecting this).

I chose to keep the default to be 250ms. I think 200ms is too aggressive, I personally struggle with this :laughing:

JakeStanger:
I'm wondering if it might be better to run the single-click action *always*, which'd avoid the delay. It would potentially be more limited, but it would be more inline with standard double-click behaviour in other applications.

postsolar:
I think it only works for situations where single-click action is some sort of a light toggle, e.g. opening a menu or a popup, where the second click would just cancel it out and trigger the right action. For anything else I think it would basically render it unusable. My personal use cases are covered either way (I only do "toggles" for now), but it would feel quite weird to have it "broken" by the virtue of always performing single click.

As for other software, I could only think of file managers where double clicks exist at all (open vs select, rename vs select, and in Nemo go a directory up vs unselect). In these cases, the semantics of actions always allow for this, and they don't expose any sort of configuration to this. Could you remind me of other examples?

I also don't really notice any delay with 250ms click interval, feels instant to me.

I wouldn't mind making it optional though. OTOH this could be an unnecessary complexity dump onto people casually reading the config guide — might be moved into something like `wiki/Advanced Configuration.md`.

postsolar:
From [kitty docs](https://sw.kovidgoyal.net/kitty/conf/#mouse-actions) (which offers freely configurable double/triple clicks):

> Note that the click and double click events have a delay of [click_interval](https://sw.kovidgoyal.net/kitty/conf/#opt-kitty.click_interval) to disambiguate from double and triple presses

So they just accept the delay. You prolly already got it from the code, but just in case note that in our implementation, if user only configured single clicks, and no double clicks, then single clicks happen instantly, there's no delay if we're not even waiting for a successive click.

JakeStanger:
Sorry for the delay - I think going based on Kitty behaviour makes the most sense, since it's the most applicable/similar scenario.

src/config/mod.rs
```diff
@@ -451,6 +490,48 @@ pub struct Config {
...
+    /// Use GTK's gtk-double-click-time setting
+    Keyword(DoubleClickTimeKeyword),
+}
+
+#[derive(Debug, Clone, Deserialize)]
+#[cfg_attr(feature = "extras", derive(schemars::JsonSchema))]
+#[serde(rename_all = "lowercase")]
+pub enum DoubleClickTimeKeyword {
+    Gtk,
 }
```
JakeStanger: There is no need to introduce an additional enum here - we can just create a Gtk variant directly.

```suggestion
    Gtk
```

src/config/mod.rs
```diff
@@ -451,6 +490,48 @@ pub struct Config {
...
+    /// Can be set to a number (e.g., `250`) or `"gtk"` to use GTK's setting.
+    ///
+    /// **Default**: `250`
+    #[serde(default = "default_double_click_time")]
+    pub double_click_time: DoubleClickTime,
+}
+
+fn default_double_click_time() -> DoubleClickTime {
+    DoubleClickTime::Ms(250)
+}
```
JakeStanger: The preference here would be to impl `Default` for `DoubleClickTime`

JakeStanger (changes requested):
Two minor bits then we're good to go

src/config/mod.rs
postsolar: I don't think it could work without a custom deserializer (which is a lot of new code). This specifically leads to config parsing failing when the value is set to `"gtk"`, only works with `null`. Is there something I missed, or should I just keep it the current way?

src/config/mod.rs
postsolar: Yeah, this is a dumb one, sorry :smile:

postsolar:
https://github.com/JakeStanger/ironbar/pull/1247/commits/fe99eaf60adbf66ad83677a520c8a35edcd8e3db is unrelated in terms of scope of the PR, but needed to pass clippy. Let me know if I should revert it :+1:

src/config/mod.rs
JakeStanger: Try including `untagged` in the enum serde attr

JakeStanger: Try including `untagged` in the enum serde attr

JakeStanger: After some digging, turns out you can put `untagged` on the last entry (entries?) in an enum rather than the entire enum, and this works as expected:

```rust
#[derive(Debug, Clone, Deserialize)]
#[serde(rename_all = "snake_case")]
pub enum DoubleClickTime {
    /// Use GTK's gtk-double-click-time setting
    Gtk,
    /// Milliseconds
    #[serde(untagged)]
    Ms(u64),
}
```

JakeStanger (approved):
gonna attempt to squash this down to the 3 main commits without git imploding and then we're good tog

JakeStanger:
The commits are a bit scuffed because they're fairly interdependent so it wasn't very easy to pick them apart, but the three commits are there and roughly contain the right code. 

Rebasing was nasty - I've checked and double checked the code to make sure it matches based on [this commit](https://github.com/JakeStanger/ironbar/tree/9afd568e515f160010b83c768c800369b102e3ea) just before the rebase. If you do spot anything wrong let me know.

JakeStanger:
Thanks for your work on this

src/config/mod.rs
postsolar: Oh, that's cool!  :exploding_head:

postsolar:
Thank you for your help and feedback! Happy to see it merged :partying_face:

PR #1250, opened 2025-11-23
feat: add marquee effect to volume popup's audio inputs

postsolar:
Part of #1011. Based on #1066 - GTK4 actually fixes that scrollbars bug, so pretty much no changes aside from updating to GTK4 and expanding what's configurable.

### Overview

- Wraps the label in a ScrolledWindow with hidden scrollbars
- Configurable max length before marquee starts
- Configurable scrolling speed
- Configurable inter-loop pause duration
- Configurable separator
- Optional pause on hover or play on hover

### Known issues

None so far.

### Limitations and future considerations

It's **opt-in so no breaking changes**, but anyone wishing to use might need to update their CSS (implementation details leak a bit):

```css
/* Before */
.volume .app-box > .title {
  ...
}

/* After: target both */
.volume .app-box > .title,
.volume .app-box > scrolledwindow > viewport > .title {
  ...
}

/* But 'any child' selector stays the same, works for both */
.volume .app-box .title {
  ...
}
```

For future extensions of other modules, even aside from just music (and its popup) I could see a lot of use-cases for making every Label scrollable like this. But this will necessarily mean replicating the same less-than-perfect UX pattern with regards to user CSS - wanted to hear your thoughts @JakeStanger on this.

### DIFF
--- a/docs/modules/Volume.md
+++ b/docs/modules/Volume.md
@@ -1,7 +1,7 @@
 Displays the current volume level.
 Clicking on the widget opens a volume mixer, which allows you to change the device output level,
 the default playback device, and control application volume levels individually.
-Use `truncate` option to control the display of application titles in the volume mixer.
+Use `truncate` or `marquee` options to control the display of application titles in the volume mixer.
 
 This requires PulseAudio to function (`pipewire-pulse` is supported).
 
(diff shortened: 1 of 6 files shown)


src/config/marquee.rs
```diff
@@ -0,0 +1,56 @@
...
+    #[serde(default)]
+    pub pause_on_hover: bool,
+
+    /// Whether to invert the pause on hover behavior.
+    /// When true, scrolling will only occur on hover.
+    /// This takes priority over `pause_on_hover`.
+    ///
+    /// **Default**: `false`
+    #[serde(default)]
+    pub play_on_hover: bool,
```
JakeStanger: Perhaps this can be reduced to an `on_hover: "play" | "pause" | "none"` option?

docs/modules/Volume.md
```diff
@@ -11,18 +11,26 @@ This requires PulseAudio to function (`pipewire-pulse` is supported).
...
+| `icons.volume_medium`           | `string`                                             | `󰖀`                    | Icon to show for medium volume levels.                                                                                            |
+| `icons.volume_low`              | `string`                                             | `󰕿`                    | Icon to show for low volume levels.                                                                                               |
+| `icons.muted`                   | `string`                                             | `󰝟`                    | Icon to show for muted outputs.                                                                                                   |
+| `truncate`                      | `'start'` or `'middle'` or `'end'` or `off` or `Map` | `off`                  | The location of the ellipses and where to truncate text from. Leave null to avoid truncating. Use the long-hand `Map` version if specifying a length. |
```
JakeStanger: Worth adding a clause here to say it's incompatible with `marquee`.

docs/modules/Volume.md
```diff
@@ -11,18 +11,26 @@ This requires PulseAudio to function (`pipewire-pulse` is supported).
...
+| `marquee.max_length`            | `integer`                                            | `null`                 | The maximum length of text (roughly, in characters) before it gets truncated and starts scrolling.                                |
+| `marquee.scroll_speed`          | `float`                                              | `0.5`                  | Scroll speed in pixels per frame. Higher values scroll faster.                                                                    |
+| `marquee.pause_duration`        | `integer`                                            | `5000`                 | Duration in milliseconds to pause at each loop point.                                                                             |
+| `marquee.separator`             | `string`                                             | `"    "`               | Separator string to place between the repeated text. Can be any string like `" • "` or `" \| "`.                                 |
```
JakeStanger: I don't think that last clause is needed. It should be implied from the fact that it's a string type and from the multi-character default that any string is supported.

```suggestion
| `marquee.separator`             | `string`                                             | `"    "`               | Separator string to place between the repeated text.                                 |
```

docs/modules/Volume.md
```diff
@@ -134,4 +142,26 @@ The following tokens can be used in the `format` config option:
...
+.volume .app-box > scrolledwindow > viewport > .title {
+  ...
+}
+
+/* But any child selector stays the same */
+.volume .app-box .title {
+  ...
+}
+```
+
```
JakeStanger: Although true, the internal widget tree is not really expected to be a stable interface, and there's not really any reason to use the child selector in this scenario. The class names are also incorrect.

This should be `.popup-volume .apps-box .app-box .title` as in the table above, which holds true in either case.

I'd argue this section can therefore go.

src/clients/volume/sink_input.rs
```diff
@@ -131,17 +131,27 @@ fn update(
...
+        }
     }
 
-    tx.send_expect(Event::UpdateInput(info.into()));
+    tx.send_expect(Event::UpdateInput(input_info.clone()));
+
+    // HACK: send `Remove` and `Add` events to force UI to recreate widget.
+    // This is needed to re-initialize scrolling animation.
+    tx.send_expect(Event::RemoveInput(input_info.index));
+    tx.send_expect(Event::AddInput(input_info));
```
JakeStanger: This will need to change prior to merge. The client should have no connection to how the data is rendered.

If recreating the widgets is definitely necessary, you can do that in the `UpdateInput` event handler in the module.

src/modules/volume.rs
```diff
@@ -46,6 +46,12 @@ pub struct VolumeModule {
     /// **Default**: `null`
     pub(crate) truncate: Option<TruncateMode>,
 
+    /// See [marquee options](module-level-options#marquee-mode).
+    ///
+    /// **Default**: `null`
+    #[serde(default)]
+    pub(crate) marquee: Option<MarqueeMode>,
```
JakeStanger: This would be cleaner without the `Option`, since you'd be able to check directly on `marquee.enable`.

Changing this would also require implementing `Default` on `MarqueeMode`, which would allow you to remove the field-level `#[serde(default)]` attrs.

src/gtk_helpers.rs
```diff
@@ -145,3 +152,158 @@ where
         snapshot.to_paintable(None)
     }
 }
+
+// Calculate pixel width of a string given the label it's displayed in
```
JakeStanger: Function and method comments should be in rustdoc format, and use 3rd person singular form (ie it should answer the question "what does it do?"):

```suggestion
/// Calculates pixel width of a string given the label it's displayed in.
```

src/gtk_helpers.rs
```diff
@@ -145,3 +152,158 @@ where
...
+
+pub fn create_marquee_widget(
+    label: &Label,
+    text: &str,
+    marquee_mode: MarqueeMode,
+) -> ScrolledWindow {
+    // Default constants
+    const DEFAULT_SCROLL_SPEED: f64 = 0.5; // pixels per tick
+    const DEFAULT_PAUSE_DURATION_MS: u64 = 5000; // 5 seconds
+    const DEFAULT_SEPARATOR: &str = "    "; // 4 spaces
```
JakeStanger: These defaults should be provided by the struct default impl and Serde. The fields should also not be `Option` types.

src/gtk_helpers.rs
```diff
@@ -145,3 +152,158 @@ where
...
+    let label = label.clone();
+    let text = text.to_string();
+
+    // Cache the original text width (calculated once upfront)
+    let original_text_width = pixel_width(&label, &text);
+
+    let is_hovered = Rc::new(RefCell::new(false));
+    let pause_started_at = Rc::new(RefCell::new(None::<Instant>));
+    let is_scrolling = Rc::new(RefCell::new(false));
+    let reset_at_cached = Rc::new(RefCell::new(None::<f64>));
```
JakeStanger: The `bool`s can probably use `Cell` rather than `RefCell` here. You might be able to for the `Option<f64>` too - not sure.

src/gtk_helpers.rs
```diff
@@ -145,3 +152,158 @@ where
...
+
+                // Calculate and cache reset position (where to loop back to)
+                let reset_at = pixel_width(&label, &format!("{}{}", &text, &sep)) as f64;
+                *reset_at_cached_clone.borrow_mut() = Some(reset_at);
+
+                *is_scrolling_clone.borrow_mut() = true;
+            }
+
+            // Use cached reset position
+            let reset_at = reset_at_cached_clone.borrow().unwrap();
```
JakeStanger: This should either be handled, or use `expect` to detail why this should never panic.

src/gtk_helpers.rs
```diff
@@ -145,3 +152,158 @@ where
...
+    let is_hovered = Rc::new(RefCell::new(false));
+    let pause_started_at = Rc::new(RefCell::new(None::<Instant>));
+    let is_scrolling = Rc::new(RefCell::new(false));
+    let reset_at_cached = Rc::new(RefCell::new(None::<f64>));
+
+    // Start a tick callback that checks size and scrolls if needed
+    let is_hovered_clone = is_hovered.clone();
+    let pause_started_at_clone = pause_started_at.clone();
+    let is_scrolling_clone = is_scrolling.clone();
+    let reset_at_cached_clone = reset_at_cached.clone();
```
JakeStanger: I'm not a fan of the `_clone` suffix - to me it doesn't immediately read that it's a clone of the values (eg `is_scrolling_clone` suggests it's a "scrolling clone". 

Generally I tend to use `2` as a suffix, or (preferably, where it doesn't cause stupid levels of indentation) stick the cloned values in their own block so that they can use the same name as the outer scope.

JakeStanger (changes requested):
Thanks for this. I've raised a few bits in the first pass - there's quite a lot going on here, so I imagine this'll need a few runs through before I catch everything.

Agreed this will be a good thing to be able to build into the `label` module and custom widget, as well as introduce into other modules where it makes sense.

RE the styling concern, I think it's a non-issue. I've addressed it in a comment against the documentation.

PR #1265, opened 2025-11-30
fix(launcher): close popup when mouse leaves

slowsage:
Currently when I have few windows open in firefox, it shows me options of windows to switch to when I mouseover. When mouse leaves (no click, move mouse down over entries, away from bar) its still there till I click on one of them.

Possibly related to #224.

### DIFF
--- a/src/modules/launcher/mod.rs
+++ b/src/modules/launcher/mod.rs
@@ -17,7 +17,7 @@ use crate::modules::launcher::pagination::{IconContext, Pagination};
 use crate::{arc_mut, lock, module_impl, rc_mut, spawn, write_lock};
 use color_eyre::Report;
 use gtk::prelude::*;
-use gtk::{Button, Orientation};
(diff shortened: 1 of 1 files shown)


JakeStanger:
Hey sorry for the delay on looking at this. This is a definite improvement over the current behaviour.

If you move the mouse out of the popup onto the parent button, the popup closes and then immediately opens again, which leads to a flicker. It should be possible to check and prevent this in your event handler, if you're able to have a look please.

JakeStanger:
I've been having a play with this myself, and it seems like it's going to be much more involved to resolve the flicker than I thought. Led to a bit of a thread unravelling situation that suggests the launcher and module update system as a whole could probably do with some attention.

I'll merge this as-is as it's certainly better than current behaviour and the flicker is a minor issue. If you do have anything in the works for the flickering it's still very welcome.

Issue #935, opened 2025-04-10
Improve upower (battery) icon handling


freijon:
The OOTB icon that is displayed with the `upower` module is not to my liking and looks very out of place with my setup.

# Improvement 1
If you really want to provide the icon OOTB, then it should be possible to disable/hide it. Today, I have to use  CSS where it seems I can only set `opacity: 0;` in combination with the setting `icon_size = 0` which uses the smallest icon possible and which hides it but the space is still occupied by the transparent icon.

# Improvement 2
I think it would be more consistent with other modules (e.g. [volume](https://github.com/JakeStanger/ironbar/wiki/volume#Configuration)) to have different strings for the icon, e.g.
- icons.battery_100
- icons.battery_75
- icons.battery_50
- icons.battery_25
- icons.battery_0

JakeStanger:
This is covered by #511, which is the next issue on my list. Should have something for you soon

freijon:
Now that this is apparently implemented, how would I achieve what I described in the original post?

JakeStanger:
I'm looking and I think this was unfairly closed, and a proper solution still needs to be put in place for this module.

The `thresholds` property lets you control CSS classes, and the module now uses the proper icon system, which means you *can* change icons to other image using the top-level [`icon_overrides`](https://github.com/JakeStanger/ironbar/wiki/configuration-guide#31-top-level-options) option, but this isn't ideal.

See #1272 for tracking the ability to hide the icon too.

PR #1274, opened 2025-12-07
Centralised thresholds system

JakeStanger:
## Overview

This is a first draft at attempting to introduce a standardised system for configuring thresholds, ie for icon boundaries.

The system would apply to the following modules:

- `backlight` (configure brightness icon, #1235)
- `network_manager` (configure wifi strength icon, #1233)
- `volume` (configure volume icon)

The following locations would also expand to use this with future updates:

- `battery` (configure battery icon, #935)
- `music` (configure volume icon)

There may also be places I've missed.

## Syntaxes

This introduces three separate syntax for configuring the thresholds in an attempt to appease everybody:

### Basic

Offers pre-defined low/medium/high keywords that abstract the thresholds. The thresholds are always divided into equal thirds from `0-max`

```corn
icons.volume.low = "icon:volume_low"
icons.volume.medium = "icon:volume_medium"
icons.volume.high = "icon:volume_high"
```

### Dynamic

Uses an array of dynamic length to define N number of thresholds, all equally divided from `0-max`. Keys are provided in order from low-high.

```corn
icons.volume = [ "icon:volume_low" "icon_volume_medium" "icon_volume_high" ]
```

### Manual

Uses a map to define N number of thresholds, manually specifying the lower bound for each threshold. This allows for more explicit configuration, and non-linear thresholds.

A zero threshold must be defined.

```corn
icons.volume.0 = "icon:volume_low"
icons.volume.33 = "icon:volume_medium"
icons.volume.66 = "icon:volume_high"
```

## API

The struct would be placed inside the config:

```rust
struct Icons {
  volume: Thresholds<String>
}
```

Currently a single `threshold_for` method is provided:

```rust
let icon = icons.volume.threshold_for(curr_volume, max_volume); // Option<String>
```

## Outstanding

- The `manual` format assumes integers are used as `HashMap` keys. This may not be supported by all formats.
- The naming is a little all over the place (level, threshold, value, ...). This needs improving.

Questions:

- Is three separate syntaxes too many? Could this introduce confusion more than it creates benefit?
  - If not, which one is the preference?
- Do we need to support negative values?
- How do we better deal with the zero value in the map syntax?
- How do we handle non-linear implementations (ie network manager)?
- This conflicts with the `battery` module existing thresholds option. What to do there?

--- 

I am looking for feedback on this! This is a bit of a "throw everything at the wall and see what sticks" approach. Anything regarding anything I've thought of above, or anything else you see.

Paging @xMAC94x @Rodrigodd as both of you have open PRs pending a decision around this - feedback from the both of you on whether you think this would work for your modules would be especially valuable.

### DIFF
--- a/docs/modules/Volume.md
+++ b/docs/modules/Volume.md
@@ -11,25 +11,26 @@ This requires PulseAudio to function (`pipewire-pulse` is supported).
 
 > Type: `volume`
 
-| Name                            | Type                                                 | Default                | Description                                                                                                                       |
-|---------------------------------|------------------------------------------------------|------------------------|-----------------------------------------------------------------------------------------------------------------------------------|
-| `format`                        | `string`                                             | `{icon} {percentage}%` | Format string to use for the widget button label.                                                                                 |
-| `max_volume`                    | `float`                                              | `100`                  | Maximum value to allow volume sliders to reach. Pulse supports values > 100 but this may result in distortion.                    |
-| `icons.volume_high`             | `string`                                             | `󰕾`                    | Icon to show for high volume levels.                                                                                              |
-| `icons.volume_medium`           | `string`                                             | `󰖀`                    | Icon to show for medium volume levels.                                                                                            |
-| `icons.volume_low`              | `string`                                             | `󰕿`                    | Icon to show for low volume levels.                                                                                               |
-| `icons.muted`                   | `string`                                             | `󰝟`                    | Icon to show for muted outputs.                                                                                                   |
-| `truncate`                      | `'start'` or `'middle'` or `'end'` or `off` or `Map` | `off`                  | The location of the ellipses and where to truncate text from. Leave null to avoid truncating. Use the long-hand `Map` version if specifying a length. Takes precedence over `marquee` if both are configured. |
-| `truncate.mode`                 | `'start'` or `'middle'` or `'end'` or `off`          | `off`                  | The location of the ellipses and where to truncate text from. Leave null to avoid truncating.                                     |
-| `truncate.length`               | `integer`                                            | `null`                 | The fixed width (in chars) of the widget. Leave blank to let GTK automatically handle.                                            |
-| `truncate.max_length`           | `integer`                                            | `null`                 | The maximum number of characters before truncating. Leave blank to let GTK automatically handle.                                  |
-| `marquee`                       | `Map`                                                | `false`                | Options for enabling and configuring a marquee (scrolling) effect for long text. Ignored if `truncate` is configured.             |
-| `marquee.enable`                | `bool`                                               | `false`                | Whether to enable a marquee effect.                                                                                               |
-| `marquee.max_length`            | `integer`                                            | `null`                 | The maximum length of text (roughly, in characters) before it gets truncated and starts scrolling.                                |
-| `marquee.scroll_speed`          | `float`                                              | `0.5`                  | Scroll speed in pixels per frame. Higher values scroll faster.                                                                    |
-| `marquee.pause_duration`        | `integer`                                            | `5000`                 | Duration in milliseconds to pause at each loop point.                                                                             |
-| `marquee.separator`             | `string`                                             | `"    "`               | String displayed between the end and beginning of text as it loops.                                                               |
-| `marquee.on_hover`              | `'none'` or `'pause'` or `'play'`                    | `'none'`               | Controls marquee behavior on hover: `'none'` (always scroll), `'pause'` (pause on hover), or `'play'` (only scroll on hover).     |
+| Name                     | Type                                                 | Default                           | Description                                                                                                                                                                                                   |
+|--------------------------|------------------------------------------------------|-----------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `format`                 | `string`                                             | `{icon} {percentage}%`            | Format string to use for the widget button label.                                                                                                                                                             |
+| `max_volume`             | `float`                                              | `100`                             | Maximum value to allow volume sliders to reach. Pulse supports values > 100 but this may result in distortion.                                                                                                |
+| `icons.volume`           | `Map<i32, string>`                                   | `{ 0 = "󰕿" 33 = "󰖀" 67 = "󰕾"}` | Map of icons to show above each defined volume level.                                                                                                                                                         |
+| `icons.volume_high`      | `string`                                             | `󰕾`                              | Icon to show for high volume levels.                                                                                                                                                                          |
+| `icons.volume_medium`    | `string`                                             | `󰖀`                              | Icon to show for medium volume levels.                                                                                                                                                                        |
+| `icons.volume_low`       | `string`                                             | `󰕿`                              | Icon to show for low volume levels.                                                                                                                                                                           |
+| `icons.muted`            | `string`                                             | `󰝟`                              | Icon to show for muted outputs.                                                                                                                                                                               |
+| `truncate`               | `'start'` or `'middle'` or `'end'` or `off` or `Map` | `off`                             | The location of the ellipses and where to truncate text from. Leave null to avoid truncating. Use the long-hand `Map` version if specifying a length. Takes precedence over `marquee` if both are configured. |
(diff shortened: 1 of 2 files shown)


xMAC94x:
IMO those 3 options are probably too many. I would prefer the `manual` approach as it offers the most flexibility. Tbh, `ironbar` is configured once and it's not a huge effort for people to configure it with the `manual` approach.
Even more important: having a good documentation and have many examples for 1 option makes it far easier to understand, and people can just copy past a template with 3 values (0, 33, 66) to support the `Basic` mode.

Instead of blindly taking the first rfind key that is slower than value, we might think about allowing multiple `Interpolation` modes, e.g. `floor` and `exponential`. where it decides that `43` within `[25, 36, 49]` fits in the bucket `36`. though bth, when going manual, it can all be calculated by people that really care about it and we dont need to have it in the config at all.

Its okay to have a breaking change with `0.19` when this comes. We should announce it at the release page somewhere, put it in the PR and maybe even put a warn box to the docu of the module, that the behavior changed. People will pick it up and adjust their config. thats fine for a major change version :)

There are a few more things to consider:
 - we want to have thresholds for sysinfo module.
 - maybe switch to BTreeMap:  https://doc.rust-lang.org/std/collections/index.html#use-a-btreemap-when
 - I would prob go with i32, it shouldn't matter to much.
 - Allow the system to be used for css access somehow.
 
 Edit: Maybe some way to determine the value via code would be awesome, e.g. specifying a script file or allow wasm. but that might be overengineered for now

Rodrigodd:
In the case of the network-manager module, the default way of configuring it would be the manual method, using the thresholds that I copied from NetworkManager, replacing that strange function I made to map strength to level.

I see value in the dynamic method, in cases where a huge list of icons might be passed and the user doesn't want to tweak all threshold values whenever they want to add or remove an item. But I'm not sure if this will happen often. It may also not map well to the signal strengths NetworkManager provides, but I'm not certain, that's why I copied them directly from NetworkManager.

The basic method sounds too specialized to me, but I see the value in case you want to migrate some modules to the new system while maintaining compatibility.

Maybe there should be a script variant if handling more complex cases ever becomes necessary? This would require adding support for passing arguments to scripts, though.

JakeStanger:
Thanks for the feedback both. Sounds like at a minimum I'll do away with the basic mode. My own thoughts below.

> In the case of the network-manager module, the default way of configuring it would be the manual method

My concern here is that you would always want the non-linear logic - I'd rather keep that integrated into the code and not in the config. 

The problem I see otherwise (and I guess this applies generally to the manual approach) is that if I wanted to change one of the icons for example, I now need to find which threshold value it currently uses and set that. If it's linear that is at least more predictable. Perhaps for network manager specifically we can do something similar to what you're doing and convert between the linear scale and NM's scale.

> Instead of blindly taking the first rfind key that is slower than value, we might think about allowing multiple Interpolation modes, e.g. floor and exponential

This sounds like it complicates things too much to me. The current "floor" approach feels like the most predictable/understandable approach to configure.

> we want to have thresholds for sysinfo module.

This sounds like a separate feature request, even if it does integrate with this. That can be achieved via external scripting already too.

> Allow the system to be used for css access somehow.

Not sure what the ask is here, but again sounds like a separate request & needs fleshing out.

> Maybe some way to determine the value via code

> Maybe there should be a script variant if handling more complex cases ever becomes necessary?

I think that may be too complex for now, but I'm open to adding it in the future. I'm just not seeing how it would work at a technical level (oneshot script triggered on event, some new dynamic string type thing?). With the long-term goal of adding a Lua API this should work nicely.

---

I'm not too worried about low-level implementation details like the map type or performance concerns yet - I'm just prototyping currently.

xMAC94x:
I am definitely fine with many things beeing part of a other feature request. I just want to encourage you to think at those when designing this API, so that we might be able to extend the config in the future and stay backwards compatible :)

Rodrigodd:
> My concern here is that you would always want the non-linear logic.

My concern is that I not sure if I know any better than the user what the mapping should be (although that sound like "evasion of responsibility"). For example, I just check gnome-shell, and there it uses another mapping (maybe my module should be based on this one, as it also uses GTK icons by default): https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/449a4e354af82a2412cfb1ee2fe26451631aeae6/js/ui/status/network.js#L46-57

> The problem I see otherwise (and I guess this applies generally to the manual approach) is that if I wanted to change one of the icons for example, I now need to find which threshold value it currently uses and set that.

I had think of putting the current default in the example configuration in the docs, so when changing the icons the user would only need to copy and paste from the example, and change them. But yes, changing the number of icons would be more annoying.

JakeStanger:
I've reworked the approach based on the feedback above into something much simpler:

- Only the "manual" format remains. I think if a great enough need for the dynamic approach appears, we can add it back. For now though, I think use-cases are limited. Agreed that one syntax is better than several.
- I've switched the map out for a `i32` instead of `u32` to allow negative thresholds, in case that ever comes up.
- The `threshold_for` method is now called `for_value` which I think makes more sense, although I'm still open to better ideas on that.
- There's a constructor for creating "basic" values, which is useful for defaults.
- I've switched the volume module over to the new system as a test/example

src/config/thresholds.rs
```diff
@@ -0,0 +1,156 @@
...
+        map.insert(67, high);
+
+        Self(map)
+    }
+
+    /// Gets the defined threshold value for the given numeric value.
+    ///
+    /// This is a 'floor' operation,
+    /// meaning the nearest threshold *below* the provided value is used.
+    pub fn for_value(&self, value: f64) -> Option<&T> {
```
xMAC94x: any reason to have the `f64` here in the interface ? I would prob keep it `i32`.

Other modules might come up with other calculation methods than floating point.

(Or we make this a template function, but i think thats overkill)

xMAC94x:
added a single technical detail comment, but it looks pretty IMO. consider it approved from my side :)

JakeStanger:
I'm still not happy with this. Looking at the volume icons config example:

```json
"icons": {
  "volume": {
    "0": "󰕿",
    "33": "󰖀",
    "67": "󰕾"
  },
  "muted": "󰝟"
}
```

This feels ugly and unintuitive to me. It's not as clear as it should be what those numbers mean. 

I've toyed with trying to merge the basic format in (so you could still use low/medium/high as aliases alongside the numbers), but I can't get Serde to work with it. Not sure where to go from here.

I'm leaning towards reintroducing the dynamic array format as the "preferred" format, and allowing the manual format for those who do want the extra control.

src/config/thresholds.rs
JakeStanger: I think my original thinking was most of the clients would be passing in `f64`, but it doesn't make a lot of sense to hide the cast inside here. Will change that.

xMAC94x:
> I'm still not happy with this. Looking at the volume icons config example:
> 
> ```json
> "icons": {
>   "volume": {
>     "0": "󰕿",
>     "33": "󰖀",
>     "67": "󰕾"
>   },
>   "muted": "󰝟"
> }
> ```
> 
> This feels ugly and unintuitive to me. It's not as clear as it should be what those numbers mean.
> 
> I've toyed with trying to merge the basic format in (so you could still use low/medium/high as aliases alongside the numbers), but I can't get Serde to work with it. Not sure where to go from here.
> 
> I'm leaning towards reintroducing the dynamic array format as the "preferred" format, and allowing the manual format for those who do want the extra control.

how would that system looks in serde?

If you want to allow both:

{
 "number": "string"
  ..
 }
 
 and
 {
  "low": "string",
  "medium": string",
  "high": "string",
 }
 
 you prob need to implement some custom serde traits. I have never done that though.
 
 I kinda agree, it's not a self explanatory format. But thats where documentation and comments come into play. And once the first thing is configured its easy to apply elsewhere. 
 
 Written like this:
 ```json
"icons": {
  "volume_thresholds": {
     "0": "󰕿",
     "33": "󰖀",
     "67": "󰕾"
  },
}
```
or this
 ```json
"icons": {
  "volume": {
    "thresholds": {
       "0": "󰕿",
       "33": "󰖀",
       "67": "󰕾"
    },
  },
}
```

it would be quite self-explonatory again

JakeStanger:
I've re-engineered and reversed the logic on the problem and come up with an entirely different approach in #1349. It's more involved to implement but it presents immediate benefits and I'm much happier with the configuration syntax than I am this.

Thanks for the help and feedback. Closing this in favour of the new PR. Will review and refine that and hope to get it shipped soon-ish.

PR #1281, opened 2025-12-17
Fix CSS style for minimal example

esther-ff:
Adjusted the style so it looks the same regardless of gtk theme.

### DIFF
--- a/examples/minimal/style.css
+++ b/examples/minimal/style.css
@@ -19,11 +19,11 @@
 
 #bar, popover, popover contents, calendar {
     background-color: var(--color-dark-primary);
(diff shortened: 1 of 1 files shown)


examples/minimal/style.css
JakeStanger, line 25: Does adding the `color` line here work? This would in theory set it across all modules rather than needing a per-module rule

esther-ff: Adding `color` to the `box, button, label` section works the same as duplicating it across individual modules, i will quickly patch that up.

JakeStanger (approved):
Thanks!

PR #1284, opened 2025-12-18
feat: music marquee

### LINKED ISSUE #1011: Truncate (or rotate) player metadata in volume and music popups

postsolar:
Works the same as #1250 just moves around its code a bit and makes music module use it, too. 

One question I'm not sure about is whether for icon-prefixed labels we want the icon to scroll as well. Currently it doesn't.

Closes #1011 

---

I think from here it would be relatively easy to extend it to just about any module, but I chose to move slower so as to:

- Collect user feedback over a few weeks maybe
- Consider design questions

E.g. music is obvious for my use case, and, say, Label or Script are obvious because they are meant to be as flexible as it gets, but outside of these it's not immediately apparent to me what should or should not be done :)

### DIFF
--- a/src/image/gtk.rs
+++ b/src/image/gtk.rs
@@ -1,4 +1,4 @@
-use crate::gtk_helpers::IronbarLabelExt;
+use crate::gtk_helpers::{IronbarLabelExt, OverflowLabel};
 use crate::image;
 use gtk::prelude::*;
 use gtk::{Button, ContentFit, Image, Label, Orientation, Picture};
@@ -251,7 +251,7 @@ impl Deref for IconLabel {
 #[cfg(feature = "music")]
 pub struct IconPrefixedLabel {
     label: Label,
-    overflow: Option<crate::gtk_helpers::OverflowLabel>,
+    overflow: Option<OverflowLabel>,
     container: gtk::Box,
 }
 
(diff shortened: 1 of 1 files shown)


src/modules/music/mod.rs
```diff
@@ -183,18 +183,18 @@ impl Module<Button> for MusicModule {
...
-            label.truncate(truncate);
-        }
+        let label = OverflowLabel::new(
+            Label::builder()
+                .use_markup(true)
+                .justify(self.layout.justify.into())
+                .build(),
+            self.truncate,
+            self.marquee.clone(),
+        );
```
JakeStanger: I'd prefer the initial label separated here, just to make it a bit more readable:

```suggestion
        let label = Label::builder()
          .use_markup(true)
          .justify(self.layout.justify.into())
          .build();
        
        let label = OverflowLabel::new(
            label,
            self.truncate,
            self.marquee.clone(),
        );
```

Happy to leave the popup ones as-is, as they're only one-liners.

src/image/gtk.rs
```diff
@@ -248,39 +248,60 @@ impl Deref for IconLabel {
     }
 }
 
-#[derive(Clone, Debug)]
 #[cfg(feature = "music")]
 pub struct IconPrefixedLabel {
     label: Label,
+    overflow: Option<crate::gtk_helpers::OverflowLabel>,
```
JakeStanger: Would rather this didn't use the fully qualified path, happy to import:

```suggestion
    overflow: Option<OverflowLabel>,
```

Same in one or two cases below

PR #1288, opened 2025-12-20
fix(nix): don't use hostPlatform.system alias

imnotpoz:
this breaks when `nixpkgs.config.allowAliases` is set to true


JakeStanger:
Thanks!

Issue #942, opened 2025-04-18
Ability to sort workspaces by Id


hattivatt:
**Is your feature request related to a problem? Please describe.**
Currently Workspaces module provides only three variants of sorting Name, Label and Added. None of this options are convenient for people like me, who wants their workspaces always had same order, and also want to name it freely, without any numbers.

**Describe the solution you'd like**
Implement ability to sort workspaces by workspace_id. Similar to waybar

JakeStanger:
I'm not sure I understand this. Why does the `name` option not work for you?

It sounds like you're asking to sort on the internal id, which is not consistently the same.

hattivatt:
Well, I guess it's about internal Id...
I use hyprland and I bind switching between workspaces to number keys. Like Mod + 1 go to workspace 1, Mod + 2 go to workspace 2, etc...  And those binds made by workspace ids. And for names I just use hyprland-autoname-workkspaces to make workspace names show icons of apps opened on this workspace.
Right now I'm using waybar and it has option to sort workspaces by id, but now I'm trying to replicate this setup on ironbar and came to this 

Basically it looks like this (waybar on top, ironbar lower) for now it's in order (workspaces 1, 2, 3, 9)
![Image](https://github.com/user-attachments/assets/cdb4634e-de8f-4ad4-b210-576b31ad03db)

But when I want to switch for let's say workspace 6, now empty, waybar put's it correctly between 3 and 9, but ironbar just put it first and that's it
![Image](https://github.com/user-attachments/assets/ae84fc29-44b3-426c-a1a0-ec3f5610b0c9)

Of course it's not a huge problem, but it feels too weird for me to, considering that I have workspaces binded to number keys and got used to this order.

JakeStanger:
Right, I see the issue. 

I don't really like the idea of relying on the IDs, because it is a compositor-specific implementation detail that will only work on Hyprland; Sway and Niri just use the IDs as an internal unique value, and there's no guarantee the same workspace number will always have the same name. That'll change with `ext-workspace` support again.

I am struggling to see another option though...

I'd be open to adding a sort of "manual" sort option, where you can stick the desired order in the config. Issue is, I can't see how that'd work either because again it needs some sort of identifier.

Any other ideas, I'm open.

hattivatt:
I see, so waybar can do it because it has different workspaces module for different compositors... I can see the problem here, but sadly I'm too don't have any ideas

deathtrip:
[...]

PR #1299, opened 2025-12-23
feat(workspaces): add index sort order and custom format options

jbms:
Workspaces can now be sorted by `index`, in order to properly support niri.

The `format_named` and `format_unnamed` configuration options can be used to
specify a custom label for the workspace buttons.

Fixes https://github.com/JakeStanger/ironbar/issues/942

### DIFF
--- a/docs/modules/Workspaces.md
+++ b/docs/modules/Workspaces.md
@@ -11,14 +11,29 @@ Shows all current workspaces. Clicking a workspace changes focus to it.
 
 > Type: `workspaces`
 
-| Name           | Type                                        | Default | Description                                                                                                                                                               |
-|----------------|---------------------------------------------|---------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| `name_map`     | `Map<string, string or image>`              | `{}`    | A map of actual workspace names to their display labels/images. Workspaces use their actual name if not present in the map. See [here](images) for information on images. |
-| `favorites`    | `Map<string, string[]>` or `string[]`       | `[]`    | Workspaces to always show. This can be for all monitors, or a map to set per monitor.                                                                                     |
-| `hidden`       | `string[]`                                  | `[]`    | A list of workspace names to never show                                                                                                                                   |
-| `icon_size`    | `integer`                                   | `32`    | Size to render icon at (image icons only).                                                                                                                                |
-| `all_monitors` | `boolean`                                   | `false` | Whether to display workspaces from all monitors. When `false`, only shows workspaces on the current monitor.                                                              |
-| `sort`         | `'added'`, `'index'`, `'label'` or `'name'` | `label` | The method used for sorting workspaces. `added` always appends to the end, `index` sorts by workspace index, `label` sorts by displayed value, and `name` sorts by workspace name.                          |
+| Name             | Type                                        | Default     | Description                                                                                                                                                               |
+|------------------|---------------------------------------------|-------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `name_map`       | `Map<string, string or image>`              | `{}`        | A map of actual workspace names to their display labels/images. Workspaces use their actual name if not present in the map. See [here](images) for information on images. |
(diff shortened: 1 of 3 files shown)


src/modules/workspaces/mod.rs
```diff
@@ -158,32 +189,59 @@ pub struct WorkspaceItemContext {
...
                     .and_then(ButtonExt::label)
                     .unwrap_or_else(|| child.widget_name())
-            } else {
-                child.widget_name()
-            }
-            .to_string();
+                    .to_string(),
+                SortOrder::Index => {
+                    let index =
+                        unsafe { child.data::<i64>("workspace_index").map(|v| *v.as_ref()) };
```
JakeStanger: You can use the `set_tag` and `get_tag` extension methods for getting/setting data throughout:

https://github.com/JakeStanger/ironbar/blob/25d6e95578085ef1422bf65427740fdedca37356/src/gtk_helpers.rs#L25-L29

The getter handles calling `as_ref` for you which should help simplify the code a bit.

docs/modules/Workspaces.md
```diff
@@ -11,14 +11,16 @@ Shows all current workspaces. Clicking a workspace changes focus to it.
...
+| `hidden`         | `string[]`                                  | `[]`        | A list of workspace names to never show                                                                                                                                   |
+| `icon_size`      | `integer`                                   | `32`        | Size to render icon at (image icons only).                                                                                                                                |
+| `format_named`   | `string`                                    | `"{label}"` | The format string for named workspaces. Supports `{label}`, `{name}`, and `{index}`.                                                                                      |
+| `format_unnamed` | `string`                                    | `"{label}"` | The format string for unnamed workspaces. Supports `{label}`, `{name}`, and `{index}`.                                                                                    |
```
JakeStanger: Can the formatting tokens be added to their own table please, to be consistent with other module documentation?

src/modules/workspaces/mod.rs
```diff
@@ -127,6 +130,28 @@ pub struct WorkspacesModule {
...
+    /// The format string for unnamed workspaces.
+    ///
+    /// The following placeholders are supported:
+    /// - `{label}`: The display label (from `name_map` or the workspace name).
+    /// - `{name}`: The actual workspace name.
+    /// - `{index}`: The workspace index.
+    ///
+    /// **Default**: `"{label}"`
+    #[serde(default = "default_format")]
+    format_unnamed: String,
```
JakeStanger: Since in most cases if you want to change the format, you'd likely want to change it for both workspace types, I think this would benefit from being restructured into a single `format` option represented by an enum.

This would take either a string (setting the format for both cases), or a struct with `.named` and `.unnamed` variants, allowing either one to be set individually if desired.

JakeStanger (changes requested):
Thanks for this. I think this should play nicely with `ext-workspaces` too. Just a few small bits

jbms:
Thanks for the feedback, I have addressed all of your comments.

docs/modules/Workspaces.md
```diff
@@ -11,14 +11,27 @@ Shows all current workspaces. Clicking a workspace changes focus to it.
...
+| `favorites`    | `Map<string, string[]>` or `string[]`       | `[]`        | Workspaces to always show. This can be for all monitors, or a map to set per monitor.                                                                                     |
+| `hidden`       | `string[]`                                  | `[]`        | A list of workspace names to never show                                                                                                                                   |
+| `icon_size`    | `integer`                                   | `32`        | Size to render icon at (image icons only).                                                                                                                                |
+| `format`       | `string` or `object`                        | `"{label}"` | The format string for the workspace button. Can be a string to apply to all, or an object with `named` and `unnamed` keys. See [formatting](#formatting) for more info.   |
```
JakeStanger: I think the two different formats need to be more clearly documented. Check the `interval` property for [sysinfo](https://github.com/JakeStanger/ironbar/wiki/sys-info#configuration) for how I've done this elsewhere - each map key is explicitly listed (similar to how you had it in the previous iteration, just with the top-level option also documented)

jbms:
Thanks, I fixed the documentation.

JakeStanger:
Sorry I'm not seeing the requested change. Can you double check please?

There's also a couple of clippy warnings that need fixing.

JakeStanger (approved):
Brilliant thanks

PR #1302, opened 2025-12-25
fix(music): fix seek bar interaction and freezing in popup

### LINKED ISSUE #1300: `music` seekbar ain't seeking
**Describe the bug**
the seek bar of the music module doesn't work and freezes when touched, until ironbar is reloaded

**To reproduce**

1. Add module `music`
2. Click popup
3. Move the seek bar

**Expected behavior**
the seekbar seeks through the audio track without freezing

**System information:**
- Distro: Arch Linux
- Compositor: [scroll](https://github.com/dawsers/scroll)
- Ironbar version: 0.19.0-pre

**Configuration**

<details><summary>sample config (YAML)</summary>

```
position: "bottom"
height: 20
layer: "top"
start:
  - type: music
    player_type: "mpris"
    truncate:
      mode: "end"
      max_length: 25
```

</details>

<details><summary>Styles</summary>
default styles

</details>

**Additional context**
happens in both mpris and mpd modes

**Screenshots**

https://github.com/user-attachments/assets/acfc261d-5605-4f31-91e9-e234a4282188

postsolar:
Closes #1300 

---

Replace incorrect GestureClick pattern with proper GTK4 Scale handling:
- Use connect_value_changed instead of manual gesture handlers
- Check has_css_class("dragging") instead of manual drag_lock _(note: same as Volume module does)_
- Only update progress when not dragging to prevent conflicts

This fixes two issues:
1. Seeking now works for both clicking and dragging
2. Progress bar and duration labels no longer freeze after seeking

The previous implementation used a manual AtomicBool lock that could get stuck, causing the UI to freeze. GTK4 automatically manages the "dragging" CSS class, making manual gesture handling unnecessary.

### DIFF
--- a/src/modules/music/mod.rs
+++ b/src/modules/music/mod.rs
@@ -2,14 +2,13 @@ use std::borrow::Cow;
 use std::cell::RefMut;
 use std::path::PathBuf;
 use std::sync::Arc;
-use std::sync::atomic::{AtomicBool, Ordering};
 use std::time::Duration;
 
 use color_eyre::Result;
 use glib::Propagation;
 use gtk::gdk::Paintable;
 use gtk::prelude::*;
-use gtk::{Button, ContentFit, EventSequenceState, GestureClick, Label, Orientation, Scale};
+use gtk::{Button, ContentFit, Label, Orientation, Scale};
 use tokio::sync::mpsc;
 use tracing::{error, warn};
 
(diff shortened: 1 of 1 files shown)


JakeStanger:
> Only update progress when not dragging to prevent conflicts

How necessary is this point? The live seeking behaviour is preferable if there's a reasonable way to achieve it without killing the server

postsolar:
It was the same before (`if !drag_lock.load(Ordering::Relaxed)`) and I aimed to keep this PR lean, but yeah could be removed I think.

Currently it's kinda half-way there, if I stop dragging but don't release the slider it would play for up to a second and then reset to the value of the slider.

If I understand you correctly, your (and mine) preference would be for it to keep playing until I move the slider again? And if I release it, it should continue playing? In other words, seek only on drag and move, but not on drag and stop/release.

postsolar:
On the other hand, I checked every other player I could think of in web and on iOS and each of them is actually way dumber – just pauses upon drag and until the release.

JakeStanger (approved):
I think based on that, this behaviour is good enough. Code lgtm, cheers

postsolar:
Thanks. This does provide an immediate fix to the issue, but to clarify I'm not a big fan of current behavior, i.e. looping on where the slider is held, feels janky. I think we should either go with the more complex behavior I described above (might be challenging) or the dumber way (just pause until release like every other player I observed).

JakeStanger:
Pause until release sounds reasonable enough to me I think, especially if that matches other players

PR #1306, opened 2025-12-26
fix(music): better seek behavior

postsolar:
Just seek on drag+release. This matches gtk3 ironbar and about every other player UI I tried. Also update elapsed/total duration according to the slider value when dragged.


postsolar:
Follow up to #1302 and its discussion thread

Issue #1222, opened 2025-10-27
Tray: duplicate nm-applet icons


MagneFire:
**Describe the bug**
Under some circumstances a duplicate nm-applet icon is shown. This does not happen every time.

**To reproduce**

1. Ensure the tray module is enabled
2. Ensure the `nm-applet` is running
3. Start bar
4a. If you do not observe duplicate icons, then kill the bar and go to step 3.
4b. Observe duplicate icons

**System information:**
- Distro: Arch Linux
- Compositor: Sway
- Ironbar version: git-2159aac

**Configuration**

<details><summary>Config</summary>

```
let {
    $tray = {
        type = "tray"
        class = "module module-tray"
        icon_size = 24
        direction = "h"
    }
}


in {
    position = "top"
    height = 1
    layer = "top"
    icon_theme = "Papirus"

    start = [  ]
    center = [ $tray ]
    end = [ ]
}

```

</details>

**Additional context**
Looking through the log the following seems to occur twice among things:
```
[:1.14/org/ayatana/NotificationItem/nm_applet/Menu] getting initial menu
```
These lines do not show twice if there's no duplicates.

It's worth noting that only one of the tray icons has a context menu, the other shows an empty drop down.

<details><summary>Log</summary>
```
2025-10-27T21:03:28.878299Z ERROR ironbar::config: 539: styles at 'examples/style.css' not found, falling back to minimal theme
2025-10-27T21:03:28.881553Z DEBUG ironbar::desktop_file: 178: resolved 484 files
2025-10-27T21:03:28.881596Z  INFO ironbar: 161: Ironbar version 0.18.0-pre
2025-10-27T21:03:28.881621Z  INFO ironbar: 162: Starting application
2025-10-27T21:03:28.882971Z DEBUG sctk: 463: Bound new global [55] wl_output v4
2025-10-27T21:03:28.882984Z DEBUG sctk: 503: Bound new global [8] zxdg_output_manager_v1 v3
2025-10-27T21:03:28.882997Z DEBUG sctk: 463: Bound new global [51] wl_seat v9
2025-10-27T21:03:28.883005Z DEBUG ironbar::clients::wayland::wlr_foreign_toplevel::manager: 25: Bound to ZwlForeignToplevelManagerV1 global
2025-10-27T21:03:28.883011Z DEBUG ironbar::clients::wayland::wlr_data_control::manager: 29: Bound to ZwlDataControlManagerV1 global
[...]

PR #1318, opened 2026-01-01
fix(tray): Handle tray race + Event::Remove. Maybe: #1222

slowsage:
1. Fixed UI Duplicate Bug: src/modules/tray/mod.rs to handle duplicate Event::Add messages from snapshot + subscription.
2. Fixed Client Memory Leak: src/clients/tray.rs process Event::Remove.

Possible fix for #1222 - hard to say since I see different manifestation

### DIFF
--- a/src/clients/tray.rs
+++ b/src/clients/tray.rs
@@ -61,6 +61,9 @@ impl Client {
                                 entry.menu = Some(menu.clone());
                             }
                         }
+                        Event::Remove(address) => {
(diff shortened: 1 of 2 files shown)


JakeStanger:
Does the "maybe" still hold true after the push earlier, or are you confident this resolves the issue?

slowsage:
Not confident. Would leave it open and have users who faced it confirm the fix once pushed.

PR #1319, opened 2026-01-02
fix(inhibit): Replace infinity symbol with nerd font for readability

slowsage:
The unicode symbol is too small in hindsight.


JakeStanger (approved):
Sounds good, cheers

PR #1344, opened 2026-01-25
feat: Access ironbar variables from cairo/lua

untoldwind:
Added the following functions/fields to lua:
* `ironbar:var_get(key)`: Lookup an ironbar variable, if `key` is a namespace all variables in that namespace are returned as a lua-table
* `ironbar:log_debug(msg)`,  `ironbar:log_info(msg)`, ...: Just some logging helpers (better than using `print` in lua.
* `ironbar:unixtime()`: Helper for high resolution unixtime. With this the `day_ms` in the clock example can be done without running an executable:
  ```lua
   function get_ms()
     return ironbar:unixtime().subsec_millis / 1000
   end
  ```
* `ironbar.config_dir`: This makes it simpler to do relative file import via `dofile` in lua
* Also added the width/height of the drawing area to the draw function.

Background for all of this:
I was trying to create a simple barchart for cpu load per core.
With the changes/additions this can be done with a rather simple script:
```lua
function draw(cr, area_width, area_height)
    local num_cores = 32 -- there is probably a better way than to hardcode this
    local bar_width = area_width / num_cores
    local bar_height = area_height - 2
    local cpu_percent = ironbar:var_get("sysinfo.cpu_percent")

    cr:set_source_rgb(1.0, 1.0, 1.0)

    for i = 0, num_cores - 1 do
      local core_percent = cpu_percent["cpu" .. i]
      local height = math.max(math.ceil(core_percent * bar_height / 100.0), 1)

      cr:rectangle(i * bar_width, area_height - height - 1, bar_width, height)
      cr:fill()
    end
end

```

### DIFF
--- a/docs/modules/Cairo.md
+++ b/docs/modules/Cairo.md
@@ -134,13 +134,6 @@ A longer example, used to create the clock in the image at the top of the page, 
 <summary>Circle clock</summary>
 
 ```lua
-function get_ms()
-    return ironbar:unixtime().subsec_millis / 1000
-    -- Only using the stdlib would require something like:
-    -- local ms = tostring(io.popen('date +%s%3N'):read('a')):sub(-4, 9999)
-    -- return tonumber(ms) / 1000
-end
-
 function draw(cr, width, height)
     local center_x = width / 2
     local center_y = height / 2
@@ -151,7 +144,7 @@ function draw(cr, width, height)
     local hours = date_table["hour"]
     local minutes = date_table["min"]
     local seconds = date_table["sec"]
-    local ms = get_ms()
+    local ms = ironbar:unixtime().subsec_millis / 1000
 
 
     local label_seconds = seconds


src/clients/lua.rs
```diff
@@ -14,6 +18,13 @@ impl LuaEngine {
     pub fn new(config_dir: &Path) -> Self {
         let lua = unsafe { Lua::unsafe_new() };
 
+        if let Err(err) = lua
+            .globals()
+            .set("ironbar", IronbarUserData::new(config_dir))
+        {
+            warn!("{err:?}");
```
JakeStanger: This should probably be an error rather than warning - as far as I'm aware, there's no 'valid' scenario where this would fail, only a Lua runtime issue.

https://docs.rs/mlua/latest/src/mlua/table.rs.html#63-84

untoldwind:
Sure, I reverted to an earlier version with `var_get`  and `var_list` that is more aligned with the cli commands.

The documentation should cover the changes. I was hesitant to add my own example though, this probably belongs to its own examples section.

docs/modules/Cairo.md
```diff
@@ -91,11 +91,34 @@ let {
...
 Outside of this, you can do whatever you like. 
 The full lua `stdlib` is available, and you can load in additional system packages as desired.
 
+Additionally there is basic access to the ironbar via the `ironbar` global:
+- `ironbar.config_dir`: Absolute path to the configuration directory. This can be used for relative file imports, e.g.:
+  ```lua
+   local_module = dofile(ironbar.config_dir .. "local_mod.lua")`
+   ```
+- `ironbar:log_debug(msg)`, `ironbar:log_info(msg)`, `ironbar:log_warn(msg)`,`ironbar:log_error(msg)`: Write a log message.
+- `ironbar:unixtime()`: Returns high-resultion unixtime (stdlib only offers second-resultion). Will return a table:
```
JakeStanger: typo:
```suggestion
- `ironbar:unixtime()`: Returns high-resolution unixtime (stdlib only offers second-resolution). Will return a table:
```

docs/modules/Cairo.md
```diff
@@ -112,14 +135,16 @@ A longer example, used to create the clock in the image at the top of the page,
 
 ```lua
 function get_ms()
-    local ms = tostring(io.popen('date +%s%3N'):read('a')):sub(-4, 9999)
-    return tonumber(ms) / 1000
+    return ironbar:unixtime().subsec_millis / 1000
+    -- Only using the stdlib would require something like:
+    -- local ms = tostring(io.popen('date +%s%3N'):read('a')):sub(-4, 9999)
+    -- return tonumber(ms) / 1000
```
JakeStanger: I reckon we can drop this - the focus of the example here is on using the API rather than using Lua.

untoldwind:
Ah sorry, got it backward.
Its a bigger change though: I activated the code in the sysinfo/upower/config as well (otherwise the variable manager would be pretty much empty).
I was not so sure about the `dynamic_value` though, this is probably not useful in the cairo/lua context.

The other way round though might be interesting (i.e. using a lua-script as source for a dynamic  value). I can take a look into that if you think that would be worthwhile.

JakeStanger:
> The other way round though might be interesting (i.e. using a lua-script as source for a dynamic value). I can take a look into that if you think that would be worthwhile.

Having a `var_set` function that sets ironvars would definitely be useful. I don't think I'd want a third system for dynamic values (eg scripts, ironvars, lua scripts), but being able to load scripts on startup that can interact with the APIs would be good.

This already *sort* of exists, because the Lua client code will run `<config_dir>/init.lua` on first run. That requires there to be a Cairo module present though currently, and has somewhat limited use without any events system. Perhaps being able to listen to ironvar changes from Lua would suffice (esp as more clients get ironvar namespace support).

PR #1348, opened 2026-01-27
feat: Load cairo/lua like a lua-module

untoldwind:
This is more a suggestion/possible solution for the somewhat hacky `script.replace("function draw" ...`.
Instead of having a draw function in the global scope the script is expected to return the drawing function itself.

This allows a greater flexibility in writing scripts without "polluting" the global scope:
* Just an anonymous function: 
  ```lua
  return function(cr, width, height)
     -- draw the thing
  end
  ```
* Some local function:
  ```lua
  local function draw_the_thing(cr, width, height)
     -- do it
  end
  
  return draw_the_thing
  ```
* A callable table:
  ```lua
  local thingy = {}

  function thingy:draw_it(width, height)
     -- do it on self.cr
  end
  
  setmetatable(thingy, {
   __call = function(o, cr, area_width, area_height)
      o.cr = cr
      return o:draw_it(area_width, area_height)
   end,
  })
  return thingy
  ```

As a bonus it is also no longer necessary to patch the error message to get the correct file reference.


JakeStanger:
Oh yeah that's way better than the current solution, I like that. Again I think this paves the way towards a more robust Lua API implementation too.

untoldwind:
Cleaned it up a bit and adjusted the documentation.
Also added an example based on the cpu graph I am using. There are still some quirks though:
* It only works if the `sys_info` module is configured/added as well (otherwise there are no variables). Not entirely sure how to work around this.
* The `height` in the widget seems to be ignored. Had no time to look into that in detail (might be a problem with my css)

src/modules/mod.rs
```diff
@@ -105,6 +105,7 @@ pub struct WidgetContext<TSend, TReceive>
 where
     TSend: Clone,
 {
+    #[allow(dead_code)]
```
JakeStanger: Is there a reason for this `allow`?

docs/examples/custom/Cpu-Cores-Graph.md
```diff
@@ -0,0 +1,164 @@
...
...dnZM3pGkFlpvSFazavevpGR3pj9zwDojZmSO7G0BUyTf/VL/Ebqi55IFZcjPddwQtOj1ZrvP0G1dijm0g+hZOmyhBWNweLyl12tIyLDKe+krCPkNf7tLWue+c5bYL18LpBd/tFtNyHpqC8divR0wwfPIy4Onn59iA8co8innpU+D9bqlzWPinZs2wtgeW0/FZPvBUBLYSqwS1L2SI19tcR8SSpo0IVvIpVCCCHOqJ/YLoQQonZcnQuExwOZxrSRZNPblg7SLX7WWhE80FoBTF/QUGLU8egcG3gkpesPp3PKSxoUI9curXkB6bq5MqM0HBZ2CiGEWKJKrkYhhBC7g6mYiY8u/lTi93n5p80jSMvHbU3vPrlLYZZ7W4L65Ug/XHef8TNN8m8SphfToWTdutlXaao7nGseQlnaJjYWAzthjWUC5fxvUZiXA6vAcPx9dwg5feHChoNxgz5G92Zh0CAlin3jBD/mchVSn65rJg2mAmz/yobfyJQP+zYBulo26KTxqpkQLBsAulMBkItW/6YV1w/A+Ynqs7bl/wnAGSnZHUIWndz53PqeojUytZLVpEzrtYdEX9IBvUnfJWg8W+cAG8EBGJfkEDDy2sNzcXv6Uh+Da7c7u954gLOquzYe5goAj2+JHR1DsXjPMz/JrXTdLxBXdsCpkPshf5NTOW+6D/sacWUHnErkvpDiIrn/3ODztqCcD1QWSEFGtuXCeopbMmj33PJMshoAuHhjlU4A0J8EYK1Sz8hxL8S6T1pNsG/KoS9YWeOGx1+5XqWQLvh6S5dy+2tvh0Xf2H0SJaraV5chTnnM72UJwxEl6Y7vDQZskmJkdS482LQIqFZvZrcDfQegbA21m0f1pwiA4f8cb64DaF8MgHxVQSOf6z9XzJcK24CT5XY/CewpnGEKymAfUCRKdGggs77eg8OwnIKuRCtg6+LY7FmywveMtnkLXIyHmq1KIvnDkUs5vgqAQQ1AiQ0g7yMAkoYAdq8F6FwAwF/TAM5mBozAPQugdSnAhXmA8kXKlNz29ciUPGIsIx2ZgfTDsQCczwhQvqUtU6FGhhe+C7nl2/KMeLqt9jFFCbKr8qo37QX/Bt8TC2HfEB8SoPxBdN4JV24BPiNxOBJnXnKpkfHKjbu9wcHNHORXbkIbAURfyFVC9BlNJfwL82huvaouLsIAAAAASUVORK5CYII=)
```
JakeStanger: I don't think GH supports base64 images unfortunately. If you paste the upload the screenshot here, I can stick it on my image host and give you a link.

src/modules/cairo.rs
```diff
@@ -139,19 +182,12 @@ impl Module<gtk::Box> for CairoModule {
 
         let lua = context.ironbar.clients.borrow_mut().lua(&config_dir);
 
-        // this feels kinda dirty,
-        // but it keeps draw functions separate in the global scope
-        let script = fs::read_to_string(&self.path)?
-            .replace("function draw", format!("function __draw_{id}").as_str());
-        lua.load(&script).exec()?;
+        // Keep draw function in a mutex so it can be replaced on file change
+        let draw_function = Rc::new(Mutex::new(self.load_draw_function(&lua)));
```
JakeStanger:
```suggestion
        let draw_function = arc_mut!(self.load_draw_function(&lua));
```

You may also get away with a cheaper `Rc<RefCell<T>>` here as everything should be on the local thread, in which case you can use `rc_mut!`.

src/modules/cairo.rs
```diff
@@ -164,15 +200,16 @@ impl Module<gtk::Box> for CairoModule {
 
                 let ptr = cr.to_glib_full();
 
-                // mlua needs a valid return type, even if we don't return anything
-                if let Err(err) =
-                    function.call::<Option<bool>>((id.as_str(), LightUserData(ptr.cast()), w, h))
+                if let Some(ref current_draw_function) = *draw_function.lock().expect("Mutex lock")
```
JakeStanger:
```suggestion
                if let Some(ref current_draw_function) = *lock!(draw_function)
```

JakeStanger:
> It only works if the sys_info module is configured/added as well (otherwise there are no variables). Not entirely sure how to work around this.

Yeah this is a hole in the implementation at the moment. In lieu of a full clients API, it might be sufficient to introduce a method on the Lua side that just inits the client. Something like `ironbar:ensure_client("sysinfo")` which then internally calls `ironbar.clients.sysinfo()` and throws away the result. 

Eventually that can be amended to return an actual object allowing you to interact with the client directly.

That's probably best suited for another PR though.

src/modules/mod.rs
untoldwind: clippy complained that no other widget was using it any more.

JakeStanger: In that case I'm happy for it to be removed

untoldwind:
Actually spend some time to make it extra small:

t

src/modules/mod.rs
untoldwind: Ok, removed it

src/modules/cairo.rs
```diff
@@ -139,19 +181,12 @@ impl Module<gtk::Box> for CairoModule {
 
         let lua = context.ironbar.clients.borrow_mut().lua(&config_dir);
 
-        // this feels kinda dirty,
-        // but it keeps draw functions separate in the global scope
-        let script = fs::read_to_string(&self.path)?
-            .replace("function draw", format!("function __draw_{id}").as_str());
-        lua.load(&script).exec()?;
+        // Keep draw function in a mutex so it can be replaced on file change
+        let draw_function = arc_mut!(self.load_draw_function(&lua));
```
JakeStanger: Did an `rc_mut` not work here? I'm surprised because there shouldn't be any threading involved

untoldwind:
You are right, I did not notice the `spawn_future_local` and was a bit too cautious.
Guess I should have more faith in the  rust compiler and the Send/Sync semantic.

docs/examples/custom/Cpu-Cores-Graph.md
JakeStanger: <https://f.jstanger.dev/github/ironbar/cario-cpu-graph.png>

Issue #415, opened 2024-01-24
Is dynamic classes & states currently possible? Also needs documentation


izelnakri:
Problem statement: As a user I'd like style my battery module when it's charging(or not) and the battery levels. Same dynamic styling is needed for disk usage, cpu, memory etc.

In waybar I could create a module with different states for different css class name styling. 

For example in config.json:
```
  "disk": {
    "format": "🗄️ {percentage_used}% [{used}]",
    "states": {
      "very-high": 80,
      "high": 60
    },
    "on-click": "tmux kill-session -t disk || alacritty -e tmux new-session -s disk \"zsh -c 'df -h'; zsh\""
  },
```
This allows me in waybar to style different states via:

```
#disk.very-high {
  color: @waybar_red_color;
}
#disk.high {
  color: @waybar_orange_color;
}
```

**How can I go about implementing this in ironbar?**

One not-so-ideal solution I found is to create different modules with `type: custom` and has different `show_if`s and `class` attributes. However this solution is very verbose and requires me to create bash scripts(I think?). I'm not even sure how to accomplish this with Pango markup, could you provide an example for this scenario in the documentation?

Basically I think we need:

- a custom `states` option object & assignments for `type: custom` and have it documented, or make the `class` value `Dynamic String`, even then I'm not sure how to add these classes dynamically on same bash(?) script check?. I'm also not sure how I can create conditionals in Pango markup. I found Pango markup documentation to be insufficient to say the least :sweat_smile: 
- If we'd like to keep things as is, with no changes. That means the not-so-ideal `show_if` solution is the way to go. It would be great if we can put a basic example in the documentation for a battery level + charging states custom css classes assignment/change.
- A documented way to declare internal `Module`s for ironbar.

JakeStanger:
Thanks for the detailed issue. It sounds like there's a few things nested into one here.

**As it stands:**

This is not currently achievable natively, and does require some external scripting, but there are ways of doing it with a single `script`  or `label` module.

Pango markup is akin to HTML, and lets you set some basic attributes on a span. This means it does give you some styling control. For example, `<span color="red">hello</span>` will work. The entire spec can be found [here](https://docs.gtk.org/Pango/pango_markup.html).

You could output the entire markup from an external script directly (doesn't have to be bash, just needs to write to stdout), or use a l`abel` along with the `ironvar` feature.

For example with ironvars:

```json
{
  "end": [
    {
      "type": "label",
[...]

PR #1349, opened 2026-01-27
Standardised profiles system

JakeStanger:
Adds a new standard configuration system for setting up module "profiles", which allow a defined subset of the configuration to change reactively when a certain value (volume, battery, brightness) reaches defined thresholds.

The primary use of this system is to allow for dynamic icons (ie the volume indicator, wifi strength indicator) to be fully configurable. This approach also allows for any options included in a profile definition to be part of the same system. 

It also introduces a more declarative reactive UI approach via the profile update callback, which should in time lead to cleaner module code.

Included in this PR is support for the following modules:

- `volume` (configure volume icon)
- `battery` (configure format)

The system would additionally apply to the following modules:

- `backlight` (configure brightness icon, #1235)
- `network_manager` (configure wifi strength icon, #1233)

The following locations would also expand to use this with future updates:

- `battery` (configure battery icon, #935)
- `music` (configure volume icon)

In future, `sys_info` may also be supported, although this is not being considered for now.

## Configuration

Profiles always include a "default" which is merged flat into the config. This means setting eg `icons.volume` *outside* of the profiles object will apply if no other profile definition takes precedent. 

Each profile **must** include a `when` key, which indicates the threshold. Once the monitored value (eg volume %) is **less than or equal to** the threshold, this profile activates. 

Only one profile is active at a time. In the example below, if the volume was `<= 33%`, the low profile would activate.

```corn
let {
  $volume = {
    // default - applies if no other profile is active
    icons.volume = "󰕾"
    
    profiles = {
        // medium profile activates when volume <= 67%
        medium.when = 67
        medium.icons.volume = "󰖀"

        // low profile activates when volume <= 33%
        low.when = 33
        low.icons.volume = "󰕿"

        // you can also override other icons
        low.icons.muted = "icons:volume-low-muted" 
    }
  }
} in {
  end = [ $volume ]
}
```

<details>
<summary>Same config in JSON for clarity & highlighting</summary>

```json
{
  "end": [
    {
      "icons": {
        "volume": "󰕾"
      },
      "profiles": {
        "medium": {
          "value": 67,
          "icons": {
            "volume": "󰖀"
          }
        },
        "low": {
          "value": 33,
          "icons": {
            "volume": "󰕿",
            "muted": "icons:volume-low-muted"
          }
        }
      }
    }
  ]
}

```

</details>

The system also supports compound matchers, which is suitable where a single state value is not considered sufficient. **These should be kept simple and contain a minimal amount of values, two or three at an absolute push.**

An example below for the battery module. A `percent` value must be provided. A `charging` value can also be provided, which takes higher priority.

```corn
let {
  $battery = {
    type = "battery"
  
    format = "HIGH {percentage}%"
  
    profiles = {
        low.when = { percent = 20 }
        low.format = "LOW {percentage}%"
  
        low-charging.when = { percent = 20 charging = true }
        low-charging.format = "LOW (CHARGING) {percentage}%"
  
        medium.when = { percent = 50 charging = false }
        medium.format = "MEDIUM {percentage}%"
  
        medium-charging.when = { percent = 50 }
        medium-charging.format = "MEDIUM (CHARGING) {percentage}%"
  
        good.when = { percent = 75 charging = false }
        good.format = "GOOD {percentage}%"
  
        good-charging.when = { percent = 75 charging = true }
        good-charging.format = "GOOD (CHARGING) {percentage}%"
  
        empty.when = { percent = 1 charging = true }
    } 
} in {
  end [ $battery ]
}
```

<details>
<summary>JSON</summary>

```json
{
  "end": [
    {
      "type": "battery",
      "format": "HIGH {percentage}%",
      "profiles": {
        "low": {
          "when": {
            "percent": 20
          },
          "format": "LOW {percentage}%"
        },
        "low-charging": {
          "when": {
            "percent": 20,
            "charging": true
          },
          "format": "LOW (CHARGING) {percentage}%"
        },
        "medium": {
          "when": {
            "percent": 50,
            "charging": false
          },
          "format": "MEDIUM {percentage}%"
        },
        "medium-charging": {
          "when": {
            "percent": 50
          },
          "format": "MEDIUM (CHARGING) {percentage}%"
        },
        "good": {
          "when": {
            "percent": 75,
            "charging": false
          },
          "format": "GOOD {percentage}%"
        },
        "good-charging": {
          "when": {
            "percent": 75,
            "charging": true
          },
          "format": "GOOD (CHARGING) {percentage}%"
        },
        "empty": {
          "when": {
            "percent": 1,
            "charging": true
          }
        }
      }
    }
  ]
}
```

</details>

Each profile is named according to its key (low, medium). When a profile is active, a class name is appended to the relevant widget in the form `profile-{name}`. This allows for dynamic styling (#415), and for replacing the battery `thresholds` option.

## Config API

The API is more involved than the original thresholds design. 

Within a configuration struct you'd place a `profiles` field of type `Profiles<S, T>`. `S` represents the state/matcher type, and `T` represents the configuration contained within a profile. 

You **must** include `#[serde(flatten)]` on the field to avoid a double `profiles.profiles` syntax and ensure the default is merged at the top-level correctly.

```rust
struct VolumeModule {
  #[serde(flatten)]
  pub(super) profiles: Profiles<f64, VolumeProfile>,
}
```

```rust
struct BatteryModule {
    #[serde(flatten)]
    profiles: Profiles<ProfileState, BatteryProfile>,
}
```

### Compound state

Complex state objects must manually implement the `PartialOrd` and `State` traits. Deriving `PartialOrd` will likely produce an incorrect sorter.

The PartialOrd implementation must obey the following rules:

- The primary field must be sorted first (eg battery percentage)
- For optional fields, a `Some` variant is *less* than a `None`. This is because internally the first matching profile is used.

```rs
impl PartialOrd for ProfileState {
    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
        if self.percent == other.percent {
            match (self.charging, other.charging) {
                (Some(_), Some(_)) | (None, None) => Some(Ordering::Equal),
                (None, Some(_)) => Some(Ordering::Greater),
                (Some(_), None) => Some(Ordering::Less),
            }
        } else {
            self.percent.partial_cmp(&other.percent)
        }
    }
}
```

The state implementation requires a single method, which checks whether an input state matches the profile matcher. Again optional fields need to be considered here - if omitted in `&self` (the config value), matching should ignore this field.

```rs
impl State for ProfileState {
    fn matches(&self, value: &Self) -> bool {
        match self.charging {
            Some(charging) => {
                charging == value.charging.expect("value should exist")
                    && value.percent <= self.percent
            }
            None => value.percent <= self.percent,
        }
    }
}
```

### Profiles

Each module should provide a default implementation for its profile type, and a set of appropriate default profiles.

```rs
#[derive(Debug, Default, Clone, Deserialize)]
#[serde(default)]
pub struct VolumeProfile {
    pub(super) icons: Icons,
}

impl Default for Icons {
    fn default() -> Self {
        Self {
            volume: "󰕾".to_string(),
            muted: "󰝟".to_string(),
        }
    }
}

```

#### Defaults

Due to serde limitations, relying on `#[serde(default = "default_profiles")]` does not work, and these must be injected manually. A new module lifecycle method has been added named `on_create` which takes a `&mut self`, allowing for alterations to the config as appropriate. 

The `Profiles` struct exposes a `setup_defaults` method which takes another `Profiles` instance and merges it in. Any keys which do not exist in the user's configuration are added from the defaults.

The `profiles!` macro allows for quick creation of the profile map.

```rs
pub(super) fn default_profiles() -> Profiles<f64, VolumeProfile> {
    profiles!(
        "low":33 => VolumeProfile::for_volume_icon("󰕿"),
        "medium":67 => VolumeProfile::for_volume_icon("󰖀")
    )
}

// -- snip --
impl Module<Button> for VolumeModule {
  fn on_create(&mut self) {
      self.profiles.setup_defaults(config::default_profiles());
  } 
}
```

## Module API

The `Profiles` struct exposes an `attach` method, which allows you to connect a 'primary' widget to the configuration (upon which the classnames are added), and an update callback.

This callback includes arguments for the widget and event data. Event data includes the value sent in the update, the associated profile configuration (including the default), and any user-data sent in the update.

The `attach` method returns a profiles manager used for dispatching updates.

```rs
let mut manager = self.profiles.attach(
    &btn_mute,
    move |btn_mute, event: ProfileUpdateEvent<VolumeProfile, BtnMuteUiUpdate>| {
        btn_mute.set_active(event.data.muted);

        let icons = &event.profile.icons;
        btn_mute.set_label(if event.data.muted {
            &icons.muted
        } else {
            &icons.volume
        });
    },
);
```

Updates can then be sent elsewhere in the module code using `manager.update`.

```rs
manager.update(
    info.volume.percent() as i32,
    BtnMuteUiUpdate { muted: info.muted },
);
```

## Outstanding

- [x] Code review
- [x] Battery implementation
- [x] Consider shorthand syntax to match current battery `thresholds` option
- [x] Configuration documentation
- [x] Development documentation
- [x] Code documentation

---

Supersedes https://github.com/JakeStanger/ironbar/pull/1274

### DIFF
--- a/src/config/profiles.rs
+++ b/src/config/profiles.rs
@@ -2,6 +2,7 @@ use glib::prelude::*;
 use gtk::Widget;
 use gtk::prelude::WidgetExt;
 use serde::Deserialize;
+use std::cmp::Ordering;
 use std::collections::HashMap;
 use std::marker::PhantomData;
 
@@ -89,12 +90,13 @@ impl State for f64 {}
 #[derive(Debug, Default, Clone, Deserialize)]
 #[cfg_attr(feature = "extras", derive(schemars::JsonSchema))]
 #[serde(default)]
-pub struct Profiles<T>
+pub struct Profiles<S, T>
 where
+    S: State,
     T: Default + Clone,
 {
     /// A map of named profiles.
-    profiles: HashMap<String, ProfileEntry<T>>,
+    profiles: HashMap<String, ProfileEntry<S, T>>,
 
     /// The default profile.
     #[serde(flatten)]
@@ -162,22 +164,24 @@ where
 /// and the configuration associated with it.
 #[derive(Debug, Default, Clone, Deserialize)]
 #[cfg_attr(feature = "extras", derive(schemars::JsonSchema))]
-pub struct Profile<T>
+pub struct Profile<S, T>
 where
+    S: State,
     T: Default + Clone,
 {
     /// The 'state' value threshold at which this profile should be activated.
     /// Values *less than or equal to* this will activate it
     /// (assuming the current value does not fall into a profile below this one).
-    value: i32,
+    when: S,
 
     /// The configuration data attached to this profile.
     #[serde(flatten)]
     data: T,
 }
 
-impl<T> Profile<T>
+impl<S, T> Profile<S, T>
 where
+    S: State,
     T: Default + Clone,
 {
     /// Creates a new profile for this matcher `when`,
@@ -187,8 +191,9 @@ where
     }
 }
 
-impl<T> Profiles<T>
+impl<S, T> Profiles<S, T>
 where
+    S: State,
     T: Default + Clone,
 {
     /// Merges a default profile set into this one,
@@ -208,7 +213,7 @@ where
     ///   }
     /// }
     /// ```
-    pub fn setup_defaults(&mut self, defaults: Profiles<T>) {
+    pub fn setup_defaults(&mut self, defaults: Profiles<S, T>) {
         for (name, profile) in defaults.profiles {
             self.profiles.entry(name).or_insert(profile);
         }
@@ -227,20 +232,21 @@ where
     ///     );
     /// });
     /// ```
-    pub fn attach<W, F, D>(&self, widget: &W, on_update: F) -> ProfilesManager<T, W, F, D>
+    pub fn attach<W, F, D>(&self, widget: &W, on_update: F) -> ProfilesManager<S, T, W, F, D>
     where
         W: IsA<Widget>,
-        F: Fn(&W, ProfileUpdateEvent<T, D>),
+        F: Fn(&W, ProfileUpdateEvent<S, T, D>),
     {
         ProfilesManager::new(self.clone(), widget.to_owned(), on_update)
     }
 }
 
-impl<T> From<HashMap<String, Profile<T>>> for Profiles<T>
+impl<S, T> From<HashMap<String, Profile<S, T>>> for Profiles<S, T>
 where
+    S: State,
     T: Default + Clone,
 {
-    fn from(profiles: HashMap<String, Profile<T>>) -> Self {
+    fn from(profiles: HashMap<String, Profile<S, T>>) -> Self {
         Self {
             profiles: profiles
                 .into_iter()
@@ -254,13 +260,14 @@ where
 /// A manager for a set of profiles,
 /// used to determine which profile to load and
 /// send updates to the attached callback.
-pub struct ProfilesManager<T, W, F, D>
+pub struct ProfilesManager<S, T, W, F, D>
 where
+    S: State,
     T: Default + Clone,
     W: IsA<Widget>,
-    F: Fn(&W, ProfileUpdateEvent<T, D>),
+    F: Fn(&W, ProfileUpdateEvent<S, T, D>),
 {
-    profiles: Profiles<T>,
+    profiles: Profiles<S, T>,
     widget: W,
     on_update: F,
 
@@ -271,24 +278,26 @@ where
 }
 
 /// An event sent by a `ProfilesManager` when `update` is called.
-pub struct ProfileUpdateEvent<'a, T, D> {
+pub struct ProfileUpdateEvent<'a, S, T, D> {
     /// The state provided in the update.
     /// Note this is not the state matcher of the profile itself.
-    pub value: i32,
+    pub state: S,
     /// The profile configuration data.
     pub profile: &'a T,
     /// Any additional non-profile data passed in the update.
     pub data: D,
 }
 
-impl<T, W, F, D> ProfilesManager<T, W, F, D>
+impl<S, T, W, F, D> ProfilesManager<S, T, W, F, D>
 where
+    S: State,
     T: Default + Clone,
     W: IsA<Widget>,
-    F: Fn(&W, ProfileUpdateEvent<T, D>),
+    F: Fn(&W, ProfileUpdateEvent<S, T, D>),
 {
-    fn new(profiles: Profiles<T>, widget: W, on_update: F) -> Self {
+    fn new(profiles: Profiles<S, T>, widget: W, on_update: F) -> Self {
         let p_map = &profiles.profiles;
+
         let mut profile_keys = p_map.keys().map(ToOwned::to_owned).collect::<Vec<_>>();
 
         profile_keys.sort_by(|a, b| {
@@ -320,7 +329,7 @@ where
     ///
     /// Note that the `on_update` callback runs every time this is called,
     /// regardless of whether the profile has changed.
-    pub fn update(&mut self, value: i32, data: D) {
+    pub fn update(&mut self, value: S, data: D) {
         let new_profile_name = self.profile_keys.iter().find(|&name| {
             let profile = &self.profiles.profiles[name];
             profile.when().matches(&value)
@@ -340,7 +349,7 @@ where
         }
 
         let update_data = ProfileUpdateEvent {
-            value,
+            state: value,
             profile,
             data,
         };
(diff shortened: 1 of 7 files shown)


xMAC94x:
Feedback:
 - this definitively needs documentation hints in code or wiki, because just from code its hard to get why and what this does on the first glance (the PR description is quite clear though, ty)
 - afaik this relies on profile names and overwriting. that means: If the default for battery is
 ```
 low: 33
 medium: 67
 ```
 
 and I want to make it steps of 20%
 ```
 critical: 20%
 charge: 40%
 medium: 60%
 strong: 80%
 ```
 
 it will get merged to something like:
 ```
 critical: 20%
 low: 33%  <-------- this comes from the default
 charge: 40%
 medium: 60%
 strong: 80%
 ```
 
 so its always important to reuse the names from the default config. Is it possible to completely null default values, in order to give users the possibility to have their own naming. e.g. by:
 ```
 critical: 20%
 low: null <---- overwring the default explicitly here!
 charge: 40%
 medium: 60%
 strong: 80%
 ```
 - one cool feature of `waybar` is that you can have different icons per `battery-percent` but you can also use a different icon when the battery is plugged in. So you can have something like: Yes the battery is at 20% but its charging - which is ofc way less critical then a battery at 20% which is not charging. If we stay with i32, we somehow have to merge this data into a single information.  maybe its worth allowing to define a own comparison function instead of hardcode to i32.
 -  other than that concerns i really like the flexibility that gives

src/config/profiles.rs
```diff
@@ -0,0 +1,196 @@
...
+#[derive(Debug, Default, Clone, Deserialize)]
+#[cfg_attr(feature = "extras", derive(schemars::JsonSchema))]
+pub struct Profile<T>
+where
+    T: Default + Clone,
+{
+    /// The value threshold at which this profile should be activated.
+    /// Values *less than or equal to* this will activate it
+    /// (assuming the current value does not fall into a profile below this one).
+    value: i32,
```
xMAC94x: `values` is really a very generic word. maybe we should change this to something like `threshold` or `active_below`

src/config/profiles.rs
```diff
@@ -0,0 +1,196 @@
...
+        (self.on_update)(&self.widget, update_data);
+    }
+
+    fn update_classes(&self, new_profile: Option<&String>) {
+        for profile in &self.profile_keys {
+            let class = format!("profile-{profile}");
+            if Some(profile) == new_profile {
+                self.widget.add_css_class(&class);
+            } else {
+                self.widget.remove_css_class(&class);
```
xMAC94x, line 366: any reason why we loop through all and remove all and just timply remove `curr_profile` ?

is it possible to do something like this
```
self.widget.remove_css_class_starting_with("profile-");
self.widget.add_css_class(format!("profile-{profile}"));
```

JakeStanger: GTK doesn't have a `remove_css_class_starting_with`. We could implement an extension method but then we'd have to loop through every class, which'd be less efficient.

JakeStanger:
> Is it possible to completely null default values, in order to give users the possibility to have their own naming. e.g. by:

Yes, but you'd pass an empty object rather than `null`. So eg `profiles.low = {}`. In this case it would still apply the profile though (ie you'd get the classname), which may be undesirable.

I guess in this case, you could just avoid adding the lower "critical" level, and just change the threshold of "low".

> Maybe its worth allowing to define a own comparison function instead of hardcode to i32.

This is a good point. I guess the system needs to allow for a generic hashable (orderable?) struct so that you could do something like:

```corn
{
  profiles.low.value = { percent = 20 charging = false }
  profiles.low.icon = "icons:charging-critical"
}
```

EDIT: I'm thinking about this, and it might not actually. If the battery profile exposes `icon` and `icon_charging`, and a `.charging` class is toggled depending on state, that'd likely give enough control. Then again, you might want to also control the `format` depending on charging state, etc. It'd end up doubling up a lot of options without this.

src/config/profiles.rs
JakeStanger: Agreed. If we support complex values as discussed for the battery case I think we'll need a new name altogether. Will have a think.

JakeStanger: I've opted for `when` for now, but considering `if`. I want to keep it short and simple and clear that it's the matcher.

JakeStanger:
I've added support for generic compound value matchers, and an implementation for the battery module (allowing you to match on charging state). I've updated the main post with details.

The system allows for matchers to be of any complexity, but implementing it would be a nightmare and I cannot imagine it'd be great for users.  The aim therefore should be to keep things no more complex than the battery module if we can help it. I can't imagine a case where we couldn't.

xMAC94x:
looks good to me :)

JakeStanger:
Okay good to go I reckon! Cheers for your feedback.

PR #1350, opened 2026-01-29
Added config option to change app icon sizes in the menu module.

benpz:
feat: allow changing app icon sizes in the menu module.

### DIFF
--- a/src/modules/menu/config.rs
+++ b/src/modules/menu/config.rs
@@ -105,7 +105,7 @@ pub struct MenuModule {
     /// Size of the `label_icon` image.
     pub(super) label_icon_size: i32,
 
-    /// Size of the `app_icon_size` images.
+    /// Size of the icon to display in menu entries.
     pub(super) app_icon_size: i32,
 
     // -- common --


docs/modules/Menu.md
```diff
@@ -20,6 +20,7 @@ It is also possible to add custom categories and actions into the menu.
 | `label`               | `string`                                             | `≡`                                  | Label to show on the menu button on the bar.                                                                                                                            |
 | `label_icon`          | `string`                                             | `null`                               | Icon to show on the menu button on the bar.                                                                                                                             |
 | `label_icon_size`     | `integer`                                            | `16`                                 | Size of the label_icon image.                                                                                                                                           |
+| `app_icon_size`       | `integer`                                            | `16`                                 | Size of the app icon images.                                                                                                                                            |
```
JakeStanger: The current description is a bit vague, which to be fair so are the existing ones. Sometihng like this would be better pls:

```suggestion
| `app_icon_size`       | `integer`                                            | `16`                                 | Size of the icon to display in menu entries.                                                                                                                                            |
```

(same for doc comment)

JakeStanger (changes requested):
lgtm, pending the documentation

Issue #576, opened 2024-05-08
Volume PopUp customization


madfanat:
I would like to deeply configure and style the volume/device control popup: change slider orientation, color palette and so on.

I tried to find these (style) settings myself, but couldn’t. If it’s my fault, please give an example of a more detailed configuration.

JakeStanger:
The orientation of the sliders is currently hard-coded, although this can be surfaced via the config. Which sliders are you looking to change? 

Would this require the popup layout to change to accommodate, or are you looking to just rotate the sliders within the current layout? This would potentially result in a very wide and short popup window.

All colours and styles should be customisable via CSS. I don't have any examples to hand, but if you're looking for specifics I can assist.

madfanat:
It’s a bit inconvenient to me that the volume slider is very thin but spends a lot of useless space (I think the reason is the device switching menu).
Maybe it would be better to move the slider to the edge and make the switching menu much shorter. Another way I see is just to rotate the slider, but this will result in a very long popup. To solve this I think we could move the app-specific tab to the top if it’s not hardcoded.

madfanat:
![IMG_0272](https://github.com/JakeStanger/ironbar/assets/168630462/ade3b405-872a-46e7-b6dc-313187046a2c)

Also I’m about to change the background color of the text (circled) to make the look uniform.

PR #1353, opened 2026-01-30
feat(volume): support horizontal slider

imgurbot12:
Adds in a very basic extra configuration option to support a horizontal bar instead of a vertical one for the volume module.
Partially solves https://github.com/JakeStanger/ironbar/issues/576 and was approved for implementation in https://github.com/JakeStanger/ironbar/discussions/1352.

### DIFF
--- a/docs/modules/Volume.md
+++ b/docs/modules/Volume.md
@@ -65,7 +65,7 @@ This requires PulseAudio to function (`pipewire-pulse` is supported).
 [[end]]
 type = "volume"
 format = "{icon} {percentage}%"
-sink_slider_orientation = "vertical",
+sink_slider_orientation = "vertical"
 max_volume = 100
 truncate = "middle"
 


src/modules/volume.rs
```diff
@@ -29,6 +29,11 @@ pub struct VolumeModule {
     /// **Default**: `{icon} {percentage}%`
     format: String,
 
+    /// The orientation of the slider
+    ///
+    /// **Default**: false (vertical)
+    horizontal: bool,
```
JakeStanger: `horizontal` is a vague property name when it only affects the sink volume slider. Additionally this would be better if it took an `Orientation` rather than `bool` for consistency. 

```suggestion
    sink_slider_orientation: Orientation,
```

Could you update accordingly please. Can you also please update the doc comment to add the same detail there (ie which slider it's referring to?

JakeStanger (changes requested):
Thanks. One change request, and if you could update the docs to include the new property too please that'd be great.

src/modules/volume.rs
imgurbot12: Done as requested. `gtk4::Orientation` did not have a Deserialize or JsonSchema implementation, but I found a type called `ModuleOrientation` in your config definition that I'm assuming you wanted to use here.

Docs have been added to `docs/` as well.

PR #1359, opened 2026-02-01
feat(volume): source/output support with additional config/css controls

### LINKED ISSUE #576: Volume PopUp customization

### LINKED ISSUE #1032: source module
I'd like an equivalent to the volume module but for the microphone, for showing the mic status and right clicking to disable it for example.

imgurbot12:
Adds microphone/audio-input support to the volume module.

Resolves #576 
Resolves #1032

Approved for implementation in https://github.com/JakeStanger/ironbar/discussions/1352.

Changelog:
  
- implements pulseaudio `Source`/`SourceOutput` integration with client/module-ui
- adds `source_slider_orientation` to control orientation of new source slider
- adds `popup_orientation` to control orientation of popup construction
- adds `mute_format` to allow variance in label format when a sink/source is muted (to hide percentage when muted for example)
- adds `mic_high/mic_medium/mic_low/mic_muted` icon types for new sources
- modifies existing config option `icons.muted` to `icons.volume_muted` to differentiate from `mic_muted`.
- applies truncate rules to sink/source names in their respective dropdowns

Examples below:

<details>
  <summary>Default Orientation</summary>

image
</details>

<details>
  <summary>Custom Orientation</summary>

scrn-2026-02-01-00-43-17

</details>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -771,9 +771,9 @@ dependencies = [
 
 [[package]]
 name = "deranged"
-version = "0.3.11"
+version = "0.5.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b42b6fa04a440b495c8b04d0e71b707c585f83cb9cb28cf8cd0d976c315e31b4"
+checksum = "ececcb659e7ba858fb4f10388c250a7252eb0a27373f1a72b8748afdd248e587"
 dependencies = [
  "powerfmt",
 ]
@@ -2455,9 +2455,9 @@ dependencies = [
 
 [[package]]
 name = "num-conv"
-version = "0.1.0"
+version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "51d515d32fb182ee37cda2ccdcb92950d6a3c2893aa280e540671c2cd0f3b1d9"
+checksum = "cf97ec579c3c42f953ef76dbf8d55ac91fb219dde70e49aa4a6b7d74e9919050"
 
 [[package]]
 name = "num-derive"
@@ -3659,30 +3659,30 @@ dependencies = [
 
 [[package]]
 name = "time"
-version = "0.3.37"
+version = "0.3.47"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "35e7868883861bd0e56d9ac6efcaaca0d6d5d82a2a7ec8209ff492c07cf37b21"
+checksum = "743bd48c283afc0388f9b8827b976905fb217ad9e647fae3a379a9283c4def2c"
 dependencies = [
  "deranged",
  "itoa",
  "num-conv",
  "powerfmt",
- "serde",
+ "serde_core",
  "time-core",
  "time-macros",
 ]
 
 [[package]]
 name = "time-core"
-version = "0.1.2"
+version = "0.1.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ef927ca75afb808a4d64dd374f00a2adf8d0fcff8e7b184af886c3c87ec4a3f3"
+checksum = "7694e1cfe791f8d31026952abf09c69ca6f6fa4e1a1229e18988f06a04a12dca"
 
 [[package]]
 name = "time-macros"
-version = "0.2.19"
+version = "0.2.27"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2834e6017e3e5e4b9834939793b282bc03b37a3336245fa820e35e233e2a85de"
+checksum = "2e70e4c5a0e0a8a4823ad65dfe1a6930e4f4d756dcd9dd7939022b5e8c501215"
 dependencies = [
  "num-conv",
  "time-core",


imgurbot12:
One thing that's mildly annoying is the escaping of standard characters for the `app-box` generated inputs/outputs. You'll notice in my screenshots that the music-player I'm using, [strawberry](https://www.strawberrymusicplayer.org/), wraps the track-name/author in quotations that turn into ugly `&apos;` literals rather than just normal single-quotes. 

I was looking into fixing but I wasn't sure what your position on that would be since it would likely require removing the escaping, or maybe doing something custom with it.

imgurbot12:
~~Might be a gtk4-rs limitation but I also couldn't seem to get the scale sliders to have actual knobs where the slider is actually larger than the scale trough, like what seems to be possible based on their [documentation](https://docs.gtk.org/gtk4/class.Scale.html):~~

image

~~Tried playing with all of the CSS rules possible and did a lot of troubleshooting, also tried enabling overflow on the scale objects in case that was the issue but nothing seems to work.~~

Edit: Nevermind, I figured it out. I had a global CSS rule with `background-image: none` which apparently breaks the slider knobs. The more ya know. Only took me a few hours to figure that one out lol

JakeStanger:
Thanks for this. I've got to get some other changes shipped first but will review this as soon as I can.

> I was looking into fixing but I wasn't sure what your position on that would be since it would likely require removing the escaping, or maybe doing something custom with it.

I'd first double check that the escaping comes directly from the player and not Ironbar, and also check that it's not caused by ironbar's label escape code (I don't think it should be):
https://github.com/JakeStanger/ironbar/blob/f3f06b565b08f32968353d11469aa491c9811a37/src/gtk_helpers.rs#L203-L209

If it is an player problem then ideally that needs fixing upstream, but if there's an easy enough solution I'm happy to stick that in Ironbar.

imgurbot12:
> Thanks for this. I've got to get some other changes shipped first but will review this as soon as I can.

Yeah, glad to contribute to a cool project like this!
As someone wanting to make use of the bar I would've probably done it anyways for myself regardless, so I'm glad its valuable to others :)

> I'd first double check that the escaping comes directly from the player and not Ironbar, and also check that it's not caused by ironbar's label escape code (I don't think it should be):

Can confirm that ironbar is doing the escaping. The `SinkInput` label is built with [`OverflowLabel`](https://github.com/imgurbot12/ironbar/blob/master/src/gtk_helpers.rs#L263) using the [`set_label_escaped`](https://github.com/imgurbot12/ironbar/blob/master/src/modules/volume.rs#L708) function.

```rust
    /// Set the label text, escaping markup unless the string already contains `<span`.
    /// When marquee is enabled, this also resets scroll state for the new text.
    pub fn set_label_escaped(&self, text: &str) {
        if let Some(marquee) = &self.marquee {
            marquee.set_text(text);
        } else {
            self.label.set_label_escaped(text);
        }
    }
```

When modified to include a print statement and label creation to skip the escape:

```rust
    /// Set the label text, escaping markup unless the string already contains `<span`.
    /// When marquee is enabled, this also resets scroll state for the new text.
    pub fn set_label_escaped(&self, text: &str) {
        if let Some(marquee) = &self.marquee {
            marquee.set_text(text);
        } else {
            println!("sink input label: {text:?}");
            self.label.set_label(text);
        }
    }
```

It shows the original name and proper result

image

image

We probably don't want to remove the escaping altogether, so I'm unsure what the best compromise should be here.

imgurbot12:
The internet seems to suggest that `markup_escape_text` supports double-quotes but not single-quotes and to support single-quotes you need to manually modify the results of `markup_escape_text`.

The following solution worked but I'm not sure if its the best approach..

```rust
impl IronbarLabelExt for Label {
    fn set_label_escaped(&self, label: &str) {
        if label.contains("<span") {
            self.set_label(label);
        } else {
            let text = markup_escape_text(label);
            let text = text.as_str().replace("&apos;", "'");
            self.set_label(&text);
        }
    }
}
```

JakeStanger:
> The following solution worked but I'm not sure if its the best approach..

It's possible that it's a problem with the way `OverflowLabel` is handling escaping. It should only run through the escape function if markup is enabled on the label widget. I can't see any issues from just reading the code though...

If it's not that causing the issue, I think this is fine.

---
I've implemented the profiles system in #1349, which is documented in the following locations:

This includes support for the volume module, but it conflicts with some of the work in your PR. Most of it should be untouched, and it should just be the icon UI code that needs reworking. Are you okay to rebase & update this to use the new system? Let me know if you have any problems with that.

Once that's in place, I should be good to review this properly.

imgurbot12:
@JakeStanger I'm working on merging with the most recent update, but I've hit a snag.

The new profile manager solution seems to be preventing dynamic input/output sources from having a mute button on the UI since the button is not getting an update on init, meaning the button never gets a label and is invisible:

image

Here's what it looks like under `GTK_DEBUG=interactive`. Normally the button would contain a label but it doesn't cause there's no update on initialization, even though you're calling [`manager.update`](https://github.com/JakeStanger/ironbar/blob/master/src/modules/volume/mod.rs#L428).

image

--

Edit: I found a way by passing the button alongside the event object:
```rust
struct BtnMuteUiUpdate {
    muted: bool,
    button: Option<ToggleButton>,
}

let mut sink_manager = self.profiles.attach(
    &sink_mute,
    move |btn_mute, event: ProfileUpdateEvent<f64, VolumeProfile, BtnMuteUiUpdate>| {
        let btn = event.data.button.as_ref().unwrap_or(btn_mute);
        btn.set_active(event.data.muted);

        let icons = &event.profile.icons;
        btn.set_label(if event.data.muted {
            &icons.muted
        } else {
            &icons.volume
        });
    },
);
```
Thankfully you can clone the `ToggleButton` object and still have it work as intended.

imgurbot12:
Something about resetting master before merging the new changes again caused github to auto-close the PR, weird.

Changes have been ported. Here's a changelog since original implementation:

- Icons now use the new profiles system.
- Added `show_monitors` option to avoid including pulseaudio sink monitors in source dropdown list by default
- Added escape improvements to `IronbarLabelExt` to support single quotes instead of converting them into `&apos;`

Happy to noodle with you on the escaping thing still if needed.

docs/modules/Volume.md
```diff
@@ -11,24 +11,30 @@ This requires PulseAudio to function (`pipewire-pulse` is supported).
...
+| `icons.muted`               | `string`                                             | `󰝟`                    | Yes      | Icon to show for muted outputs.                                                                                                                                                                               |
+| `mic_icons.volume`          | `string`                                             | ``                    | Yes      | Icon to show for high microphone volume levels.                                                                                                                                                               |
+| `mic_icons.muted`           | `string`                                             | ``                    | Yes      | Icon to show for muted microphone inputs.                                                                                                                                                                     |
```
JakeStanger: I would prefer these all under the same `icons` object. So eg `icons.mic_volume`, `icons.mic_muted`

src/gtk_helpers.rs
```diff
@@ -204,7 +204,9 @@ impl IronbarLabelExt for Label {
         if label.contains("<span") {
             self.set_label(label);
         } else {
-            self.set_label(&markup_escape_text(label));
+            let text = markup_escape_text(label);
+            let text = text.as_str().replace("&apos;", "'");
```
JakeStanger: This could do with a comment just to detail what it's here for

imgurbot12:
> The commits should reflect the individual changes going into master, as the changelog will be automatically generated from these. Something like the below:

Can do, but since I just manually rebuilt the changes after merging it ended up just being one commit. Do I need to manually re-implement the changes one at a time in order to fit the generated changelog? Should I just make essentially empty commits with each feature as the commit message? Can I include a commit detailed summary with the changelog to get applied properly?

> What is a "source output"? I appreciate this module already suffers from pulse's jargon but I think we need to document what each term means at the very least, because I haven't a clue and I'm sure I'm not the only one.

Sounds good, should I just put a "jargon" section up at the top or how do you want to approach that?
For clarity:

- sink = sound device coming out of your machine (speakers)
- source = sound device going into your machine (microphone)
- sink input = an application/program sending sound to an existing sink (app using speakers)
- source output = an application/program sending sound to a source (app using microphone)

> The code has a lot of repetition in (both the client and module code). It'd be great to see if we can abstract some of the common ground and cut it down a bit.

I noticed that a lot during development as well, but I wasn't sure what your preferred implementation would be for that. The challenge with any implementation is likely that even though the code is very similar between sink/source/input/output it references many different local variables or types, so any simplification seems like it would need to come from a macro.

docs/modules/Volume.md
imgurbot12: Updated to match your requested design in the latest commit.

JakeStanger:
> Do I need to manually re-implement the changes one at a time in order to fit the generated changelog?

Your best bet is to [undo](https://stackoverflow.com/a/15772171) the commits already present, and then stage/commit the files (or chunks) involved in each individual commit. You can then force push this branch when ready.

> should I just put a "jargon" section up at the top

Yeah I think just sticking the bullet points at the top of the volume module docs is fine.

> source output = an application/program sending sound to a source (app using microphone)

I assume this should be "receiving sound from a source"?

> but I wasn't sure what your preferred implementation would be for that

Macros would work, but aren't always the best solution, especially if they end up containing large chunks of code. It varies from case to case, but generally the better approach is either a trait with some default method implementations, or some higher-order type that you can pass the unique data into but let it handle the common logic.

src/gtk_helpers.rs
imgurbot12: Comment added in latest commit, lmk if it needs tweaking.

imgurbot12:
> Your best bet is to [undo](https://stackoverflow.com/a/15772171) the commits already present, and then stage/commit the files (or chunks) involved in each individual commit. You can then force push this branch when ready.

Alright, I'll just force commit after we get a final approval on the changes. Sounds like I get to have some fun with my temporary commits in the meantime :P

> I assume this should be "receiving sound from a source"?

Ahyup. Fixed and updated the docs thanks.

> Macros would work, but aren't always the best solution, especially if they end up containing large chunks of code. It varies from case to case, but generally the better approach is either a trait with some default method implementations, or some higher-order type that you can pass the unique data into but let it handle the common logic.

I'm a bit exhausted at the moment, so I can't think of a great way to do this as of right now but I'd be interested to hear some suggestions on implementation. There aren't any existing types that I can think of that would make to build traits for that could work like this so we'd have to invent another layer of abstraction on the current implementation I think. That could just be my tired brain over-complicating things though lol.

JakeStanger:
Some simple closures/functions would probably work in some cases, I should add too. Whatever gets built will be specific to its usecase, nothing complex and generic. That means yes it'll be new code, but it doesn't need to be fancy or a whole abstraction layer. 

No rush anyway. I'll try and do a more thorough review in the next few days and we can iterate.

imgurbot12:
Fair, sounds like a plan. Ill give it some thought and try to come up with a solution tomorrow.

JakeStanger:
> Sounds like I get to have some fun with my temporary commits in the meantime :P

No worries about this so long as we get the result at the end. My own approach though is to get the commits first right, and then if I need to make a change I make a new local commit, squash/fixup that new commit into the exisiting target commit via a rebase, and then force push.

imgurbot12:
@JakeStanger I scrounged up some brainpower to implement as many abstractions as I can think of at the moment to reduce redundancy, and I'm ready for another pass at review whenever you've got some time.

Would love to get your feedback on the new design, and if you have any recommendations for tweaks or further simplification if necessary. Thanks!

src/modules/volume/mod.rs
```diff
@@ -73,6 +98,88 @@ impl ObjectSubclass for DropdownItemData {
     type Type = DropdownItem;
 }
 
+impl VolumeModule {
```
JakeStanger: Can you add some whitespace throughout this impl please? General preference is a blank line after each fn, and after each block.

JakeStanger (approved):
This looks really good, huge thanks for taking the time to refactor. If you can just have a quick format, and then I'll give this a quick test, I reckon we're good to go.

There may be opportunity in the module code to reduce some of the `add_css_class` boilerplate, but I'm not too worried about that right now. I think any further reductions would come from macros, and I don't think there's a reasonable need for those currently, so happy with this.

src/modules/volume/mod.rs
imgurbot12: added as requested :)

imgurbot12:
Epic. I updated the whitespace as requested, and code should be ready for testing.

Feel free to post any additional feedback and I'll try to get back to you as quickly as I can.
Once you give me the final go ahead I'll see about reswizzling the commits. Thanks!

JakeStanger:
Just had a play with this, All seems to work well! 

I've made a few minor changes, mostly around styling:

- Added some missing css classes, and tweaked some names (just `input` and `output` felt very confusing to me, so I've used `sink-input` and `source-output` for classes.
- Added new `show_sinks` and `show_sources` config options to fully enable/disable speaker or microphone features
- Changed the `sink_slider_orientation` and `source_slider_orientation` defaults to Horizontal as it fits the default themes better.
- Tweaked the `desktop` theme to add some padding to accommodate the orientation change, and the new label on the bar.

Since I've added a few things on here myself, I'll push them up here in a sec and then look at rebasing my end to save you some headache.

JakeStanger:
Ok why on earth did force pushing close the PR and why on earth can't I reopen it???

Are you able to reopen? If not I'll merge locally

imgurbot12:
> Just had a play with this, All seems to work well!

Epic thank you. Sounds great on those changes!

> Ok why on earth did force pushing close the PR and why on earth can't I reopen it???

Uh oh. Im not sure what happened but it seems like the commit history got wiped out. Master branch of the fork says its 2 commits behind ironbar's master, and none of my changes are there. Since the PR is behind master without any unique commits, I think it auto-closed the PR. Thankfully I have a local copy I can force-push back into my fork, but It won't include your tweaks. What the heck.

imgurbot12:
I will not let me reopen the pull-request either.
I think I'm going to have to post a new one, unless you're able to reopen now @JakeStanger.

PR #1380, opened 2026-02-16
Update style.css to make volume sliders visible on minimal config

ikoas:


JakeStanger:
Could you send a screenshot of how this looks for you please? Not sure I'm seeing the intended result on my machine

ikoas:
Hi there. I'm sending some pictures I took with my phone because I still haven't figured out how to take screenshots, I hope they help.

This one is with the changes in the commit:
IMG_20260216_225837_827


And this one is without the changes:
IMG_20260216_225915_910

I'll try to figure out how to take screenshots and I'll send them.

ikoas:
Screenshot of volume popup using "minimal" config and changes in this commit:
20260217_10h59m53s_grim

Screenshot of volume popup using "minimal" config and without changes in this commit:
20260217_11h01m41s_grim

ikoas:
This is a bit unrelated to the subject of this PR but, it seems that some of the options in for `*` in the css mess up some things in the minimal config.

I'm also using the bluetooth module and the options for `*` result in the following:
20260217_14h57m48s_grim

I can wrap my head around the changes so it's not a big deal but I'm guessing that wasn't the way it was intended to look since commenting out `background-image: none;` results in a more legible popup:
20260217_14h58m58s_grim

JakeStanger:
It's clear a fix is needed, but I don't think this is it. The sliders are broken across all modules, not just volume, so the fix needs to apply to that. 

I also don't think this is terribly good looking - it introduces a new colour to the palette used exclusively for the slider background, while leaving the actual slider a thin blue colour.

Happy to accept a broader fix, but I'm going to close this one.

ikoas:
I'll try to fix all sliders, but I fear that I'll but way too much effort just to end up with design language that doesn't work with the broader theme.

Could you provide me with some style guidelines or some concrete points, for me to follow?

PR #1390, opened 2026-02-20
Update style.css to make all slider visible

ikoas:
20260220_11h55m24s_grim

### DIFF
--- a/examples/minimal/style.css
+++ b/examples/minimal/style.css
@@ -25,7 +25,7 @@ scale > trough > highlight {
     background-color: var(--color-active);
     border-style: solid;
     border-color: var(--color-active);
-    border-width: var(--margin-sm);
+    border-width: 0.3em;
 }
 
 scale > trough > slider {
--- a/examples/minimal/style.css
+++ b/examples/minimal/style.css
@@ -17,6 +17,21 @@
     font-family: monospace;
 }
 
+scale > trough {
+    background-color: var(--color-dark-secondary);
+}
+
+scale > trough > highlight {
+    background-color: var(--color-active);
+    border-style: solid;
+    border-color: var(--color-active);
+    border-width: var(--margin-sm);
+}
+
+scale > trough > slider {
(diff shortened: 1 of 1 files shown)


ikoas:
Sorry for opening a new PR. I feel this new patch is much better.

ikoas:
There is still one issue I can't wrap my head around. The GTK Scale in this module has 3 CSS nodes involved for this issue, `highlight` is the part that represents the current volume , `trough` is the bar where `highlight` moves and `slider` is the "handle".

20260220_12h09m22s_grim

As you can see if you set volume to 0 percent, the `highlight` overflows the `slider`. This would be fixed by just making the border-width of `highlight` smaller, I would suggest 0.3em.

While writing this, I noticed that I should have done this before submitting this PR. I apologize in advance.

ikoas:
I also changed the css for switches but I feel like making another PR would be stupid so I'll add the code here and a screen shot.

```
switch > slider {
    background-color: var(--color-white);
}

switch:checked {
    background-color: var(--color-active);
}
```

20260220_12h56m13s_grim

I'm still trying to figure out my way around github, I would like to add another commit to this PR to include this and the change for the size of the volume bar but I don't think it's possible.

ikoas:
Forgot to add the css to change the background color if the switch is not checked:

```
switch:not(:checked) {
  background-color: var(--color-dark-secondary);
}
```

JakeStanger:
This looks better, thank you.

You can include the other changes in this PR by just committing to your `patch-1` branch (I assume that was created through the browser - it should still work fine so long as the branch name matches).

ikoas:
> This looks better, thank you.
> 
> You can include the other changes in this PR by just committing to your `patch-1` branch (I assume that was created through the browser - it should still work fine so long as the branch name matches).

Thanks. I couldn't wrap my head around pull requests but now that you told me it makes sense

JakeStanger:
LGTM, thanks!

Issue #820, opened 2024-12-22
ext-workspace support


ToxicMushroom:
**Is your feature request related to a problem? Please describe.**
Currently the supported compositors for workspace module is very limited and requires compositor specific implementations.
This is a newly merged wayland protocol for workspaces that aims to unify this process

**Describe the solution you'd like**
Use this protocol in the workspace module. 
Ironbar can then hopefully be used to show workspaces on any comp that implements this protocol :)

I think sway already does since it was based on their wlr workspace thing but not sure.

**Describe alternatives you've considered**
/

**Additional context**
https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/40

JakeStanger:
I'd like to see both of Sway and Hyprland support the protocol first, as those are the two with merged support. Once they ship releases, I'll definitely be happy to move away from relying on their IPCs

Brisingr05:
Niri will be implementing ext-workspace in https://github.com/YaLTeR/niri/pull/1800.

JakeStanger:
Hyprland has merged support

Brisingr05:
Niri has also merged support: https://github.com/YaLTeR/niri/issues/135#issuecomment-3066726736

leoVici9:
this is the only reason why i do not use ironbar on mangowc. mangowc just merged support.

JakeStanger:
I've started working on this in the past few days, and I've gotten to the point where I can log events from the protocol (you can test this on the `feat/ext-workspaces` branch). 

After testing on Hyprland, Niri and COSMIC, it *seems* like the protocol has been (consistently) incorrectly implemented. 

I've additionally not tested Jay but from reading the code, it looks to do the same thing. LabWC looks like it uses all three states from the code, but I've not been able to get it working well enough to test.

<img width="1476" height="402" alt="Image" src="https://github.com/user-attachments/assets/c09104b3-ed8f-4cf1-974e-b8dfff02fd5b" />

<https://wayland.app/protocols/ext-workspace-v1#ext_workspace_handle_v1:enum:state>

The protocol offers 3 state values for a workspace, which I would interpret as follows:

- **Active**: The workspace is currently focused
- **Urgent**: The workspace has something requesting attention
- **Hidden**: The workspace is open, but not currently visible

Compositors seem to be implementing it as follows though:

- **Active**: The workspace is currently visible
- **Urgent**: The workspace has something requesting attention
- **Hidden**: *Not used*

[...]

PR #1403, opened 2026-02-28
feat(workspaces): style based upon workspace having windows

antlilja:
Adds the ability to style workspaces based upon it containing any windows using the `empty` css class.

Example:
```css
.workspaces .item:not(.empty) label {
    font-weight: bold;
}
```

I've only implemented this for hyprland so far. I wanted to open up this draft PR to see if this is a desired feature before continuing on implementing it for sway and niri.

### DIFF
--- a/docs/modules/Workspaces.md
+++ b/docs/modules/Workspaces.md
@@ -124,6 +124,7 @@ end:
 | `.workspaces .item.visible`    | Workspace button (workspace visible, including focused) |
 | `.workspaces .item.urgent`     | Workspace button (workspace contains urgent window)     |
 | `.workspaces .item.inactive`   | Workspace button (favourite, not currently open)        |
+| `.workspaces .item.notempty`   | Workspace button (workspace contains windows)           |
 | `.workspaces .item .icon`      | Workspace button icon (any type)                        |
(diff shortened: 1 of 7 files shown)


JakeStanger:
I'm happy for this to be added, but some consideration is needed as to how this will work with #820. 

I also think this would be better inverted btw, ie an `.empty` class. Not empty becomes the default of omitted, and you can use `:not(.empty)` where required that way, which avoids the double negative.

antlilja:
Totally agree with `empty` over `notempty`, much cleaner. I've pushed a change to switch to the `empty` css class instead. 

I took a look at [ext-workspace-v1](https://wayland.app/protocols/ext-workspace-v1) as well as the IPC for sway and niri. It does not seem to be possible to implement this feature for the wayland extension. Both Niri and Sway seem possible to implement this feature in.

JakeStanger:
Yeah I suspect the play will be having an (optional) IPC layer on top of the wayland layer to bring in the extra functionality. I don't think it'll have too much impact at this stage, but worth keeping in mind.

linde12:
would appreciate something like this, i lose myself all the time without these highlights :joy:

PR #1407, opened 2026-03-01
fix(script): scripts get terminated and launched again upon reload

### LINKED ISSUE #1102: New scripts get spawned on each config reload
**Describe the bug**

New scripts get spawned on each ironbar reload.

**To reproduce**

1. Add module `script`

For example:

```json
    {
      "cmd": "tail -f /dev/null",
      "mode": "watch",
      "type": "script"
    }
```

3. Issue an `ironbar reload` a few times
4. Observe there are many child processes instead of just one

**Expected behavior**

Either reuse the previous instance upon reload and don't create new ones, or kill it before spawning the new one.

**System information:**
- Distro: NixOS
- Compositor: Hyprland
- Ironbar version: 106e23e3116db02d57c548820024163c38491370

**Configuration**

```json
{
  "center": [
    {
      "cmd": "tail -f /dev/null",
      "mode": "watch",
      "type": "script"
    }
  ]
}
```

</details>

<details><summary>Styles</summary>

```css

```

</details>

**Additional context**

**Screenshots**

Image

esther-ff:
made the script module register and terminate scripts, recreating them on reload and preventing the problem with leftover processes.

fixes #1102

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -2,6 +2,7 @@
 #![deny(clippy::unwrap_used)]
 
 use std::cell::RefCell;
+use std::collections::HashMap;
 use std::env;
 use std::future::Future;
 use std::process::exit;
@@ -18,6 +19,7 @@ use gtk::gdk::{Display, Monitor};
 use gtk::prelude::*;
 use smithay_client_toolkit::output::OutputInfo;
 use tokio::runtime::Runtime;
+use tokio::sync::oneshot::Sender;
 use tokio::task::{JoinHandle, block_in_place};
 use tracing::{debug, error, info};
 
@@ -150,7 +152,7 @@ pub struct Ironbar {
     css_source: Rc<CssSource>,
     config_location: ConfigLocation,
     css_location: Option<ConfigLocation>,
-
+    scripts: Rc<RefCell<HashMap<String, Sender<()>>>>,
     desktop_files: DesktopFiles,
     image_provider: image::Provider,
 }
@@ -178,6 +180,7 @@ impl Ironbar {
             css_source: Rc::new(css_source),
             config_location,
             css_location,
+            scripts: rc_mut!(HashMap::new()),
             desktop_files,
             image_provider,
         }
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -348,6 +351,23 @@ impl Ironbar {
             .collect()
     }
 
+    /// Associates the command (`cmd`) of a script with a sender meant to
+    /// remotely kill the process.
+    ///
+    /// When ran it will terminate the previous script (if present) and register
+    /// this as the "new" script.
+    pub fn register_or_terminate_script(&self, cmd: &str, send: Sender<()>) {
```
JakeStanger: `register_or_terminate` makes it sound like the method does one or the other, when actually it does both. I think `register_script` is fine, so long as the method comment details the termination behaviour.

I also think calling the `Sender` just `send` is a bit vague. Perhaps `tx_terminate` or `tx_kill` would be more appropriate?

src/script.rs
```diff
@@ -193,24 +194,36 @@ impl Script {
...
-                ScriptMode::Poll => match self.get_output(args).await {
+        let (send, recv) = channel::<()>();
+        std::mem::forget(send);
+        self.run_with_recv(args, recv, callback).await
+    }
+
+    /// Runs the script, passing `args` if provided.
+    /// Uses the flag to determine if the script should be terminated.
+    /// Runs `f`, passing the output stream and whether the command returned 0.
+    pub async fn run_with_recv<F>(&self, args: Option<&[String]>, recv: Receiver<()>, callback: F)
```
JakeStanger: Like with the sender, I'd prefer to call the receiver something like `rx_terminate` or `rx_kill`. There's a handful of other instances throughout too, if you could review those please.

Could you also pls update the method doc to detail the new receiver behaviour, as this method is part of the public API?

JakeStanger (changes requested):
Thanks for this. Implementation looks good, but I think some of the naming needs a little attention.

src/script.rs
esther-ff: done! all methods that utilized the oneshot sender or receiver got some documentation on what they do and the arguments were renamed to `tx_terminate` and `rx_terminate` whenever appropriate.

JakeStanger (approved):
Thanks

PR #1410, opened 2026-03-03
feat(battery): add messages for all battery states

### LINKED ISSUE #1310: Battery: show "something" when charging and battery full

The battery popup is empty when the charger is plugged in and the battery is fully charged. It's confusing.

Something should be shown in the popup in this state. A text like "Fully charged" or something similar to indicate what state the battery / power is in.

**Additional context**

OT perhaps but I'd like to see the popup info without clicking on the icon.

tobias47n9e:
Fixes #1310


JakeStanger (approved):
Thanks, lgtm

PR #1424, opened 2026-03-17
fix(network-manager): not accepting char to display icons

### LINKED ISSUE #1423: Icon Network Manager
**Describe the bug**
Setting a character (Nerd font characters for example, 󰱓) does not set in the module widget. The widget tries to resolve it as a system icon.

**To reproduce**

1. Add module `network_manager`
2. Set 󰱓 as an icon for a profile
3. Observe a blank or default icon (behavior is inconsistent across reboots and reloads)

**Expected behavior**
Setting 󰱓 as an icon should work. This is already the case elsewhere for example in the volume module where  is a valid icon to set.

**System information:**
- Distro: ArchLinux
- Compositor: Niri
- Ironbar version: ironbar-git

MrFantOlas:
Fixes #1423

As it is my first PR here it is hopefully not too bad.
Also note that I had no previous Gtk experience so if I did something atrocious, please point it out.

Questions that still remains:
Should all icons contain "icon:" ? From what I understand specifying a path should also be accepted. Maybe a better condition there would be good or a rule to differentiate icons from custom text.

Opening this as draft since I could not find a satisfying answer to that question.

### DIFF
--- a/src/image/gtk.rs
+++ b/src/image/gtk.rs
@@ -127,6 +127,7 @@ impl Deref for IconButton {
     feature = "menu",
     feature = "music",
     feature = "workspaces",
+    feature = "network_manager"
 ))]
 pub struct IconLabel {
     provider: image::Provider,
@@ -144,7 +145,8 @@ pub struct IconLabel {
     feature = "keyboard",
     feature = "menu",
     feature = "music",
-    feature = "workspaces"
+    feature = "workspaces",
+    feature = "network_manager"
 ))]
 impl IconLabel {
     pub fn new(input: &str, size: i32, image_provider: &image::Provider) -> Self {
@@ -241,7 +243,8 @@ impl IconLabel {
     feature = "keyboard",
     feature = "menu",
     feature = "music",
-    feature = "workspaces"
+    feature = "workspaces",
+    feature = "network_manager"
 ))]
 impl Deref for IconLabel {
     type Target = gtk::Box;
--- a/src/image/mod.rs
+++ b/src/image/mod.rs
@@ -7,7 +7,8 @@
     feature = "menu",
     feature = "music",
     feature = "notifications",
-    feature = "workspaces"
+    feature = "workspaces",
+    feature = "network_manager"
 ))]
 mod gtk;
 mod provider;
@@ -21,7 +22,8 @@ mod provider;
     feature = "menu",
     feature = "music",
     feature = "notifications",
-    feature = "workspaces"
+    feature = "workspaces",
+    feature = "network_manager"
 ))]
 pub use self::gtk::*;
 pub use provider::Provider;
--- a/src/modules/networkmanager.rs
+++ b/src/modules/networkmanager.rs
@@ -1,17 +1,19 @@
 mod config;
 
+use std::rc::Rc;
+
 use crate::channels::{AsyncSenderExt, BroadcastReceiverExt};
 use crate::clients::networkmanager::state::DeviceTypeData;
 use crate::clients::networkmanager::{Client, DeviceType, NetworkManagerUpdate};
 use crate::gtk_helpers::IronbarGtkExt;
-use crate::image::Provider;
+use crate::image::IconLabel;
 use crate::modules::{Module, ModuleInfo, ModuleParts, WidgetContext};
 use crate::{module_impl, spawn};
 
 use color_eyre::Result;
+use gtk::Box as GtkBox;
 use gtk::prelude::WidgetExt;
 use gtk::prelude::*;
-use gtk::{Box as GtkBox, ContentFit, Picture};
 use tokio::sync::mpsc::Receiver;
 
 pub use config::NetworkManagerModule;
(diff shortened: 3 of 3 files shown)


JakeStanger:
> Should all icons contain "icon:" ?

Not if it's coming from the config. The system is designed to allow arbitrary text if preferred, or parse as different formats (`http(s)://`, `file://`, ...).

MrFantOlas:
I would think that adding a custom key, like "raw:", would be easier to parse to display a label instead of an icon. This way we do not have to wait and see if the string resolves before showing as a label. This also avoids showing an unresolved path in the label.

I do not know however since this module is the only module that currently allows svg icons (from what I have read of the code) that this would be later expanded to other modules (and then maybe a better handling to avoid destroying peoples configs).

JakeStanger:
I can see potential utility in adding a new `raw:` prefix to skip image parsing, but I'm not sure I'm following the rest, nor do I see how that ties into this particularly. 

> This also avoids showing an unresolved path in the label.

What do you mean by this?

> this module is the only module that currently allows svg icons

All modules use the central image parsing system, which supports SVGs.

MrFantOlas:
> What do you mean by this?

I meant that when the path turns out unresolved you could assume it was not a path and show it in the label. But that seems like an horrible solution: you could get full paths in the bar, it does not give a clue to what happened, you have to wait before showing a label that the resolution is done.

> All modules use the central image parsing system, which supports SVGs.

I am not sure about this. For example I was not able to parse an SVG image in the volume module. I was also unable to use the low and medium profiles there with the TOML config so maybe there is something that I missed there ?

Anyway I think I will try to make a clean "raw:" etiquette system as it is not far of from what I have already written.

JakeStanger:
> I meant that when the path turns out unresolved you could assume it was not a path and show it in the label.

This isn't the current behaviour though. If you pass a path with `file://`, it will fail to resolve the image, show blank and log an error, eg:

```
 WARN ironbar::image::provider: 143: failed to load image: zen: 
   0: Error opening file /does-not-exist.png: No such file or directory
```

> For example I was not able to parse an SVG image in the volume module

Either a config issue or a bug. It's definitely *supposed* to be supported anywhere images are supported. There's specific support for it here:

https://github.com/JakeStanger/ironbar/blob/6e05d5516ef2dd4b1dfb390e10aa1057c9393646/src/image/provider.rs#L254-L264

> I was also unable to use the low and medium profiles there with the TOML config so maybe there is something that I missed there ?

What do you mean by "use"? Regardless they're definitely there, and they definitely work.

MrFantOlas:
To be clear applying this config 

```toml
[[end]]
type = "volume"
sink_slider_orientation = "horizontal"
source_slider_orientation = "horizontal"
format = " {icon} {percentage}% {name} "
max_volume = 100

[end.icons]
volume_high = "T"
volume_medium = "T"
volume = "icon:network-wired-symbolic"
volume_low = "T"
muted = ""
```

Only changes the high volume output to be a string of "icon:network-wired-symbolic". This is may be the wiki not being up to date. I have seen that in the code there are profiles implemented so there _should_ be a way to make it definitely work. I think it would be best to have this discussion in another issue maybe (which I can open later).

The relevant part for this PR is that "icon:network-wired-symbolic" does not get translated to an SVG icon. I mention this because the method you linked is in used in the following modules: `networkmanager`, `custom`, `launcher`, `menu`, `music` and `tray`. (Searching for a reference of `load_into_picture(_silent)` as it is the only method that calls `get_paintable`)

> All modules use the central image parsing system, which supports SVGs.

So I am unsure what you meant by all modules ?

I mention this because then it might be better to have a shared system (the `:raw` prefix) that applies to all of these modules and at the same time extend the support of SVG icons to other modules ?

JakeStanger:
> This is may be the wiki not being up to date

Ah yeah you're right. I'll log that to get fixed as soon as I can.

> it might be better to have a shared system (the :raw prefix) that applies to all of these modules and at the same time extend the support of SVG icons to other modules ?

This is what already exists, which is what all image loading across all the modules already uses. It already supports SVGs, and that's what I linked above.

MrFantOlas:
Ok I finally understood what you meant, sorry for having dragged this.

I think then that only the `networkmanager` module would benefit from this, therefore I can open this now :)

JakeStanger:
Okay sorry this is the first time I've properly looked at the code, but I think this needs reworking. 

There already exists an `IconLabel` struct which implements this logical, albeit inverted - it looks for an explicit prefix (`icon:`, `file://`, `https://`), and treats it as the raw string if none are present.

https://github.com/JakeStanger/ironbar/blob/6e05d5516ef2dd4b1dfb390e10aa1057c9393646/src/image/gtk.rs#L192-L199

Rather than implementing a new system, using that would be preferable

MrFantOlas:
No worries. I have rewritten using the `IconLabel`.

I am not the most proficient with the gtk lib and the api so hopefully them being marked as immutable is true.

JakeStanger:
Cool thanks, code lgtm. Can you fix the build issue please? Just a feature flag that needs adjusting.

Once that's done I'll give this a quick test and make sure it's all good, and go through wine a fine comb, then I think we're good to go.

> I am not the most proficient with the gtk lib and the api so hopefully them being marked as immutable is true.

`gtk-rs` provides safe bindings so you can trust the compiler. If it doesn't ask for mutability, you're good. GObject structs are basically just pointers that get used for FFI, so they don't get modified in rust-land and can be cloned cheaply.

MrFantOlas:
> Just a feature flag that needs adjusting.

Yes I can see now. I have added it !

Edit: That was not the case, now it is. I had no idea you had to log into GitHub to see job details.

> gtk-rs provides safe bindings so you can trust the compiler. If it doesn't ask for mutability, you're good. GObject structs are basically just pointers that get used for FFI, so they don't get modified in rust-land and can be cloned cheaply.

That's good to know :eyes:

JakeStanger (approved):
Briliant thanks, lgtm

Issue #964, opened 2025-05-09
(Re upload) Worspace buttons disappear


lordmujin:
**Describe the bug**
When switching workspaces, after hyprland and ironbar start, if I switch to another workspace, the one I was at the start vanishes. If I do `ironbar reload` it goes back to normal but the behavior repeats.

Duplicate of: https://github.com/JakeStanger/ironbar/issues/464
 
**To Reproduce**
Set favorite workspaces, start hyprland in this case, switch workspaces, and sometimes it works correctly, but fails more often.

**Expected behavior**
That all favorite workspaces are persistent

**System information:**
 - Distro: NixOS
- Compositor: Hyprland
 - Ironbar version: 0.16.1 (nix-unstable)

**Configuration**

<details><summary>Config</summary>

```
{ "type": "workspaces",
		  "name_map": {
  		  	"1": "一",
  		  	"2": "二",
  		  	"3": "三",
  		  	"4": "四",
  		  	"5": "五"
  		  },
		  "favorites": ["1", "2", "3", "4", "5"],
		  "hidden": ["special:scratch_music", "special:scratch_term", "special:special"]
  		}
```
</details>

<details><summary>Styles</summary>

```css
.workspaces {
	margin-left: 10px;
}

.workspaces .item.focused {
	color: @blue;
	background-color: @gray-dark;
    box-shadow: inset 0 -3px;
}

.workspaces .item.urgent {
    background-color: @red;
}

.workspaces .item:hover {
	color: @sky;
	background-color: @base;
    box-shadow: inset 0 -3px;
}
```
</details>

**Additional context**
[](https://github.com/JakeStanger/ironbar/issues/464)

**Screenshots**

psynyde:
Facing the same issue with same specs. (same os and ironbar version)

JakeStanger:
I've not been able to replicate this. I'm going to need some more information around the repro steps please. 

Specifics like *which* workspaces you're switching to/from, whether there's anything open on those workspaces, etc are very helpful.

If you can also please capture debug logs using:
```
IRONBAR_LOG='ironbar::modules::workspaces=debug,ironbar::clinets::compositor=debug'
```
that will hopefully give some insight into what's happening under the hood.

lordmujin:
Thanks for the tip, its a odd behavior, I noticed it happens when the workspace has no window open, and it can happen in any.
[...]

PR #1439, opened 2026-04-02
fix: issue with favorite workspace buttons not working for inactive/empty workspaces #964

CameronBarnes:
Favorite workspace buttons use a default workspace id value of -1 which is not generally an active or valid workspace and does not usually align with numbered favorite workspaces.
Parse the name into a number where possible to resolve that issue, and adjust how removing workspace buttons occurrs to prevent this from being undone.
A separate fix is likely required to address issues with favorite workspaces with non-numerical names.

### DIFF
--- a/src/modules/workspaces/mod.rs
+++ b/src/modules/workspaces/mod.rs
@@ -16,6 +16,8 @@ use gtk::prelude::*;
 use serde::Deserialize;
 use std::cmp::Ordering;
 use std::collections::HashMap;
+use std::ops::Deref;
+use std::rc::Rc;
 use tokio::sync::mpsc;
 use tracing::{debug, trace, warn};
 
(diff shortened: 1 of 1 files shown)


CameronBarnes:
Originally I was trying to find the source of the signals with -1 values instead of the solution I've used here with the removal function. But I could not for the life of me figure out where it was coming from. The implementation here is probably correct anyway, though in future both might be ideal. 

The use of RC for the favorites is definitely unnecessary given the small memory footprint in the vast majority of cases, but the optimization is unlikely to hurt. It could potentially be expanded back to the original collection of favorites if the benefit was considered worth the small increase in code complexity. 

Handling for named workspaces is definitely possible, and I have a few ideas for approaches, but it's likely to be specific to the particular DE. Where as I suspect this numbed fix will resolve issues in more than just Hyprland, though i haven't tested it with other desktops personally.

JakeStanger:
> I have a few ideas for approaches, but it's likely to be specific to the particular DE

I'd be cautious of spending too much time on anything compositor-specific at this point, as the aim is to move to the `ext-workspaces` protocol soon.

JakeStanger (approved):
Thanks, happy with this. It'd probably be cheap enough to clone without the `Rc` since it's only ever cloned once, but I don't think it'll make any measurable difference either way.

PR #1449, opened 2026-04-12
Add cargo-binstall metadata

meghprkh:
This allows the package to be easily installable with `cargo-binstall`

This can be smoke tested with

```shell
cargo binstall --pkg-url="{ repo }/releases/download/v{ version }/{ name }-v{ version }-{ target-arch }{ archive-suffix }" ironbar
# Optionally use arguments: --force -v --dry-run
```


JakeStanger (approved):
Thanks

PR #1451, opened 2026-04-13
Expands tilde in cairo module path

sanpii:
It's a little confusing not to be able to use tilde un cairo module path when it’s work fine with custom modules.


JakeStanger (approved):
lgtm, thanks

PR #1459, opened 2026-04-18
feat(clock): add parameter to set gtk_calendar_set_show_week_numbers

janquednau:
Just a simple addition to the clock module implementing a GTK Calendar instance method.


JakeStanger (approved):
Thanks, lgtm.

Build failure is due to unrelated clippy warning introduced in latest rust version.

PR #1463, opened 2026-04-20
Various profile fixes

### LINKED ISSUE #1371: Built-in brightness profiles conflict with user configuration
As stated in the [wiki](https://github.com/JakeStanger/ironbar/wiki/brightness#icons), the new brightness module ships with default profiles which set the icon according to the percentage. This seems to override the user-specified `format` property: in particular, my use case is showing the number _without_ the percentage symbol (I use a vertical bar, so having the "%" symbols takes up unnecessary space and I find it quite ugly), but something like this doesn't work:

```toml
[[end]]
type = "brightness"
format = "{percentage}"
justify = "center"

[end.mode]
type = "systemd"
subsystem = "backlight"
```

In fact, I believe that the `format` option is overridden by the profiles which set the icons. Interestingly though I found out that the only way to override these profiles is by creating a new one _for each of the levels mentioned in the wiki_: for instance, something like the following only works when the percentage is exactly 30, everything below will still have the built-in formatting with the percentage symbol.

```toml
[[end]]
type = "brightness"
format = "{percentage}"
justify = "center"

[end.mode]
type = "systemd"
subsystem = "backlight"

[end.profiles.30]
when = 30
format = "\n{percentage}"
```

I hope this could be fixed, and thanks for the amazing project!

JakeStanger:
- Adds a new `use_default_profiles` option (`true` by default) which can be disabled to avoid using the default profiles in situations where they get in the way.
- Adjusts battery profile matcher so that the default percent is `100`, allowing it to be omitted and match purely on `charging`.
- Improves documentation

Resolves #1371


docs/modules/Brightness.md
```diff
@@ -5,14 +5,15 @@ Allows to change the respective value via scrolling.
...
+|------------------------|-------------------------|-----------------|----------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `format`               | `string`                | `{percentage}%` | Yes      | Format string to use for the widget button label.                                                                                                                                                 |
+| `icon_label`           | `string`                | `null`          | Yes      | Icon to show alongside the label. Supports [image](images) icons.                                                                                                                                 |
+| `mode.type`            | `systemd` or `keyboard` | `systemd`       | No       | The data backend of the brightness module, this can be either the KdbBrightness dbus which is good for keyboard data, or the more general login1 dbus in combination with /sys/class/<subsystem>. |
```
JakeStanger: `<subsystem>` is detected as an html tag here and not rendered

PR #1465, opened 2026-04-22
`ext-workspace` support

### LINKED ISSUE #820: ext-workspace support

JakeStanger:
Adds a new backend for the `workspaces` module which uses the Wayland `ext-workspace` protocol instead of compositor-specific IPC.

This still needs some work to tidy up the code, but is in a working state and good to be tested.

Resolves #820

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1917,6 +1917,7 @@ dependencies = [
  "tracing-subscriber",
  "walkdir",
  "wayland-client",
+ "wayland-protocols",
  "wayland-protocols-wlr",
  "zbus",
 ]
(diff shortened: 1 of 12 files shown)


Kommynct:
https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/486 appears to be getting merged, may want to look into this

JakeStanger:
That's great news. That should remove the dependency on needing IPC to fill the gap.

PR #1471, opened 2026-04-27
feat(volume): add truncate_formats option

### LINKED ISSUE #1422: Add volume bar truncate option
**Describe the bug**
Using truncate in the volume module does not truncate format but the popup window of the widget. This not specified anywhere in the documentation.

**To reproduce**

1. Add module `volume`
2. Add truncate mode and max_length
3. Add a format with the name of the device: "{icon} {name}"
4. Observe the format not being truncated

**Expected behavior**
I would expect that there would be a way to truncate the format since it can take quite a bit of space in the bar. I would expect that this would not truncate the names of the sinks in the module popup window.

**System information:**
- Distro: ArchLinux
- Compositor: Niri
- Ironbar version: ironbar-git

MrFantOlas:
Closes #1422
Adds `truncate_formats` to the wiki.
Specifies that `truncate` also applies to sink and sources in the volume mixer.

### DIFF
--- a/docs/modules/Volume.md
+++ b/docs/modules/Volume.md
@@ -1,7 +1,9 @@
 Displays the current volume level.
 Clicking on the widget opens a volume mixer, which allows you to change the device output level,
 the default playback device, and control application volume levels individually.
-Use `truncate` or `marquee` options to control the display of application titles in the volume mixer.
+Use `truncate_popup` or `marquee` options to control the display of application titles in the volume mixer.
+Use `truncate_popup` option to control the display of sink and sources in the volume mixer.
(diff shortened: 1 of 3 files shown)


JakeStanger:
Thanks, looks good. I'm thinking it would be better to effectively switch the names though, so that `truncate` refers to the bar content. Perhaps `truncate_popup` for the existing setting?

MrFantOlas:
> Perhaps `truncate_popup` for the existing setting?

I agree. I did not want to make breaking changes but if it makes more sense.

JakeStanger (approved):
Cool thanks, lgtm.

Generally I take the approach that breaking changes now are better than inconsistency later.

Issue #1166, opened 2025-09-17
Volume module does not respect default sink


bandithedoge:
**Describe the bug**
I'm using PipeWire's PA server implementation. I have two available output devices, an audio interface and a filter chain node (which is set as the default output). When starting Ironbar with the volume module enabled, it instantly sets the audio interface as the default output and reroutes all running apps.

**To reproduce**

1. Create a Filter-Chain node in PipeWire and set it as the default output
2. Add module `volume`
3. Run Ironbar
4. Observe the audio output changes

**Expected behavior**
Volume module keeps the default audio output selected.

**System information:**
- Distro: NixOS
- Compositor: Niri
- Ironbar version: master (0.17.0)

**Configuration**

<details><summary>Config</summary>
```
            {
              type = "volume";
              format = "${icon "{icon}"} {percentage}%";
              icons = {
                volume_high = "󰕾";
                volume_medium = "󰖀";
                volume_low = "󰕿";
                muted = "󰝟";
              };
              on_scroll_up = "wpctl set-volume @DEFAULT_AUDIO_SINK@ 1%+";
              on_scroll_down = "wpctl set-volume @DEFAULT_AUDIO_SINK@ 1%-";
            }
```
</details>

JakeStanger:
Not fully resolved, but #1475 should go a way to helping this

JakeStanger:
I think #1482 should fully resolve this now, so going to close. If this still occurs let me know.

PR #1475, opened 2026-04-28
fix(volume): volume respects default sink and source

MrFantOlas:
Should help #1166.
This prevents the drop menu from selecting the first appended item and then trying to set it as the default sink.

Edit: This appends when the list is empty, then one item is appended to the list. This results in the notify event being triggered and then the sink to change to that first sink appended. Same for source.

The drop menu now has a default state where an item No Sink or No Source is preappended to it. This prevents setting the default sink multiple times resulting in inconsistent sink selection (or source).

The following section works for sink and sources.
This now only fires a single notify event to select the default sink sent by PA. If PA sends no default sink, i.e. no`SinkUpdate` then the default no sink available is shown. If this update is send and the sink is valid this option is removed.

I tested this restarting the bar multiple time back to back s this would facilitate the race condition where my sink or source would get changed. This also prevents the bar switching aggressively between two sinks at startup, and may also help with #1411.

It is to be noted that selecting no item in the drop down menu does not work, and this may also be a bug on the gtk side. I though I would offer a solution as I am unsure if this is the default behavior of the drop down menu (the documentation is minimal and I am not very proficient in gtk to say the least).

### DIFF
--- a/src/modules/volume/mod.rs
+++ b/src/modules/volume/mod.rs
@@ -610,7 +610,7 @@ impl Module<Button> for VolumeModule {
                                 .append(&DropdownItem::new(&info.name, &info.description));
 
                             if info.active {
-                                let offset = sink_options.find(&no_sink).is_some() as usize;
+                                let offset = source_options.find(&no_source).is_some() as usize;
                                 source_selector.set_selected((sources.len() + offset) as u32);
                                 source_slider.set_value(info.volume.percent());
 
--- a/src/modules/volume/mod.rs
+++ b/src/modules/volume/mod.rs
@@ -126,7 +126,7 @@ impl VolumeModule {
     where
         F: Fn(String) -> Update + 'static,
     {
-        selector.connect_selected_notify(move |selector| {
+        selector.connect_selected_item_notify(move |selector| {
             if let Some(item) = selector.selected_item().and_downcast_ref::<DropdownItem>() {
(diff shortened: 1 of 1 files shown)


src/modules/volume/mod.rs
```diff
@@ -604,7 +610,8 @@ impl Module<Button> for VolumeModule {
                                 .append(&DropdownItem::new(&info.name, &info.description));
 
                             if info.active {
-                                source_selector.set_selected(sources.len() as u32);
+                                let offset = sink_options.find(&no_sink).is_some() as usize;
```
JakeStanger: think this should be:

```suggestion
                                let offset = source_options.find(&no_source).is_some() as usize;
```

JakeStanger (changes requested):
Thanks, look good. Think there's one typo from a copy/paste job

src/modules/volume/mod.rs
MrFantOlas: Indeed that one managed to slip through

MrFantOlas:
Tell me if you want me to squash all of this before merging

JakeStanger (approved):
lgtm, thanks. I'll squash it during merge

ffs-akl:
I tried,  no help with https://github.com/JakeStanger/ironbar/issues/1411.

PR #1481, opened 2026-05-09
fix(battery): sysfs based threshold battery time calculation

slowsage:
upower time to full does not take into account threshold caps (sometimes 80% to prolong battery life). This reads from sysfs to account for one time 100% charge limits as well.


JakeStanger:
Thanks, looks good. Does this account for discharging too? Can't quite wrap my head around the logic, but obvs when discharging the cap should be ignored.

slowsage:
I think that is stored as `time_to_empty`. This only modifies `time_to_full`.

JakeStanger:
Ah of course yeah. All good then :)

JakeStanger (approved):

PR #1486, opened 2026-05-11
feat: system-wide /etc/ configuration support

victorvintorez:
Refactors Config and CSS Source resolution out of config loader and into a separate `sources.rs` file, and adds `/etc/` as a system-wide configuration source.

Config Resolution order:
1. `IRONBAR_CONFIG` / `IRONBAR_CSS` env vars
2. `--config` / `--theme` CLI args
3. `$XDG_CONFIG_HOME/ironbar/` user level config
4. `/etc/ironbar/` system level config
5. Builtin Minimal config fallback

### DIFF
--- a/docs/Configuration guide.md
+++ b/docs/Configuration guide.md
@@ -15,6 +15,8 @@ The examples make use of [Nerd Fonts](https://www.nerdfonts.com/#home) for displ
 
 The config file lives inside the `ironbar` directory in your XDG_CONFIG_DIR, which is usually `~/.config/ironbar`.
 
+Alternatively, you can use a system-level config in `/etc/ironbar/` to set a default config for all users on the machine.
+
 Ironbar supports a range of configuration formats, so you can pick your favourite:
 
 - `config.json`
@@ -23,7 +25,7 @@ Ironbar supports a range of configuration formats, so you can pick your favourit
 - `config.corn` (Includes variable support for re-using blocks.
   See [here](https://github.com/jakestanger/corn) for info)
 
-You can also override the default config path using the `IRONBAR_CONFIG` environment variable.
+You can also override the default config path using the `IRONBAR_CONFIG` environment variable, or the `--config` CLI argument.
 
 A hosted schema is available for the latest Git version ~~and each versioned release~~.
 JSON and YAML both support schema checking by adding the `$schema` key 
--- a/src/config/sources.rs
+++ b/src/config/sources.rs
@@ -12,16 +12,7 @@ pub enum Builtin {
 }
 
 impl Builtin {
-    pub fn css(&self) -> &'static str {
-        match self {
-            Self::Minimal => include_str!("../../examples/minimal/style.css"),
-            Self::Desktop => include_str!("../../examples/desktop/style.css"),
-        }
-    }
-}
-
-#[cfg(feature = "config")]
-impl Builtin {
+    #[cfg(feature = "config")]
     pub fn config(&self) -> (&'static str, config::FileFormat) {
         cfg_if! {
             if #[cfg(feature = "config+corn")] {
@@ -45,6 +36,13 @@ impl Builtin {
                     Self::Desktop => (include_str!("../../examples/desktop/config.toml"), FileFormat::Toml)
                 }
             }
+        }
+    }
+
+    pub fn css(&self) -> &'static str {
+        match self {
+            Self::Minimal => include_str!("../../examples/minimal/style.css"),
+            Self::Desktop => include_str!("../../examples/desktop/style.css"),
         }
     }
 }
@@ -74,11 +72,9 @@ pub enum CssSource {
 pub fn resolve_sources(
     cli_config: Option<ConfigLocation>,
     cli_css: Option<ConfigLocation>,
-    env_config: Option<String>,
-    env_css: Option<String>,
 ) -> (ConfigSource, CssSource) {
-    let config = resolve_config(cli_config, env_config);
-    let css = resolve_css(cli_css, env_css, &config);
+    let config = resolve_config(cli_config);
+    let css = resolve_css(cli_css, &config);
     (config, css)
 }
 
@@ -86,21 +82,13 @@ pub fn resolve_sources(
 pub fn resolve_sources(
     cli_config: Option<ConfigLocation>,
     cli_css: Option<ConfigLocation>,
-    env_config: Option<String>,
-    env_css: Option<String>,
 ) -> (ConfigSource, CssSource) {
     panic!(
         "Ironbar has been configured without config support. This won't work. Please reconfigure with at least one `config` feature flag enabled."
     )
 }
 
-fn resolve_config(cli: Option<ConfigLocation>, env: Option<String>) -> ConfigSource {
-    // IRONBAR_CONFIG env var
-    if let Some(path) = env {
-        return ConfigSource::File(PathBuf::from(path));
-    }
-
-    // --config CLI arg
+fn resolve_config(cli: Option<ConfigLocation>) -> ConfigSource {
     if let Some(loc) = cli {
         return match loc {
             ConfigLocation::Minimal => ConfigSource::Builtin(Builtin::Minimal),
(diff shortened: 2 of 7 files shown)


docs/Configuration guide.md
JakeStanger, line 1: Most of the diff in this file is unwanted formatting changes. Please revert those

docs/Configuration guide.md
```diff
@@ -13,7 +13,8 @@ The examples make use of [Nerd Fonts](https://www.nerdfonts.com/#home) for displ
 
 ## 1. Create config file
 
-The config file lives inside the `ironbar` directory in your XDG_CONFIG_DIR, which is usually `~/.config/ironbar`.
+The config file lives inside the `ironbar` directory in your XDG_CONFIG_DIR, which is usually `~/.config/ironbar`. Alternatively, you can use a system level config in `/etc/ironbar` to set a default config for all users
+on the machine.
```
JakeStanger: Nit but I prefer to put md line breaks between clauses as it makes reading the source easier:


```suggestion
The config file lives inside the `ironbar` directory in your XDG_CONFIG_DIR, which is usually `~/.config/ironbar`. 
Alternatively, you can use a system level config in `/etc/ironbar` to set a default config for all users on the machine.
```

docs/Configuration guide.md
```diff
@@ -24,13 +25,15 @@ Ironbar supports a range of configuration formats, so you can pick your favourit
   See [here](https://github.com/jakestanger/corn) for info)
 
 You can also override the default config path using the `IRONBAR_CONFIG` environment variable.
+You can also override the default config using the `IRONBAR_CONFIG` environment variable, or `--config` CLI
+argument.
```
JakeStanger: This has been duped

src/config/sources.rs
```diff
@@ -0,0 +1,193 @@
...
+            }
+        }
+    }
+}
+
+fn config_source_exists(base: &Path) -> bool {
+    config::Config::builder()
+        .add_source(config::File::from(base))
+        .build()
+        .is_ok()
```
JakeStanger: This is a very expensive check just to see if the file exists, as it loads the config and then throws it away. 

Since we know the supported file extensions, you can just iterate through these.

src/config/sources.rs
```diff
@@ -0,0 +1,193 @@
...
+    Builtin(Builtin),
+    File(PathBuf),
+}
+
+#[cfg(feature = "config")]
+pub fn resolve_sources(
+    cli_config: Option<ConfigLocation>,
+    cli_css: Option<ConfigLocation>,
+    env_config: Option<String>,
+    env_css: Option<String>,
```
JakeStanger: Handling env vars separately is not needed, `clap` does this for us:

https://github.com/JakeStanger/ironbar/blob/d44268bdf050826f4246d6c5d9f1a920ea7d9686/src/cli.rs#L30-L35

src/main.rs
```diff
@@ -150,24 +169,15 @@ pub struct Ironbar {
...
-            if #[cfg(feature = "config")] {
-                let (mut config, css_source, _) =
-                    Config::load(config_location.clone(), css_location.clone());
-            } else {
-                let (mut config, css_source) =
-                    Config::load(config_location.clone(), css_location.clone());
-            }
-        );
+    fn new(config_source: ConfigSource, css_source: CssSource) -> Self {
+        let (mut config, _) = Config::load(&config_source);
```
JakeStanger, line 172: Is the `css_source` return value still used anywhere else?

src/config/sources.rs
```diff
@@ -0,0 +1,193 @@
...
+                        "styles at '{}' not found, falling back to minimal theme",
+                        style.display()
+                    );
+                    CssSource::Builtin(Builtin::Minimal)
+                }
+            }
+        };
+    }
+
+    // CSS from Config dir
```
JakeStanger: This comment is inaccurate & adds nothing. Same applies with a lot of the comments above.

General advice is that outside of documentation, comments are only necessary to describe why, not what as you can already tell 'what' from the code. Often if you find yourself needing to write a comment as to why something is written the way it is, the better option is to refactor it.

Good vid on this:
https://youtu.be/Bf7vDBBOBUA

src/config/sources.rs
```diff
@@ -0,0 +1,193 @@
...
+        match self {
+            Self::Minimal => include_str!("../../examples/minimal/style.css"),
+            Self::Desktop => include_str!("../../examples/desktop/style.css"),
+        }
+    }
+}
+
+#[cfg(feature = "config")]
+impl Builtin {
+    pub fn config(&self) -> (&'static str, config::FileFormat) {
```
JakeStanger: This doesn't need to be in a separate `impl` block

JakeStanger (changes requested):
Thanks, this is a good addition & the code looks a lot cleaner than it was. I've raised a few comments on first pass but nothing serious.

src/main.rs
victorvintorez: css_source is loaded into the Ironbar struct, cloned here:
https://github.com/JakeStanger/ironbar/blob/d44268bdf050826f4246d6c5d9f1a920ea7d9686/src/main.rs#L200

then passed to load_css here:
https://github.com/JakeStanger/ironbar/blob/d44268bdf050826f4246d6c5d9f1a920ea7d9686/src/main.rs#L224

docs/Configuration guide.md
victorvintorez: reverted the commit with the formatting changes

docs/Configuration guide.md
victorvintorez: fixed in docs commit

src/config/sources.rs
victorvintorez: replaced with feature gated checks for each extension

docs/Configuration guide.md
victorvintorez: fixed in docs commit

src/config/sources.rs
victorvintorez: removed env_config and env_css from the various resolve functions

src/main.rs
JakeStanger: Sorry my bad I meant `css_location`, as in the 2nd return value of `Config::load` that's discarded here

src/main.rs
```diff
@@ -64,8 +64,11 @@ fn main() {
         if #[cfg(feature = "cli")] {
             run_with_args();
         } else {
-            let config_location = ConfigLocation::from_env("IRONBAR_CONFIG").unwrap_or_default();
-            start_ironbar(false, config_location, ConfigLocation::from_env("IRONBAR_CSS"));
+            let (config_source, css_source) = resolve_sources(
+                None,
+                None,
```
JakeStanger: Passing `None` for both values here means it will never resolve anything. When CLI is disabled, you do want to use the environment variables directly.

src/main.rs
JakeStanger: No ignore me I'm talking absolute gibberish

src/config/sources.rs
```diff
@@ -0,0 +1,188 @@
...
+        return true;
+    }
+
+    #[cfg(feature = "config+json")]
+    if base.with_extension("json").exists() {
+        return true;
+    }
+
+    #[cfg(feature = "config+yaml")]
+    if base.with_extension("yaml").exists() {
```
JakeStanger: We also need to check for `.yml` here

src/config/sources.rs
```diff
@@ -0,0 +1,188 @@
...
+#[derive(Debug, Clone)]
+pub enum CssSource {
+    Builtin(Builtin),
+    File(PathBuf),
+}
+
+#[cfg(feature = "config")]
+pub fn resolve_sources(
+    cli_config: Option<ConfigLocation>,
+    cli_css: Option<ConfigLocation>,
```
JakeStanger: I'm not a fan of the `cli` prefix because that's not guaranteed to be the source (arguably `resolve_sources` could be called with parameters from anywhere). Can those be dropped here please:

```suggestion
    config: Option<ConfigLocation>,
    css: Option<ConfigLocation>,
```

There's also a few args below that use `cli` as the name, if those can be addressed too please.

JakeStanger (changes requested):
Thanks, just a few more things.

If you feel up for it, some basic tests for the resolver functions would be great.

JakeStanger (commented):
Code lgtm now, cheers. I'll give this a test & approve if all good.

I'll be aiming to merge this for 0.20.0 so it may be a while, depending on how long 0.19.1 takes

JakeStanger (approved):

PR #1493, opened 2026-05-17
chore(nix): add rust-analyzer to devshell

esther-ff:
also slightly changed how the flake fetches packages for the devshell to be consistent with the ironbar package


JakeStanger (approved):
lgtm thanks

PR #1495, opened 2026-05-19
feat(nix): allow null package in nix module

davidaf3:
This PR makes `programs.ironbar.package` nullable so the nix module can be used along with a package installed from any other package manager. I have added the necessary checks to keep it backwards compatible


JakeStanger:
Thanks. Build is failing because I've run out of minutes on Garnix this month. I'll test manually when I can

JakeStanger (approved):

PR #1496, opened 2026-05-20
fix(workspaces): clear stale `.visible` class on cross-monitor focus

orestisfl:
Compositors only emit a `Focus` event for the (old, new) workspace pair, so a third workspace which was visible on the newly-focused output would keep its `.visible` CSS class indefinitely.

No known issue. 

See video with before/after:
https://github.com/user-attachments/assets/f5aa39df-9422-401a-a119-f930ffddc98d

My ironbar style.css differentiates button colors between visible, focused and hidden. You can see the colors change as I focus the other monitor (not visible in recording).

Before: As I switch to the second monitor, the buttons of the first switch to the visible state color. However, when I switch back to the main monitor to a different workspace, they never lose their "visible" status.
After: Switching to a different workspace in the first monitor updates the other buttons.

### DIFF
--- a/src/modules/workspaces/button.rs
+++ b/src/modules/workspaces/button.rs
@@ -11,6 +11,8 @@ use tokio::sync::mpsc;
 pub struct Button {
     button: IconButton,
     workspace_id: i64,
+    monitor: String,
+    open_state: OpenState,
     conn_id: Option<SignalHandlerId>,
     tx: mpsc::Sender<i64>,
 }
@@ -20,6 +22,7 @@ impl Button {
         id: i64,
         index: i64,
         name: &str,
+        monitor: &str,
         open_state: OpenState,
         context: &WorkspaceItemContext,
     ) -> Self {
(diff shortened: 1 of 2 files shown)


JakeStanger:
Thanks for this. I'm just trying to wrap my head around what's going on - not sure I'm fully following. Is this a bug specific to `all_monitors = true`?

Can you share a basic config and repro steps please?

orestisfl:
> Thanks for this. I'm just trying to wrap my head around what's going on - not sure I'm fully following. Is this a bug specific to `all_monitors = true`?
> 
> Can you share a basic config and repro steps please?

I can actually reproduce this with the default config

Requirements:
1. 2 or more monitors
2. 2 or more workspaces in one of the monitors

Monitor A = monitor with 2+ workspaces. In this example: workspaces `1` and `3`.
Monitor B = other monitor with arbitrary amount of workspaces

Repro:
1. Switch to workspace `1`
2. Switch to monitor B
3. Switch to workspace `3`

Actual result (notice `3` is correctly styled as focused but `1` is incorrectly styled as visible):
image

Expected result (`1` styled as hidden):
image

orestisfl:
Pushed 096ed5c26a439c9bf380db84a69f1d00b2143a89 which does not change the behavior, just my subjective preference code-wise. LMK if you prefer the previous one and I'll revert.

JakeStanger (approved):
Gotcha, thanks. Code lgtm. Will give a quick test when I can & merge if all good.

PR #1501, opened 2026-05-25
fix: volume module fixes

davidaf3:
I've found two issues with the volume module:

1. When the default source in PulseAudio is a monitor and `show_monitors` is False, an empty label is added to the bar. If I set another source that is not a monitor as the default and then switch back to the monitor, the previous source still appears in the bar as the default, but the actual default is the monitor.
2. When the selected sink or source in the popup dropdown changes because of a PulseAudio event, the dropdown  triggers the selected signal and the handler sets that sink or source as the default. But this should only happen when I manually select a sink or source. With PipeWire, this messes up the priority of the sinks and sources.

To fix the first issue, I changed the widget code so that the sink and source labels are only added when there is a default sink or source that should be shown, and are removed otherwise.

To fix the second one, I updated the `select_notify` handler to ignore signals triggered within the popup event handler, where the dropdowns are updated programmatically.

### DIFF
--- a/src/modules/volume/mod.rs
+++ b/src/modules/volume/mod.rs
@@ -1,7 +1,8 @@
 mod config;
 
-use std::cell::RefCell;
+use std::cell::{Cell, RefCell};
 use std::collections::HashMap;
+use std::rc::Rc;
 
 use glib::subclass::prelude::*;
 use glib::{Object, Properties};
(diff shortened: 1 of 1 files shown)


JakeStanger:
Thanks for this. With any luck this should finally crack #1411 too.

JakeStanger (approved):
Code lgtm, I'll give this a quick test when I can & merge if all good.

Issue #1428, opened 2026-03-22
Tray module doesn't detect already dispatched items


Ynverxe:
**Describe the bug**
If there are any items that were dispatched before Ironbar opens, it doesn't detect them the first run, but if you run Ironbar a second time then they show in the tray.

I saw this behaviour with blueman-applet.

**To reproduce**

1. Add module `tray`
2. Make sure there's an applet dispatched before running Ironbar (like blueman-applet or spotify)
3. Run Ironbar
4. Close Ironbar and re-run it.

Edit: Most of Rolling Release stuff, included Ironbar, was installed via Nix

**Expected behavior**
I think the expected behaviour must be that Ironbar should look for already dispatched items.

**System information:**
- Distro: ``Linux war-machine 6.12.74+deb13+1-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.12.74-2 (2026-03-08) x86_64 GNU/Linux``
- Compositor: Hyprland
- Ironbar version: 0.19.0-pre

**Configuration**

<details><summary>Config</summary>

```
height: 30
position: "top"
margin:
  left: 20
  top: 5
  right: 20
start:
- type: "sys_info"
  format:
  - "{memory_used#M} MB / {memory_total#M} MB"
  intervals:
    memory: 1
- type: "label"
  label: "IPv4: {{ ip route get 1 | awk '{print $7}' }}"
end:
- type: "battery"
- type: "volume"
  sink_slider_orientation: "horizontal"
  marquee:
    enable: true
    max_length: 40
- type: "notifications"
center:
- type: "tray"
- type: "workspaces"
  name_map:
    1: "1"
    2: "2"
    3: "3"
    4: "4"
    5: "5"
    6: "6"
    7: "7"
    8: "8"
    9: "9"
  favorites: [ "1", "2", "3", "4", "5", "6", "7", "8", "9" ]
- type: "clock"
  format: "%A %d de %B, %r"
```

</details>

<details><summary>Styles</summary>

```css
@define-color text #3D3D3D;
@define-color workspaces #383838;
@define-color workspaces-inner-background #2a2a2a;
@define-color workspace-active #7ec8e3;
@define-color workspace-hover #7ec8e3;

* {
  border: none;
  box-shadow: none;
  font-family: JetBrains Mono NFM;
  font-weight: 900;
}

.background,
#bar {
  background: transparent;
}

.tray,
.label,
.sysinfo,
.volume,
.networkmanager,
.clock,
[...]

PR #1506, opened 2026-05-31
fix(tray): serialize icon loads per item to preserve dispatch order

slowsage:
This PR enforces dispatch order on tray icon updates (per tray item) while keeping dispatch async across different tray items.
Additionally, it makes `picture` visible explicitly in case it was preceded by a failed load.

*May* be related to #1428 but not sure.

### DIFF
--- a/src/modules/tray/mod.rs
+++ b/src/modules/tray/mod.rs
@@ -6,7 +6,7 @@ use crate::clients::tray;
 use crate::config::{CommonConfig, ModuleOrientation, default};
 use crate::image;
 use crate::modules::{Module, ModuleInfo, ModuleParts, ModuleUpdateEvent, WidgetContext};
-use crate::{lock, module_impl, spawn};
+use crate::{await_sync, lock, module_impl, spawn};
 use color_eyre::{Report, Result};
 use gtk::prelude::*;
 use gtk::{ContentFit, Orientation, Picture};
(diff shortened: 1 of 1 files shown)


JakeStanger:
What's necessitated this? I feel like if events are happening out of order there's probably a more important underlying issue to address, rather than forcing sync loading. That could potentially have performance implications too.

I'm not sure I'm fully understanding though

slowsage:
I use [onedrive_tray](https://github.com/DanielBorgesOliveira/onedrive_tray). I created a service file to make it launch after `graphical-session.target` on boot (hyprland). I noticed onedrive_tray launches but is rendered as text - the app (not maintained) has a bug where it renders an empty pixmap followed by a valid one. But the tray of ironbar does not update with the valid one.

Don't think there is a deeper underlying issue to fix. The tray should render what its given in the order its given IMO. Consecutive `spawn_future_local` have no ordering guarantees. I considered aborting the earlier load when two loads are called in sequence for the same tray item. But thats an issue in the caller IMO.

This fix basically makes each load for a given tray item `await` the previous load on the glib context thread (per tray item). Different tray items still load concurrently.  It should proceed even if `Err` is returned. I verified it fixes the issue.

Please lmk if you had more questions.

JakeStanger:
Right okay, gotcha. I reckon there's possibly a couple of cleaner routes for the fix though, to try and avoid adding new fields/logic specifically for it:

- Use `await_sync` to block on image loading using Tokio. https://github.com/JakeStanger/ironbar/blob/e6326d690cff4d3feaf3689d41ec188b485a547f/src/main.rs#L544-L553
- Modify the image provider struct to provide a sync method.

The image provider is only async because of HTTP support. The only scenario that would come up is if somebody were to add an image override pointing to a remote location; the rest of the time it's only sync operations happening. I'm somewhat tempted to (soft-)deprecate the HTTP feature anyway.

slowsage:
Worked. Please check and lmk. Thank you.
FWIW - the `resolve_location()` for icon files on disk is also async but this approach is less complex.

JakeStanger (approved):
Thanks, LGTM

PR #1508, opened 2026-06-02
fix(workspaces): apply visibility hint classes on startup

orestisfl:
The early-return guard in `set_open_state` short-circuited at button construction time, making the call a no-op.

### DIFF
--- a/src/modules/brightness.rs
+++ b/src/modules/brightness.rs
@@ -334,7 +334,7 @@ impl Module<Button> for BrightnessModule {
                                     default_resource_name.as_deref(),
                                     new_brightness,
                                 )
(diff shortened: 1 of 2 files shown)


orestisfl:
Failing CI check seems unrelated but I pushed a fix:

```
Diff in /home/runner/work/ironbar/ironbar/src/modules/brightness.rs:334:
                                     default_resource_name.as_deref(),
                                     new_brightness,
                                 )
-                                    .await
+                                .await
                                 {
                                     tracing::error!(?err, "Could not change brightness");
                                 }
```

JakeStanger (approved):
LGTM, thanks

PR #1511, opened 2026-06-02
ci: add nix-build job to replace garnix

JakeStanger:
Move job into GH Actions since Garnix is retiring 

This also moves back to Cachix

TODO:

- [x] Update documentation to point to Cachix


JakeStanger:
Is it compiling by hand

JakeStanger:
If anybody who knows what they're doing has any clues as to how we can make the build not take 15 minutes that'd be very much appreciated

JakeStanger:
Warm cache = ~45 seconds, nvm

PR #1513, opened 2026-06-04
fix(volume): refactor popup

davidaf3:
I have found some issues with the volume popup, like the mute button and volume slider appearing when there is no sink or source, the source selector not going back to "No source available" after removing my mic, and the same profiles manager being shared between devices and streams.

After fixing these problems, there was a lot of duplicate code, so I refactored the shared logic into new `DeviceUi` and `StreamUi` structs, and moved the type-specific code into `SinkUi`, `SourceUi`, `SinkInputUi`, and `SourceOutputUi`.

I have tried to keep the existing logic, so aside from these fixes, the behavior should be the same as before.

### DIFF
--- a/src/modules/volume/mod.rs
+++ b/src/modules/volume/mod.rs
@@ -11,11 +11,11 @@ use gtk::{
     Button, DropDown, Expression, Label, ListItem, Orientation, Scale, SignalListItemFactory,
     ToggleButton, gio,
 };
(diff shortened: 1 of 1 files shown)


JakeStanger (approved):
Thanks, appreciate the fixes, and this definitely looks tidier than before. LGTM, no notes.

PR #1517, opened 2026-06-10
fix[style]: Use dark backgrounds for more UI

tmccombs:
This fixes a couple of places, where the UI was falling back to the default GTK styles, which was unreadable with a light gtk theme combined with the dark colors of the "minimal" style.

In particular:

-  Use primary background for "popover .view" so that the background of dropdown menus (for example the device dropdown in the volume module)  uses a dark background, and the light colord text is visible.
- Use secondary dark color for "*:selected"  so that the "selected" item in the dropdown uses a dark background, and is more readable. Also impacts action items in tray drop downs.

### DIFF
--- a/examples/minimal/style.css
+++ b/examples/minimal/style.css
@@ -44,7 +44,7 @@ switch:not(:checked) {
   background-color: var(--color-dark-secondary);
 }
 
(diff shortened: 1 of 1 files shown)


JakeStanger (approved):
Thanks!

PR #1535, opened 2026-06-22
chore(autohide): cleanup setup_autohide

AlexNomadrg:
Cleaned up `setup_autohide` by adding some helper functions and moving `timeout_id` to `AutohideState`

Issue #1534, opened 2026-06-22
(autohide) Reveal bar on workspace switch


AlexNomadrg:
**Is your feature request related to a problem? Please describe.**
I would like the bar to be briefly shown when a workspace switch occurs

**Describe the solution you'd like**
There should be a config option that would allow this behavior.
Something like `autohide_show_on_switch <ms>`

**Describe alternatives you've considered**
Not sure if any alternative solutions are possible

JakeStanger:
Currently bars can be shown/hidden over IPC, and that works with the autohide feature. It should be possible currently to write an external script that listens to your compositor's IPC events (or even `ext-workspace` events), and send the request to Ironbar.

Long-term the plan is to have a more integrated Lua API which will allow you to interact with the Ironbar clients directly from in-process.

AlexNomadrg:
> Currently bars can be shown/hidden over IPC, and that works with the autohide feature. It should be possible currently to write an external script that listens to your compositor's IPC events (or even `ext-workspace` events), and send the request to Ironbar.
> 
> Long-term the plan is to have a more integrated Lua API which will allow you to interact with the Ironbar clients directly from in-process.

Just to clarify, does that mean you wouldn't want this implemented as a built-in Ironbar feature?

JakeStanger:
It doesn't necessarily mean that, but it's something I think needs more time and thought to design something that scales nicely and works across the board.

PR #1530, opened 2026-06-21
feat(autohide): configurable hotspot height

AlexNomadrg:
Adds `autohide_hotspot_height` property to allow the configuration the height of the hotspot

Issue #1243, opened 2025-11-20
FR: Autohide additional triggers


balgerion:
First off, love Ironbar, it’s awesome!  

It would be amazing if `autohide` could have more than one trigger — not just moving the mouse to the screen edge, but also moving to the edge and using scroll (or any other button/event) to show the bar.  

Reason: I work on an ultrawide monitor and love having the freedom to move my mouse anywhere without accidentally triggering the bars. :)

JakeStanger:
This sounds simple enough to me. The autohide works using a tiny hidden window at the screen edge, so no reason event listeners other than mouse enter can't be hooked up to it.

PR #1536, opened 2026-06-22
feat(autohide): show on workspace change

AlexNomadrg:
Depends on #1535
Implements #1534

Adds `autohide_show_on_workspace_change` option to the config.

I think it's a good idea to add a separate config module for autohide if you merge #1530 too and/or #1243 gets implemented because the these names are getting messy :face_with_diagonal_mouth:


JakeStanger:
Thanks but I'm going to close this as I don't like the approach here.

I'm not sure I'd want this as a config option (especially not at the bar level) - I'd potentially consider this as an option on the workspace module, although I'm not sure how that'd be implemented. Ideally I'd rather see this as a scriptable/event-driven approach.

PR #1537, opened 2026-06-23
feat(autohide): reveal on scroll option

### LINKED ISSUE #1243: FR: Autohide additional triggers

AlexNomadrg:
Resolves #1243

Turned out to be kind of a nothingburger. Tried to add a keyboard listener but then realized that the hotspot window never gets focus so it's not possible.

Only adds a `scroll` listener.


src/config/mod.rs
```diff
@@ -291,6 +291,15 @@ pub enum BarPosition {
     Right,
 }
 
+#[derive(Debug, Default, Deserialize, Copy, Clone, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+#[cfg_attr(feature = "extras", derive(JsonSchema))]
+pub enum AutohideListner {
```
JakeStanger: Typo:

```suggestion
pub enum AutohideListener {
```

src/config/mod.rs
```diff
@@ -409,6 +418,11 @@ pub struct BarConfig {
     /// **Default**: `null`
     pub autohide: Option<u64>,
 
+    /// Type of listener for revealing the hidden bar
+    ///
+    /// **Default**: `motion`
+    pub autohide_listener: AutohideListner,
```
JakeStanger: Property also needs adding to the docs please, along with possible options.

src/config/mod.rs
```diff
@@ -291,6 +291,15 @@ pub enum BarPosition {
     Right,
 }
 
+#[derive(Debug, Default, Deserialize, Copy, Clone, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+#[cfg_attr(feature = "extras", derive(JsonSchema))]
+pub enum AutohideListner {
+    #[default]
+    Motion,
```
JakeStanger: I think this would be better named `Hover`, as that's more descriptive to the end-user than an internal GTK event.

JakeStanger (changes requested):
Thanks. A couple of changes before this is ready please.

Since I won't be looking to merge this until the 0.20.0 window, if you want to have a go at adding a click listener too I reckon we can regard the issue complete.

AlexNomadrg:
Thanks for the suggestion, added a `click` listener too

balgerion:
Thank you ! I was rly rly waiting for that

JakeStanger (approved):

PR #1554, opened 2026-07-07
🐛 fix(hyprland): switch workspaces on Lua config providers

### LINKED ISSUE #1548: Workspace clicks fail on Hyprland 0.55+ Lua config (dispatch sent in legacy syntax)
**Describe the bug**

Clicking a workspace button in the `workspaces` module does nothing when Hyprland (0.55+) is running the new **Lua config** (`configProvider: lua`). The button reacts visually but focus never changes. ironbar dispatches workspace focus in the legacy IPC syntax, which Hyprland's Lua config no longer accepts.

The workspace-switch log line:
```
ERROR ironbar::clients::compositor::hyprland: Couldn't focus workspace '5':
A dispatcher returned a non-`ok`, value which is probably an error:
error: [string "return hl.dispatch(workspace 5)"]:1: ')' expected near '5'
 → Note: dispatch in lua is a shorthand for hl.dispatch(...), your syntax might need to be updated.
```

**To reproduce**

1. Run Hyprland 0.55+ with a Lua config (`~/.config/hypr/hyprland.lua`)
2. Add module `workspaces`
3. Click a workspace on the bar
4. Observe that focus does not change; the error above is logged

**Expected behavior**

Clicking a workspace focuses it, as it does under a legacy `hyprland.conf` config.

**System information:**
- Distro: Arch Linux
- Compositor: Hyprland 0.55.4 (Lua config, `configProvider: lua`)
- Ironbar version: 0.19.0

**Configuration**

```json
{
  "position": "top",
  "start": [
    { "type": "workspaces", "sort": "index", "all_monitors": false }
  ]
}
```

</details>

**Additional context**

**Root cause.** `WorkspaceClient::focus()` in `src/clients/compositor/hyprland.rs` dispatches via `hyprland-rs`:

```rust
Dispatch::call(DispatchType::Workspace(WorkspaceIdentifierWithSpecial::Id(id as i32)))
```

`hyprland-rs` serializes this to the legacy IPC command `dispatch workspace 5`. Under a Lua config, Hyprland wraps any dispatch payload as `return hl.dispatch(<payload>)` and evaluates it as Lua, so `workspace 5` becomes the invalid expression `hl.dispatch(workspace 5)`. The binding-level issue is hyprland-community/hyprland-rs#391, but that crate is effectively unmaintained per its maintainer.

Only **dispatch** is affected — event subscription (socket2) and JSON queries still work, so the blast radius is a single call site (`focus()`).

**What works.** Under a Lua config, the equivalent Lua dispatch succeeds:

```
$ hyprctl dispatch 'hl.dsp.focus({ workspace = "5" })'
ok
```

And Hyprland exposes the active provider, so it's detectable at runtime:

```
$ hyprctl systeminfo | grep configProvider
configProvider: lua
```

**Proposed fix (non-breaking, minimal).**

Since there's one dispatch site and the config type is detectable, this can be fixed without dropping `hyprland-rs`:

1. Detect the provider once at client init (send `systeminfo` over the ctl socket, check for `configProvider: lua`, cache it).
2. Branch `focus()`:
   - **Lua config** → send `dispatch hl.dsp.focus({ workspace = "<id>" })` over the ctl socket and verify the reply is `ok`.
   - **legacy config** → keep the existing `Dispatch::call(..)` path unchanged.

Legacy users are unaffected. One detail: special workspaces (negative ids) should be routed by name rather than `focus({ workspace = "-N" })`.

**Screenshots**

N/A (behavioral — see the log output above).

@JakeStanger would you accept a PR with the approach described above?

dcaixinha:
Clicking a workspace button did nothing on Hyprland 0.55+ when running with a Lua config provider.

`hyprland-rs` serializes the focus dispatch to the socket as the legacy `dispatch workspace <id>`. Under a Lua config provider, Hyprland evaluates every dispatch payload as a Lua expression (`return hl.dispatch(<payload>)`), so `workspace <id>` is invalid Lua and fails with `"')' expected near '<id>'"`, and thus the switch never happens and no visual feedback occurs.

Detect the config provider once at client construction via `hyprctl systeminfo` (the `configProvider: lua` line; hyprland-rs does not expose `systeminfo`), cache it, and on a Lua provider dispatch `hl.dsp.focus {workspace="<id>"}` instead. The legacy path is left untouched, and detection failures fall back to legacy so nothing changes for existing setups. Only `dispatch` payloads are Lua-wrapped, so other commands (e.g. keyboard-layout switching) are unaffected.

Tested this branch on my live system, and clicking now works on Hyprland using a Lua config.

Closes #1548

### DIFF
--- a/src/clients/compositor/hyprland.rs
+++ b/src/clients/compositor/hyprland.rs
@@ -35,6 +35,9 @@ pub struct Client {
     #[cfg(feature = "workspaces+hyprland")]
     workspace: TxRx<WorkspaceUpdate>,
 
+    #[cfg(feature = "workspaces+hyprland")]
+    use_lua_dispatch: bool,
+
     #[cfg(feature = "keyboard+hyprland")]
     keyboard_layout: TxRx<KeyboardLayoutUpdate>,
 
@@ -47,6 +50,8 @@ impl Client {
         let instance = Self {
             #[cfg(feature = "workspaces+hyprland")]
             workspace: TxRx::new(),
+            #[cfg(feature = "workspaces+hyprland")]
+            use_lua_dispatch: detect_lua_config(),
             #[cfg(feature = "keyboard+hyprland")]
             keyboard_layout: TxRx::new(),
             #[cfg(feature = "bindmode+hyprland")]
@@ -406,9 +411,15 @@ impl Client {
 #[cfg(feature = "workspaces+hyprland")]
 impl super::WorkspaceClient for Client {
     fn focus(&self, id: i64) {
-        let identifier = WorkspaceIdentifierWithSpecial::Id(id as i32);
+        let res = if self.use_lua_dispatch {
+            let arg = format!("{{workspace=\"{id}\"}}");
+            Dispatch::call(DispatchType::Custom("hl.dsp.focus", &arg))
+        } else {
+            let identifier = WorkspaceIdentifierWithSpecial::Id(id as i32);
+            Dispatch::call(DispatchType::Workspace(identifier))
+        };
 
-        if let Err(e) = Dispatch::call(DispatchType::Workspace(identifier)) {
+        if let Err(e) = res {
             error!("Couldn't focus workspace '{id}': {e:#}");
         }
     }
@@ -494,6 +505,23 @@ impl KeyboardLayoutClient for Client {
 impl BindModeClient for Client {
     fn subscribe(&self) -> super::Result<Receiver<BindModeUpdate>> {
         Ok(self.bindmode.tx.subscribe())
+    }
+}
+
+#[cfg(feature = "workspaces+hyprland")]
+fn detect_lua_config() -> bool {
+    match std::process::Command::new("hyprctl")
+        .arg("systeminfo")
+        .output()
+    {
+        Ok(output) => String::from_utf8_lossy(&output.stdout).lines().any(|line| {
+            let line = line.trim();
+            line.starts_with("configProvider:") && line.contains("lua")
+        }),
+        Err(err) => {
+            warn!("Failed to detect Hyprland config provider, assuming legacy: {err}");
+            false
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/clients/compositor/hyprland.rs
```diff
@@ -497,6 +508,23 @@ impl BindModeClient for Client {
     }
 }
 
+#[cfg(feature = "workspaces+hyprland")]
+fn detect_lua_config() -> bool {
+    match std::process::Command::new("hyprctl")
+        .arg("systeminfo")
+        .output()
```
JakeStanger: Even as a temporary patch, I do not want to introduce code that spawns an external CLI process. That's way too expensive and fragile.

src/clients/compositor/hyprland.rs
```diff
@@ -497,6 +508,23 @@ impl BindModeClient for Client {
...
+    {
+        Ok(output) => String::from_utf8_lossy(&output.stdout).lines().any(|line| {
+            let line = line.trim();
```
JakeStanger: Does Hyprland's IPC not support JSON output?

JakeStanger (changes requested):

src/clients/compositor/hyprland.rs
dcaixinha: Yeah, I didn't like this approach either, but I couldn't find another suitable approach, since `hyprland-rs` doesn't expose this `configProvider` anywhere. Any ideas on how to do this differently? This only runs _once_ btw.

JakeStanger: I can't merge this as it stands, regardless of how often it runs. I don't think there's much choice but to open the IPC socket and send either a `systeminfo` or `version` command and parse the response. For a single command you can probably hand-roll it in a single function easy enough.

dcaixinha: I just pushed a new commit that follows this approach. It uses the IPC socket and sends a `j/status` message over it to get the `configProvider` from the JSON response.

src/clients/compositor/hyprland.rs
dcaixinha: I wasn't aware of this, but it does indeed. As described on the other thread, I'm now using `j` messages on the IPC socket.

JakeStanger (approved):
lgtm, cheers

PR #1560, opened 2026-07-17
feat: better support for 'cargo deb'

Sudrien:
https://github.com/kornelski/cargo-deb is an alternative to `cargo install` that creates deb packages for Debian's apt.

`cargo deb` works without this patch, sans service file, but complains in a few places once installed. This cleans up those complaints.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,16 @@ keywords = ["gtk", "bar", "wayland", "wlroots", "gtk-layer-shell"]
 [package.metadata.binstall]
 pkg-url = "{ repo }/releases/download/v{ version }/{ name }-v{ version }-{ target-arch }{ archive-suffix }"
 
+[package.metadata.deb]
+maintainer = "Jake Stanger"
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -209,4 +209,14 @@ schemars = { version = "1.2.1", optional = true, features = ["indexmap2"] }
 clap_complete = {  version = "4.6.5", optional = true }
 
 [build-dependencies]
-vergen-git2 = "9.1.0"
\ No newline at end of file
+vergen-git2 = "9.1.0"
+
+[package.metadata.deb]
```
JakeStanger: Can the section be moved up towards the top of the file please, under the `[package.metadata.binstall]` header?

Sudrien: Done. 

It seems the Web UI is not letting me remove that final newline.

JakeStanger: Newline is fine, it comes and goes

JakeStanger (approved):
lgtm, cheers

Issue #1296, opened 2025-12-22
Battery: Power draw in watts


imnotpoz:
**Is your feature request related to a problem? Please describe.**

I'd like to see the power draw in watts when I'm on battery

**Describe the solution you'd like**

Implement power draw in watts in the battery module

**Describe alternatives you've considered**

Using wa*bar (it has that feature)

**Additional context**

https://github.com/Alexays/Waybar/wiki/Module:-Battery
https://github.com/Alexays/Waybar/blob/master/src/modules/battery.cpp (sorry for making you read c++ code)

JakeStanger:
Crate hasn't been updated in 5 years so I don't want to depend on it, but I think this has support for reading power draw and it's Rust so it's probably easier to port: 

https://github.com/svartalf/rust-battery

bill88t:
It's particularly simple diy'ing it. Under `/sys/class/power_supply`, the batteries with current monitoring support, expose the attribute `current_now`.

This value may be negative with some drivers, *cough cough* sbs_battery on ARM systems, but it usually is positive, even for energy loss. The logic cannot be just a blind `// 1000` mA. It has to be evaluated over the presence of attribute `online` of detected power inputs (also under the same sysfs folder), falling back to simple percentage direction monitoring (50% -> 52%, means were charging).
Still, hardly a matter worthy of dependency churn.

I'd do it myself, but I just installed this bloody thing, not even having launched it. It'll have to wait till next year.

Still if you wish to do this, the only last bit of knowledge needed to 100% this is that some HID devices may also show up as batteries and charging sources, which you'll need to exclude from these scans. Carpet-blocking `r"*hid*"` works.
There isn't much to anything else in this needed to implement this.

PR #1568, opened 2026-07-21
feat(battery): implement power draw in watts in the battery module

imnotpoz:
this PR implements the solution asked for and outlined by @imnotpoz in https://github.com/JakeStanger/ironbar/issues/1296

the power draw works but updates pretty slowly (probably dbus (fuck dbus)) but overall it seems fine and usable

also very nice code in this bar of iron I really enjoyed it :-)

### DIFF
--- a/src/modules/battery.rs
+++ b/src/modules/battery.rs
@@ -27,44 +27,63 @@ const MINUTE: i64 = 60;
 #[serde(default)]
 struct ProfileState {
     /// Battery charged percentage.
-    percent: f64,
+    /// Omit to match regardless of battery percentage.
+    percent: Option<f64>,
     /// Whether the battery is currently charging.
     /// Omit to match regardless of charging state.
     charging: Option<bool>,
+    /// Battery power draw in watts.
+    /// Omit to match regardless of battery power draw.
+    power_draw: Option<f64>,
 }
 
 impl Default for ProfileState {
     fn default() -> Self {
         Self {
-            percent: 100.0,
+            percent: None,
             charging: None,
+            power_draw: None,
         }
     }
 }
 
 impl PartialOrd for ProfileState {
     fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
-        if self.percent == other.percent {
-            match (self.charging, other.charging) {
-                (Some(_), Some(_)) | (None, None) => Some(Ordering::Equal),
+        if self.percent != other.percent {
+            return match (self.percent, other.percent) {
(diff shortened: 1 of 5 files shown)


src/modules/battery.rs
JakeStanger, line 73: I have no idea what or how because I couldn't come up with anything better than the existing mess, but I'd love to see a solution for this that does away with the nesting.

JakeStanger (commented):
Code lgtm, although one potential area for improvement if you can think of a way (otherwise don't worry).

Regarding the slowness, is it slow enough to be a concern? There is the potential to read directly from file as in https://github.com/JakeStanger/ironbar/issues/1296#issuecomment-370267465.

Will test at some point in the future and approve if all good.

src/modules/battery.rs
```diff
@@ -31,24 +31,31 @@ struct ProfileState {
     /// Whether the battery is currently charging.
     /// Omit to match regardless of charging state.
     charging: Option<bool>,
+    /// Battery power draw in watts.
+    power_draw: f64,
```
JakeStanger: Can you ensure this is documented on the wiki please?

imnotpoz: repushed, done

JakeStanger: It needs adding [here](https://github.com/imnotpoz/ironbar/blob/de4fc87bdf1c9ecc67878b7c96462a63cdbda87d/docs/modules/Battery.md?plain=1#L23-L26) too please for the profiles system

imnotpoz: okay sorrgy :pleading_face:

src/modules/battery.rs
imnotpoz:
```rust
impl PartialOrd for ProfileState {
    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
        if self.percent != other.percent {
            return self.percent.partial_cmp(&other.percent);
        }
        if self.power_draw != other.power_draw {
            return self.power_draw.partial_cmp(&other.power_draw);
        }

        match (self.charging, other.charging) {
            (Some(_), Some(_)) | (None, None) => Some(Ordering::Equal),
            (None, Some(_)) => Some(Ordering::Greater),
            (Some(_), None) => Some(Ordering::Less),
        }
    }
}
```

couldn't come up with anything better so far but there's at least no nesting

I don't think it's in any way possible to chain `partial_cmp` calls, like with `PartialEq` implementations

imnotpoz:
repushed with all 3 fields being optional

making this into a draft until the profiles refactor

PR #1581, opened 2026-08-05
build(nix): replace nix-systems with `lib.systems.flakeExposed`

imnotpoz:
this removes an unneeded input and makes sure anyone who can use flakes has access to this one's outputs

another alternative is `lib.platforms.all` (don't ask my why one is under systems and the other platforms) which is all the systems nix has (but probably unnecessary and clutters `nix flake show` and similar)

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -30,21 +30,6 @@
         "type": "github"
       }
     },
-    "nix-systems": {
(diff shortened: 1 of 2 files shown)


JakeStanger (approved):
Wonderful thanks

Issue #1538, opened 2026-06-26
Some `Tray` functionality missing for (e.g.) `radiotray-ng`


boogiewoogit:
https://github.com/ebruck/radiotray-ng seems to be somewhat of a litmus test for systrays: it makes some systray functionality details visible. This is also true for ironbar.

Concrete current issues using `radiotray-ng` in ironbar's `Tray` module:
+ Scroll inputs (mousewheel up/down, to adjust volume in the case of radiotray-ng) are not supported.
+ Tray's 'menu' command opens the menu -- as expected. But it only opens it **downwards**. This means that the menu is unaccessible if the bar's position is at the bottom of the screen.
+ The menu shown doesn't update its contents, e.g. if you change station, the menu won't reflect that.

JakeStanger:
~~Is the scroll event issue in regard to Ironbar's cross-module `on_scroll_{up,down}` handlers? It's not clear what you're expecting to work here, or where.~~ I've done some digging, and it looks like `radiotray-ng` uses `libappindicator`, which is the old GTK2/3 way of doing things from back in the Unity days (!). It works by providing a space for the host application to render its own tray item, so the scroll events are handled internally. Ironbar will never support this.

Also have you observed the menu opening the wrong way? GTK handles the directly automatically based on the screen edge & all four orientations should work as expected automatically (and do for me).

I've found the cause of the menu update bug - this is the first tray app I've seen that actually sends menu property update events rather than replacing the whole thing, and I just skipped implementing it previously as I had nothing to test against. I'll look to get that implemented.

boogiewoogit:
> I've done some digging, and it looks like radiotray-ng uses libappindicator, which is the old GTK2/3 way of doing things from back in the Unity days (!). It works by providing a space for the host application to render its own tray item, so the scroll events are handled internally. Ironbar will never support this.

Ah, gotcha. Interesting. About `on_scroll_{up,down}`, they just weren't included in the [wiki page](https://github.com/JakeStanger/ironbar/wiki/tray#configuration), which I just thought to be exhaustive. For the record, they absolutely do work with the explicit radiotray-ng commands, i.e.
```
# toml
type = "tray"
on_scroll_up = "qdbus com.github.radiotray_ng /com/github/radiotray_ng com.github.radiotray_ng.volume_up"
on_scroll_down = "qdbus com.github.radiotray_ng /com/github/radiotray_ng com.github.radiotray_ng.volume_down"
```

[...]

PR #1583, opened 2026-08-09
fix(tray): apply menu property updates and activate the correct radio item

tomasz-feliksik:
Two related tray menu fixes which together make menus with radio groups work
end-to-end. The first implements the menu update bug you diagnosed in #1538;
the second fixes a wrong-id activation bug in the radio path that the first
fix made observable.

## 1. Apply `MenuDiff` updates so menus repaint after a click

The tray interface handled `UpdateEvent::Menu` (full replace) and
`MenuConnect`, but only `trace!`'d `UpdateEvent::MenuDiff` and dropped it.
`MenuDiff` is how the standard `com.canonical.dbusmenu`
`ItemsPropertiesUpdated` signal reaches the module, so every post-click
property change — a moved radio toggle-state, a refreshed label, an
enabled/disabled entry — was discarded and the widget kept rendering the
previous state. This is the case you described in #1538:

> this is the first tray app I've seen that actually sends menu property
> update events rather than replacing the whole thing, and I just skipped
> implementing it previously as I had nothing to test against.

It affects any conforming tray implementation, not one toolkit: Qt,
libdbusmenu and ksni all report property-only changes this way and only
escalate to `LayoutUpdated` when the structure changes.

**Why apply the diffs in ironbar rather than in `system-tray`:** the crate's
own `apply_menu_diffs` walks only the top-level submenus and consumes diffs
positionally (`next_if(|d| d.id == item.id)`), so nested groups are never
patched — but fixing that alone wouldn't be enough, because ironbar dropped
the event before the widget could repaint. The widget side needs a retained
menu model and a rebuild regardless, so `TrayMenu` now keeps the last full
menu model and applies diffs itself, matching by global item id at any depth
(`src/modules/tray/diff.rs`, unit-tested including the nested-radio case).
Happy to send a recursion fix for the crate's `apply_menu_diffs` as a
follow-up if you'd take it.

## 2. Activate the clicked radio item, not the group's first

Clicking any radio item sent one fixed dbusmenu id, so the application either
did nothing or silently re-applied a setting it already had. Two defects
combined:

1. **Colliding action names.** `as_menu` numbered radio groups with a counter
   local to the function, but recursed into each submenu carrying the same
   `SimpleActionGroup`. The counter restarted per submenu while the namespace
   did not, so every submenu's first group was named `action_radio_1` —
   and `ActionMap::add_action` replaces on a duplicate name, leaving one
   surviving action driven by every radio item in the menu.
2. **The change-state handler discarded its parameter.** One action backs a
   whole group, so the clicked option is identified only by the target string
   in the new state. The handler was `move |_, _| activate(.., id)`, using an
   id captured at build time — the group's first item. On its own this breaks
   every group of more than one option, even without collisions.

The fix names the action from the group's first dbusmenu id (unique
menu-wide, so collision-free by construction) and reads the clicked id back
out of the change-state payload. The group's state is also parked on the item
the application reports as selected, so the radio mark renders on the actually
active option instead of always the first.

Standard and checkmark items were never affected — they use the globally
unique `action_{id}` path — which is why tray menus looked mostly functional
and only radio groups misbehaved.

## Testing

- Unit tests for the diff application (nested radio toggle-state move, label
  update, unknown-id no-op) and for the radio action contract (action-name
  uniqueness across groups, target string round-tripping the item id).
- Verified live against a ksni app whose menu has five radio groups in five
  submenus. Before: every click in every group sent id 19 (confirmed via
  `busctl … GetLayout` against the app's real menu ids plus trace logging of
  the outgoing `ActivateRequest`). After: the sent id varies per click,
  selections apply across different groups and both options within a group,
  and the menu repaints to show the moved selection without reopening.
- `cargo fmt --check`, `cargo clippy --all-targets` (0 warnings), builds with
  `--no-default-features`, `cargo test` all pass.

Refs #1538 — this implements the menu update part; the scroll-event and
libappindicator points from that issue are out of scope here.

### DIFF
--- a/src/modules/tray/diff.rs
+++ b/src/modules/tray/diff.rs
@@ -1,0 +1,153 @@
+//! Incremental application of dbusmenu property updates (`MenuDiff`) to a
+//! retained menu model.
+//!
+//! The `system-tray` crate delivers post-interaction changes — a radio
+//! selection moving to another entry, a submenu label changing — as
+//! [`MenuDiff`]s keyed by the *global* dbusmenu item id, which can sit at any
+//! depth in the menu tree. The crate's own `apply_menu_diffs` only walks the
+//! top-level items, so nested items (e.g. a radio group inside a submenu) never
+//! receive their updates. We therefore walk the whole tree and patch the item
+//! whose id matches, recursing into every submenu.
+
+use system_tray::menu::{MenuDiff, MenuItem, MenuItemUpdate};
+
+/// Applies every diff to the matching item anywhere in `items`, recursing into
+/// nested submenus. A diff whose id matches no item is ignored.
+///
+/// The `remove` field of a diff (properties reset to their default) is not
+/// applied — this matches the property set the `system-tray` crate itself
+/// understands, and the reported failures are all `update`s (toggle-state,
+/// label).
+pub(crate) fn apply_menu_diffs(items: &mut [MenuItem], diffs: &[MenuDiff]) {
+    for item in items {
+        if let Some(diff) = diffs.iter().find(|d| d.id == item.id) {
+            apply_menu_item_update(item, &diff.update);
+        }
+        apply_menu_diffs(&mut item.submenu, diffs);
+    }
+}
+
+/// Overwrites exactly the fields the update carries, leaving the rest intact.
+fn apply_menu_item_update(item: &mut MenuItem, update: &MenuItemUpdate) {
+    if let Some(label) = &update.label {
+        item.label.clone_from(label);
+    }
+    if let Some(enabled) = update.enabled {
+        item.enabled = enabled;
+    }
+    if let Some(visible) = update.visible {
+        item.visible = visible;
+    }
+    if let Some(icon_name) = &update.icon_name {
+        item.icon_name.clone_from(icon_name);
+    }
+    if let Some(icon_data) = &update.icon_data {
+        item.icon_data.clone_from(icon_data);
+    }
+    if let Some(toggle_state) = update.toggle_state {
+        item.toggle_state = toggle_state;
+    }
+    if let Some(disposition) = update.disposition {
+        item.disposition = disposition;
+    }
+}
+
+#[cfg(test)]
+mod tests {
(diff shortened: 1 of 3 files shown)


tomasz-feliksik:
PR description written using AI, but I've analysed and tested the code myself.
@JakeStanger - this is my first contribution to Ironbar, and since I'm using it every day probably not the last one. Let me know if anything can be improved ;)

tomasz-feliksik:
Pushed one more commit: `fix(tray): detach the previous shortcut controller on menu rebuild`.

While soak-testing this branch as my daily bar I found `build_menu_widget` leaked a `ShortcutController` per menu rebuild — `add_controller` appends, and nothing ever removed the previous controller (`insert_action_group`, by contrast, replaces by name). The leak predates this PR, but the MenuDiff handling here makes rebuilds more frequent, so it seemed wrong to ship the one without the other.

Measured with a scripted SNI app emitting 6000 toggle-state property updates (each one a rebuild): pre-fix RSS grew ~11 MB and accelerating; with the fix it stays flat (+0.17 MB total).

JakeStanger:
The repo's AI policy clearly disallows vibe coding, and I would say this definitely qualifies as that. Since the damage is already done, I'm not going to close it and I'll look to merge this. Just don't expect a quick review because I've got to unpick a lot of fluff, and don't expect to credited.

tomasz-feliksik:
Fair - and to be fully straight: yes, the code was AI-assisted.

For what it's worth: I'd already fixed this for my own use - upstreaming it was meant as a thank-you for Ironbar, not a demand on your time. I tried to put more effort into the PR than the review would cost; I can see the extra prose achieved the opposite, and that's mine to fix. The diagnosis and testing were mine and I can answer for every line, so one ask: judge the code as code, at whatever pace suits you, or not at all - no obligation.

If it helps, I'll strip the comments and tests down to house style and split the leak fix into its own PR - say the word. No expectations on credit either.

JakeStanger (approved):

JakeStanger:
Cool, that works

PR #1590, opened 2026-08-12
Added mangowm compositor handling and workspaces supporting and Updated.

8dmusichannels-star:
As I said, I added compositor handling and workspace support with a simple implementation for Mangowm, but I did it in mmsg-based format. An entry point is required. Since there will be the possibility of breaking changes in future versions regarding dwl-ipc.h stabilization, I supported it in mmsg-based format because mmsg can remain more stable in this regard.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1933,7 +1933,7 @@ dependencies = [
 
 [[package]]
 name = "ironbar"
-version = "0.19.0"
+version = "0.20.0"
 dependencies = [
  "bluer",
  "cfg-if",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ironbar"
-version = "0.19.0"
+version = "0.20.0"
 edition = "2024"
 license = "MIT"
 description = "Customisable GTK Layer Shell wlroots/sway bar"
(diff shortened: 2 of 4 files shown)


JakeStanger:
I'll leave this open for visibility for now but this won't be merged, as I'll be merging `ext-workspace` support next update which MangoWM supports.

8dmusichannels-star:
Okay, I understand, that's why I already added this Mangowm workspaces feature, but since focusing was not working, I didn't add it without asking your opinion about it before adding the protocol support that this feature supports, but if you want, I can help you add the Wayland ext-workspace protocol that Mangowm supports.

JakeStanger:
There's a PR with working support in #1465. It needs some attention and refactoring before merging in.

You can have a look at that or give it a test run and report back with any problems if you want to help with that.

8dmusichannels-star:
I tested it, I merged my patch with the patch you mentioned and built it, and it really works. It directly tells you the workspace index number and shows which window number is focused, but the striped color focusing only works while you're touching it, because the striped color doesn't show which window is focused, it only shows which window is in your workspace, that's all. But it works. Repositories used:
https://github.com/xMAC94x/ironbar
https://github.com/8dmusichannels-star/ironbar
Merge repository containing both:
https://github.com/8dmusichannels-star/ironbar-ext-workspaces-mangowm
I only opened this repository for testing and building purposes; otherwise, you need to know that Cargo.toml and binary.yml are modified before directly merging.

8dmusichannels-star:
https://github.com/user-attachments/assets/af57f0b9-732a-4cfa-a279-9be63d4d431f

8dmusichannels-star:
> There's a PR with working support in #1465. It needs some attention and refactoring before merging in.
> 
> You can have a look at that or give it a test run and report back with any problems if you want to help with that.

Dude, are you interested or have I combined the updates myself?

JakeStanger:
I'm not sure I understand your comments, but I think you're asking if I want a patch to add the focused workspace support on top of `ext-workspaces`? 

We'll definitely want that as a stopgap (see https://github.com/JakeStanger/ironbar/pull/1465#issuecomment-4303159423), but I don't know what the architecture of that will look like yet as I'll need to do the same for the other 3 compositors too.

8dmusichannels-star:
What I mean in the comments is that focusing is problematic, sometimes the windows I focus on do not even appear in the bar, so this implementation is not very sufficient in terms of functionality and has many shortcomings, but despite this, it can be improved.

8dmusichannels-star:
Architecturally, I recommend you to use Dwl as a basis. In this way, you will be able to support many Dwl-based composers, but since they are all different from each other and providing common support is difficult and not possible.

8dmusichannels-star:
Therefore, they can all be supported under separate implementations.

PR #1591, opened 2026-08-13
fix(sway): avoid workspace focus race on rename

Rijul-A:
Client::focus() looked up the workspace's current name and sent `workspace <name>`. Between that lookup and the command actually running, another process reacting to the resulting focus change (e.g. a workspace renamer) could rename the workspace, leaving the command targeting a name that no longer existed, which made sway create a new, empty workspace instead of focusing the existing one.

Use `workspace number <num>` when the workspace has one, since sway's workspace number is stable across renames and unaffected by this race. Falls back to the previous name-based command for workspaces with no number (sway reports num as -1 for those).

### DIFF
--- a/src/clients/compositor/sway.rs
+++ b/src/clients/compositor/sway.rs
@@ -16,18 +16,31 @@ impl super::WorkspaceClient for Client {
         spawn(async move {
             let mut client = client.lock().await;
 
-            let name = client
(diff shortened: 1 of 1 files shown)


JakeStanger (approved):
LGTM, thanks!

PR #1597, opened 2026-09-03
Add keyboard layout support for niri

avasich:
### Refactor `niri::Client` and implement `KeyboardLayoutClient` for it

Shared workspace and keyboard layout state now uses `Mutex`. Previously, during single transition, individual reads and writes where performed under separate lock guards. It made state transitions non-atomic; theoretically, another task could observe an intermediate state or modify it, invalidating indices into `workspaces` vec.

In practice, it was unlikely to happen, because the only reason the state is shared is that `WorkspaceClient::subscribe` and `KeyboardLayoutClient::subscribe` need it during startup. After that, the state is only accessed by the task spawned in `Client::new`, making `RwLock` overhead unnecessary.

Event handlers are organized as a chain, each handler either consumes the event or passes it to the next one.

Removed the `first_event` special case: workspace ordering is already handled by `WorkspaceModule` according to its `SortOrder`. When the order is `SortOrder::Added`, there is no meaningful initial sorting anyway.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -97,9 +97,10 @@ focused = []
 inhibit = ["chrono"]
 
 keyboard = ["dep:colpetto", "dep:evdev-rs", "dep:rustix", "futures-lite"]
-"keyboard+all" = ["keyboard", "keyboard+sway", "keyboard+hyprland"]
+"keyboard+all" = ["keyboard", "keyboard+sway", "keyboard+hyprland", "keyboard+niri"]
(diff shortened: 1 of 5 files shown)


JakeStanger:
Hey, thanks for this. Looks like one of the feature flags is failing. Could you move around the feature gates in the code to resolve please?

I'll be looking to merge this next update, which will also drop the compositor-specific workspace client implementations in favour of a wayland one #1465, so heads up this may need some reworking to handle that.

avasich:
Done and noted!

PR #1604, opened 2026-09-09
fix(tray): call AboutToShow before opening tray menus

molnari811023:
## What

Follows the [DBusMenu spec](https://wiki.ubuntu.com/NotificationArea/MenuImplementation) by sending an `AboutToShow` signal to the SNI item **before** (and while) its popover menu is opened, using the `about_to_show_menuitem(address, path, id)` method already exposed by the `system-tray` crate.

## Why

Many items build their menus lazily and only refresh their layout in response to `AboutToShow`. The most visible example is `nm-applet`: the list of available Wi-Fi networks is only populated when something explicitly triggers it. Native panels (GNOME Shell, Plasma, etc.) do this automatically, so this is invisible to most users — but on a standalone bar like ironbar the network list stays empty/stale until a coincidental `LayoutUpdated` arrives.

## Changes

- `src/clients/tray.rs`: passthrough `about_to_show_menuitem` on ironbar's tray client (mirrors the existing `activate` passthrough).
- `src/modules/tray/mod.rs`: new `UiEvent::AboutToShow { address, path }`, handled in the module controller by calling `about_to_show_menuitem(address, path, 0)`.
- `src/modules/tray/interface.rs`: the click handler now sends this event when opening the menu. To let the handler read the D-Bus object path (previously only stored on the struct and not used at click time), the path is shared through an `Rc<RefCell<Option<String>>>` instead of a plain `Option<String>`.

### DIFF
--- a/src/clients/tray.rs
+++ b/src/clients/tray.rs
@@ -107,6 +107,17 @@ impl Client {
     pub async fn activate(&self, req: ActivateRequest) -> system_tray::error::Result<()> {
         self.client.activate(req).await
     }
+
+    /// Requests the menu to refresh its contents before being shown
+    /// (DBusMenu `AboutToShow`). Returns whether the menu should be re-fetched.
+    pub async fn about_to_show_menuitem(
+        &self,
+        address: String,
+        menu_path: String,
+        id: i32,
+    ) -> system_tray::error::Result<bool> {
+        self.client.about_to_show_menuitem(address, menu_path, id).await
+    }
 }
 
 register_fallible_client!(Client, tray);
--- a/src/modules/tray/interface.rs
+++ b/src/modules/tray/interface.rs
@@ -11,7 +11,9 @@ use gtk::{
     ShortcutTrigger, prelude::*,
 };
 use gtk::{Button, Label, PopoverMenu};
+use std::cell::RefCell;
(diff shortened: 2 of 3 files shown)


JakeStanger:
Looks good - could you fix the build issues please

molnari811023:
The code works. If a few missing spaces are a dealbreaker for a working bugfix, feel free to close the PR. I'm done here.

JakeStanger:
Yes, I ask that you fix both linting and formatting issues, as per [the contribution guide](https://github.com/JakeStanger/ironbar/blob/master/CONTRIBUTING.md) which I expect any potential contributors to read before submitting code, as to avoid wasting anybody's time. Enforcing such standards on a public repository is both good practice and normal in an attempt to meet some minimum level of code quality.

The guidelines do also ask that you don't submit fully AI generated content, and this is clearly AI generated. I don't appreciate the lack of effort to even write your own description, and then the curt reply when I ask for the absolute bare minimum changes.

If you can't or don't want to continue working on a PR for any reason that's fine, you just have to say. Since the fix supposedly works and the code is most of the way there, I'll pick it up from here.

JakeStanger (approved):
One `cargo fmt` and one line change later, good to go

